PR #8, opened 2023-04-13
Update links to new org bitcoin-sv

sirdeggen:
This probably means we have to update all of the library references too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## Documentation
 
-- Find full documentation at [https://taal-gmbh.github.io/arc](https://taal-gmbh.github.io/arc)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,9 +71,9 @@ package main
 import (
 	"fmt"
 
-	"github.com/TAAL-GmbH/arc/api"
-	apiHandler "github.com/TAAL-GmbH/arc/api/handler"
-	"github.com/TAAL-GmbH/arc/api/transactionHandler"
+	"github.com/bitcoin-sv/arc/api"
```
boecklim, line 74: If this example code is changed, then also all the references in the code have to be changed. That starts with changing the first line in `go.mod` from `module github.com/TAAL-GmbH/arc` to `module github.com/bitcoin-sv/arc`.
Thus I'd either change all the paths in the code too or revert the change in this example

PR #10, opened 2023-04-17
Add server of hosted API so that documentation allows trying API endp…

boecklim:
…oints with correct URL

### DIFF
--- a/api/arc.json
+++ b/api/arc.json
@@ -8,6 +8,11 @@
       "url": "https://bitcoinassociation.net/open-bsv-license/"
     }
   },
+  "servers": [
+    {
+      "url": "https://tapi.taal.com/arc"
+    }
+  ],
   "paths": {
     "/v1/policy": {
       "get": {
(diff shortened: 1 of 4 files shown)


icellan:
@boecklim This is hard coding the docs of ARC to Taal. Is that the idea here? ARC is an open-source project by BA.

boecklim:
@icellan The idea of this PR was that in Swagger-UI we can try out the endpoints like you can see here: https://boecklim.github.io/arc/api
Also you get a `curl` example

The servers array allows you to specify multiple base URLs where the API is deployed https://swagger.io/docs/specification/api-host-and-base-path/

icellan:
@boecklim I understand that, but it is hard coding the TAAL endpoints into a BA open-source project that is used by others (GorillaPool).

Is that the idea here? Is that correct?

boecklim:
No, it is not the idea to hard code our base url into the project. It is just about the documentation.

We had the Taal base url in the previous documentation also
image

Maybe, I find a way to have the base url in our hosted API (https://bitcoin-sv.github.io/arc/api.html) without putting it in `arc.yaml`
Alternatively I could disable the `Try it out` feature of Swagger if you prefer that.

But as it is now, it is not correct. Trying out uses the wrong base url:
image

PR #12, opened 2023-04-20
Limit transaction size to 100mb

boecklim:


api/transactionHandler/metamorph.go
```diff
@@ -26,14 +26,17 @@ type Metamorph struct {
...
-func NewMetamorph(targets string, blockTxClient blocktx.ClientI) (*Metamorph, error) {
+func NewMetamorph(targets string, blockTxClient blocktx.ClientI, grpcMessageSize int) (*Metamorph, error) {
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithChainUnaryInterceptor(grpc_prometheus.UnaryClientInterceptor),
 		grpc.WithChainStreamInterceptor(grpc_prometheus.StreamClientInterceptor),
 		grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`), // This sets the initial balancing policy.
 	}
-	conn, err := grpc.Dial(targets, tracing.AddGRPCDialOptions(opts)...)
+
```
icellan, line 36: @boecklim Why didn't you just do:
```
opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(grpcMessageSize)))
```

and leave the rest of the code as it is?

It's not that important, but anytime you introduce a new variable, new memory needs to be allocated which is expensive.

icellan (approved):

PR #28, opened 2023-07-18
Unsubscribe from invalidtx topic

boecklim:
Unsubscribe from `invalidtx` topic so that chained transactions are not rejected immediately


metamorph/zmq.go
```diff
@@ -166,10 +166,6 @@ func (z *ZMQ) Start() {
 		z.logger.Fatal(err)
 	}
 
-	if err = zmq.Subscribe("invalidtx", ch); err != nil {
```
shotasilagadzetaal, line 169: Can we only omit the error `IsMissingInputs`. You can add `continue` in check `if txInfo.IsMissingInputs` so that we still catch error `txInfo.IsDoubleSpendDetected`

farzan1990 (approved):

shotasilagadzetaal (approved):

PR #31, opened 2023-07-21
add retry policy for transactions that are not seen on the network fo…

shotasilagadzetaal:
We fix retry policy for the transactions that are broadcasted but never seen on the network. In that case we broadcast those transactions periodically for a few (15 now) times.

### DIFF
--- a/metamorph/processor.go
+++ b/metamorph/processor.go
@@ -204,25 +204,21 @@ func (p *Processor) processExpiredSeenTransactions() {
 func (p *Processor) processExpiredTransactions() {
 	// filterFunc returns true if the transaction has not been seen on the network
 	filterFunc := func(p *processor_response.ProcessorResponse) bool {
-		return p.GetStatus() < metamorph_api.Status_SEEN_ON_NETWORK && time.Since(p.Start) > 60*time.Second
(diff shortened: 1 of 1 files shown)


metamorph/processor.go
```diff
@@ -69,6 +69,14 @@ type Processor struct {
 	retries            *stat.AtomicStat
 }
 
+const (
+	// number of times we will retry broacasting transaction if we haven't seen it on the network
+	MaxRetries = 15
+	// length of interval for checking transactions if they are seen on the network
+	// if not we resend them again for a few times
+	UnseenTransactionRebroadcastingInterval = 180
```
boecklim, line 77: I think this 180 seconds maybe too long. As that means that a user may have to wait 3min to see the `SEEN_ON_NETWORK` status of a transaction.

Have you tried the same functionality with 60 seconds? What was the result if you changed this constant to 60s?

shotasilagadzetaal: It's not related to returning user response. It's just how we re-announce transaction if we don't have seen_on_network status yet. 

60sec will also work, we don't have a "correct" values here. 

I can change it to 60 seconds if that sound better, what about MaxRetries?

boecklim: Yes, I think 60s is better to start with and see how that improves. Max retries of 15 is reasonable for me 👍

boecklim (approved):

PR #33, opened 2023-07-27
Get transaction blocks for all unseen transactions in a single database query

boecklim:
For the seen txs it is periodically checked if these txs have been mined with a query which runs for each of these transactions.
There can be a lot of them. Here is an example from the logs
`Jul 26 23:33:45 ip-172-28-48-167 arc[36333]: | processor.go:176    | proc | INFO  | checking 11153 expired seen transactions in blocktx`

This PR introduces one query for all the seen txs which should improve performance and load on the DB. Once block height and hash for each seen tx is retrieved it is then looped over the result and each tx marked as mined.

The periodicity has been reduced to 5 min from 10 min.
There are some further modifications in order to make the function `processExpiredSeenTransactions()` testable.

Files which are called `mock.go` do not need to be looked at as they were generated


blocktx/store/sql/GetTransactionBlock.go
```diff
@@ -25,9 +17,19 @@ func (s *SQL) GetTransactionBlock(ctx context.Context, transaction *blocktx_api.
 		WHERE t.hash = $1
 		AND b.orphanedyn = false
 	`
+)
+
+func (s *SQL) GetTransactionBlock(ctx context.Context, transaction *blocktx_api.Transaction) (*blocktx_api.Block, error) {
+	start := gocore.CurrentNanos()
+	defer func() {
+		gocore.NewStat("blocktx").NewStat("GetTransactionBlock").AddTime(start)
```
shotasilagadzetaal, line 25: I think you can just write `defer gocore.NewStat("blocktx").NewStat("GetTransactionBlock").AddTime(start)`

blocktx/store/sql/GetTransactionBlocks.go
```diff
@@ -1,54 +1,101 @@
...
-		blocks = append(blocks, &blocktx_api.Block{
-			Hash: blockHash,
-		})
+		newBlockTransaction := &blocktx_api.TransactionBlock{
+			BlockHash:       BlockHash,
+			BlockHeight:     BlockHeight,
+			TransactionHash: TransactionHash,
+		}
+
+		results.TransactionBlocks = append(results.TransactionBlocks, newBlockTransaction)
```
shotasilagadzetaal, line 97: Maybe I'm missing something but isn't `results.TransactionBlocks` nill at this point? Shouldn't we initialise internal slice first?

blocktx/store/sql/GetTransactionBlock.go
boecklim: I added another commit which fixes this.

blocktx/store/sql/GetTransactionBlocks.go
boecklim: Now that you point it out I'm also not sure when `results.TransactionBlocks` gets initialized.

I added a unit test for that function `TestGetTransactionBlocks` which shows that this slice is already initialized at that point. If it weren't, there would a be a panic when running the test.
https://github.com/bitcoin-sv/arc/pull/33/commits/dc59fba6bbe09bb149c41f1f18adaa7c89c7228c

shotasilagadzetaal (approved):

metamorph/processor_response/ProcessorResponse.go
```diff
@@ -87,7 +87,10 @@ func (r *ProcessorResponse) Close() {
 	defer func() {
 		_ = recover()
 	}()
-	close(r.statusUpdateCh)
+
+	if (r.statusUpdateCh) != nil {
```
icellan, line 91: Doesn't the linter complain about the parenthesis here?

PR #34, opened 2023-07-30
Chain transaction failure fix

shotasilagadzetaal:
Fixing chained transactions by executing them sequentially and storing the transaction information before announcing it to the network.


api/handler/default.go
```diff
@@ -249,16 +246,23 @@ func (m ArcDefaultHandler) POSTTransactions(ctx echo.Context, params api.POSTTra
 			return ctx.JSON(int(api.ErrStatusBadRequest), e)
 		}
 
-		transactions = make([]interface{}, len(txBody))
+		transactionInputs := make([]*bt.Tx, 0)
```
icellan, line 249: Better would be to initialize the array with a cap, otherwise a malloc is done everytime an element is added:
`transactionInputs := make([]*bt.Tx, 0, len(txBody))`

metamorph/Server.go
```diff
@@ -208,6 +208,65 @@ func (s *Server) PutTransaction(ctx context.Context, req *metamorph_api.Transact
...
+	defer gocore.NewStat("PutTransactions").AddTime(start)
+
+	// prepare response object before filling with tx statuses
+	ret := &metamorph_api.TransactionStatuses{}
+	ret.Statuses = make([]*metamorph_api.TransactionStatus, len(req.Transactions))
+
+
+	// for each transaction if we have status in the db already set that status in the response
+	// if not we store the transaction data and set the transaction status in response array to - STORED
+	requests := make([]*store.StoreData, 0)
```
icellan, line 224: Better would be to init with a cap:
`requests := make([]*store.StoreData, 0, len(req.Transactions))`

api/handler/default.go
```diff
@@ -441,6 +399,61 @@ func (m ArcDefaultHandler) processTransaction(ctx context.Context, transaction *
 	}, nil
 }
 
+// processTransactions validates all the transactions in the array and submits to metamorph for processing
+func (m ArcDefaultHandler) processTransactions(ctx context.Context, transactions []*bt.Tx, transactionOptions *api.TransactionOptions) ([]interface{}, error) {
+	span, tracingCtx := opentracing.StartSpanFromContext(ctx, "ArcDefaultHandler:processTransactions")
+	defer span.Finish()
+
+	// validate before submitting array of transactions to metamorph
+	transactionsInput := make([][]byte, 0)
```
boecklim, line 408: I think also here memory can be preallocated like this
`transactionsInput := make([][]byte, len(transactions))`

api/handler/default.go
shotasilagadzetaal: Fixed. Though malloc isn't called on EVERY element. The capacity increases exponentially. But good catch anyway

boecklim (approved):

PR #37, opened 2023-08-02
Upgrade go-bt

boecklim:
Upgrade go-bt so that fix from this [PR](https://github.com/libsv/go-bt/pull/164) is fixed in ARC.


sirdeggen (approved):
lgtm

oskarszoon (approved):

PR #41, opened 2023-08-11
Fix/failing unit tests

boecklim:
- add processor shutdown function used in unit test
- pass default policy as a parameter to arc handler

### DIFF
--- a/api/handler/load.go
+++ b/api/handler/load.go
@@ -41,14 +41,10 @@ func LoadArcHandler(e *echo.Echo, logger utils.Logger) error {
 	}
 
 	var apiHandler api.HandlerInterface
-	defaultPolicySetting, found := gocore.Config().Get("defaultPolicy")
-	var defaultPolicy *bitcoin.Settings
-	if found && defaultPolicySetting != "" {
-		if err = json.Unmarshal([]byte(defaultPolicySetting), &defaultPolicy); err != nil {
-			return err
-		}
+	defaultPolicy, err := GetDefaultPolicy()
+	if err != nil {
+		return err
 	}
-
 	// TODO WithSecurityConfig(appConfig.Security)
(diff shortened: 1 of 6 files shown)


api/handler/load.go
```diff
@@ -39,8 +41,16 @@ func LoadArcHandler(e *echo.Echo, logger utils.Logger) error {
 	}
 
 	var apiHandler api.HandlerInterface
+	defaultPolicySetting, found := gocore.Config().Get("defaultPolicy")
+	var defaultPolicy *bitcoin.Settings
+	if found && defaultPolicySetting != "" {
+		if err = json.Unmarshal([]byte(defaultPolicySetting), &defaultPolicy); err != nil {
```
shotasilagadzetaal, line 47: Are you passing pointer to the pointer here? Is it what the function expects? Have you tested that part ? 🤔

metamorph/processor.go
```diff
@@ -171,6 +171,22 @@ func (p *Processor) Set(req *ProcessorRequest) error {
 	return p.store.Set(spanCtx, req.Hash[:], req.StoreData)
 }
 
+// Close all channels and goroutines for graceful shutdown
+func (p *Processor) Shutdown() {
+	p.logger.Infof("Shutting down processor")
+	p.processExpiredSeenTxsTicker.Stop()
+	p.processExpiredTxsTicker.Stop()
+	if p.ch != nil {
```
shotasilagadzetaal, line 179: where do we use `p.ch` 🤔  Don't remember what channel is that. I don't see it in the code, am I missing something?

api/handler/load.go
boecklim: Good point! Will resolve and test.

metamorph/processor.go
boecklim: Also good point. Will check that.

shotasilagadzetaal (approved):

PR #42, opened 2023-08-14
remove incorrect block hash query

shotasilagadzetaal:
We had error in blocktx (issue BPAAS-934) because we incorrectly assumed that the block hash would be present at the point of query. Removed the assumption and query.

### DIFF
--- a/blocktx/store/sql/RegisterTransaction.go
+++ b/blocktx/store/sql/RegisterTransaction.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/bitcoin-sv/arc/blocktx/blocktx_api"
 	"github.com/lib/pq"
+	"github.com/libsv/go-p2p/chaincfg/chainhash"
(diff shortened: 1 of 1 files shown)


blocktx/store/sql/RegisterTransaction.go
```diff
@@ -95,27 +94,6 @@ func (s *SQL) RegisterTransaction(ctx context.Context, transaction *blocktx_api.
 			return "", nil, 0, err
 		}
```
icellan, line 96: This is a bit too rigorous of a change. The idea here was to check whether the transaction had already been mined and return the block hash and height of the block it was mined in.

The discussion we had on this earlier was regarding whether this logic was entirely correct, which it might not have been, but if the transaction has already been mined, that should be returned from this function call, which now has been removed.

icellan (changes requested):

icellan (approved):

PR #45, opened 2023-08-30
Add handled errors to response of POSTTransactions endpoint

boecklim:
After the change in PR https://github.com/bitcoin-sv/arc/pull/34

Handled errors were not anymore in the response of a batch submission, but only the successful statuses.

This PR:
- re-adds the handled error to the response
- Calculate sizings which has also been lost in PR https://github.com/bitcoin-sv/arc/pull/34
- Fixes unit tests


api/handler/default.go
```diff
@@ -249,6 +249,7 @@ func (m ArcDefaultHandler) POSTTransactions(ctx echo.Context, params api.POSTTra
 			return ctx.JSON(int(api.ErrStatusBadRequest), e)
 		}
 
+		txMap := make(map[string]*bt.Tx)
```
icellan, line 252: Why do you create a map of the whole transaction, instead of just the sizing, which is the only thing it is being used for?

icellan (changes requested):

icellan (approved):

PR #46, opened 2023-08-31
Viper config structure

boecklim:
- Introduction Viper package for setting configuration
- Creation of a default settings in file `config.yaml`
- Rename and structure the settings

Out of scope (next PR)
- Adding documentation about config file in README.md
- Documenting different settings in config file


api/handler/default.go
```diff
@@ -561,14 +562,34 @@ func (m ArcDefaultHandler) getTransaction(ctx context.Context, inputTxID string)
...
+
+	peerRpcHost := viper.GetString("peerRpc.host")
+	if peerRpcHost == "" {
+		return nil, errors.Errorf("setting peerRpc.host not found")
+	}
+
+	peerRpcPort := viper.GetInt("peerRpc.port")
+	if peerRpcPort == 0 {
+		return nil, errors.Errorf("setting peerRpc.port not found")
+	}
```
dpcat237, line 584: If this RPC configuration is loaded only when the app starts, then, in my opinion, is cleaner to pass this configuration in `NewDefaul()`. As there are multiple parameters, it would be cleaner to pass them inside a struct.
Benefits:
- Config parameters in one place.
- Visible what parameters are necessary to instantiate `api.HandlerInterface`.
- Loading/validating config parameters is not directly related to the logic of getting a policy from a node. Moving this logic out of `getPolicyFromNode()` makes it easier to test.

If `getPolicyFromNode()` could be called at run time and config parameters could be changed without restarting the app, then it would be cleaner to move the loading process of these parameters inside a method of config struct inside `handler` package.

blocktx/peer_settings.go
```diff
@@ -0,0 +1,54 @@
...
+	if p.Host == "" {
+		return "", fmt.Errorf("host not set for peer %s", p.Host)
+	}
+
+	return fmt.Sprintf("%s:%d", p.Host, p.Port.P2P), nil
+}
+
+func GetPeerSettings() ([]Peer, error) {
+	var peers []Peer
+	err := viper.UnmarshalKey("peers", &peers)
```
dpcat237, line 48: In my opinion is cleaner to validate/format the settings when they are loaded in `GetPeerSettings()`. This way `GetZMQUrl()` and `GetP2PUrl()` are not necessary.

dpcat237 (approved):
Overall the changes look good and I don't see any critical issue.

On another hand, I see several drawbacks that loading and validation of the settings are spread around the code:
- Settings loading/validation is mixed with business logic.
- Involved functions are not idempotent which requires settings loading in unit tests and more time reading the code to find all required parameters.
- Is more difficult to maintain the settings.
- The same configuration is loaded multiple times.

PS. I see that the `viper` library gives more freedom for good and for bad. And libraries like [cleanenv](https://github.com/ilyakaznacheev/cleanenv) are smaller in features but cleaner.

farzan1990 (approved):

shotasilagadzetaal (approved):

PR #52, opened 2023-09-07
regtest config

farzan1990:


oskarszoon (approved):

main.go
```diff
@@ -36,6 +37,8 @@ func main() {
 	startCallbacker := flag.Bool("callbacker", false, "start callbacker")
 	useTracer := flag.Bool("tracer", false, "start tracer")
 	help := flag.Bool("help", false, "Show help")
+	var regtest bool
+	flag.BoolVar(&regtest, "regtest", false, "Use regtest-config.yml instead of the default config.yaml")
```
boecklim, line 41: I would suggest here to allow giving the name of the config file optionally. That would be more flexible.

`
config := flag.String("configFile", "config", "name of the config yaml file")
`

main.go
```diff
@@ -61,15 +64,21 @@ func main() {
...
 	}
 
-	viper.SetConfigName("config") // name of config file (without extension)
-	viper.SetConfigType("yaml")   // REQUIRED if the config file does not have the extension in the name
-	viper.AddConfigPath(".")      // optionally look for config in the working directory
-	err := viper.ReadInConfig()   // Find and read the config file
-	if err != nil {               // Handle errors reading the config file
-		fmt.Printf("failed to read config file config.yaml: %v \n", err)
-		return
+	if regtest {
```
boecklim, line 67: Then you could do here this

`
	configName := "config"
	if config != nil {
		configName = *config
	}
	viper.SetConfigName(configName) // name of config file (without extension)
`

PR #55, opened 2023-09-12
Validate callback url in API and in metamorph

boecklim:
- Validate the callback URL in both API and metamorph
- This ensures that the user will be informed about his callback URL being invalid
- Otherwise the user could never understand why the callback URL has not been called if the transaction was mined

### DIFF
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -25,3 +25,11 @@ func FeesToBtFeeQuote(minMiningFee float64) *bt.FeeQuote {
 
 	return btFeeQuote
 }
+
+func StringToPointer(input string) *string {
+	return &input
+}
+
+func IntToPointer(input int) *int {
+	return &input
+}
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -25,11 +25,3 @@ func FeesToBtFeeQuote(minMiningFee float64) *bt.FeeQuote {
 
 	return btFeeQuote
 }
-
-func StringToPointer(input string) *string {
-	return &input
-}
-
-func IntToPointer(input int) *int {
-	return &input
-}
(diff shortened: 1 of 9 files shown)


icellan (approved):

api/helpers.go
```diff
@@ -25,3 +25,11 @@ func FeesToBtFeeQuote(minMiningFee float64) *bt.FeeQuote {
 
 	return btFeeQuote
 }
+
+func StringToPointer(input string) *string {
```
shotasilagadzetaal, line 29: Have never seen those. Do we really need a function for that?

metamorph/Server.go
```diff
@@ -224,6 +228,10 @@ func (s *Server) PutTransactions(ctx context.Context, req *metamorph_api.Transac
 	// if not we store the transaction data and set the transaction status in response array to - STORED
 	requests := make([]*store.StoreData, 0, len(req.Transactions))
 	for ind, req := range req.Transactions {
+		if _, err := url.ParseRequestURI(req.CallbackUrl); req.CallbackUrl != "" && err != nil {
```
shotasilagadzetaal, line 231: If err is not nil it automatically means that req.CallbackUrl is not empty (so don't need to check) but that's ok

shotasilagadzetaal (approved):

api/helpers.go
boecklim: Unfortunately it is not possible to get a pointer to a string like this `&"some string"`
I was looking for a standard library function for that. But didn't find one.

boecklim: Ok, I think I may have found it though: https://pkg.go.dev/go.octolab.org/pointer#ToString

metamorph/Server.go
boecklim: No, because if the callback is empty (which is a valid input), then there will also be an error. Therefor we must return an error only if `err != nil && req.CallbackUrl != ""`

api/helpers.go
shotasilagadzetaal: `somestring := string("shota")
	ptr := &somestring` or you wanted in one line?

PR #58, opened 2023-09-14
Merkle path

shotasilagadzetaal:


blocktx/PeerHandler.go
```diff
@@ -290,20 +291,29 @@ func (bs *PeerHandler) HandleBlock(wireMsg wire.Message, peer p2p.PeerI) error {
...
-		return fmt.Errorf("unable to mark block as mined %s: %v", blockHash.String(), err)
+	transactionHashes := make([]string, len(msg.TransactionHashes))
+	for i, hash := range msg.TransactionHashes {
+		transactionHashes[i] = hash.String()
 	}
 
-	transactionHashes := make([][]byte, len(msg.TransactionHashes))
-	for i, hash := range msg.TransactionHashes {
-		transactionHashes[i] = hash[:]
+	calculatedMerkleRoot, err := bc.BuildMerkleTreeStore(transactionHashes)
```
boecklim, line 299: It would be better to call this `calculatedMerkleTree` instead of `calculatedMerkleRoot` as the root is only the last element of that slice

boecklim (commented):
We'll need some added unit test coverage for `HandleBlock()`

icellan (commented):
It looks good, but why did you guys add all the merkle proofs for all transactions into the database? It should be fairly quick to calculate the merkle proof if you store the list of transactions in an efficient way to load them.

shotasilagadzetaal:
> It looks good, but why did you guys add all the merkle proofs for all transactions into the database? It should be fairly quick to calculate the merkle proof if you store the list of transactions in an efficient way to load them.

Then you would need to calculate the path every time you would need it. The only optimization is to just calculate those transaction merkle paths that are interesting for us.

boecklim (approved):

boecklim (commented):
I think we should start naming files according to Golang name conventions: https://blog.devgenius.io/golang-naming-conventions-72bbaf84e959

In this convention files should be named in snake case. So `InsertBlockTransactions.go`  would become `insert_block_transactions.go`

PR #60, opened 2023-09-14
retrieve merkle path from db and return in response for users

shotasilagadzetaal:


metamorph/metamorph_api/metamorph_api.proto
```diff
@@ -85,10 +85,11 @@ message TransactionStatus {
   google.protobuf.Timestamp announced_at = 3;
   google.protobuf.Timestamp mined_at = 4;
   string txid = 5;
-  Status status = 6;
-  string reject_reason = 7;
-  uint64 block_height = 8;
-  string block_hash = 9;
+  string merkle_path = 6;
```
boecklim, line 88: I think it would be better to add the merkle path at the end as the proto object becomes in compatible with a previous version.

blocktx/blocktx_api/blocktx_api.proto
```diff
@@ -100,15 +103,21 @@ message Source {
...
 message TransactionAndSource {
   bytes hash = 1;
   string source = 2;
 }
 
 message RegisterTransactionResponse {
   string source = 1;
-  bytes block_hash = 2;
-  uint64 block_height = 3;
+  string merkle_path = 2;
```
boecklim, line 118: I think it would be better to add the merkle path at the end as the proto object becomes in compatible with a previous version.

blocktx/Server.go
```diff
@@ -108,11 +108,28 @@ func (s *Server) LocateTransaction(ctx context.Context, transaction *blocktx_api
...
+			return nil, ErrTransactionNotFound
+		}
+		return nil, err
+	}
+
+	return &blocktx_api.MerklePath{MerklePath: merklePath}, nil
+}
+
+func (s *Server) RegisterTransaction(ctx context.Context, transaction *blocktx_api.TransactionAndSource) (*blocktx_api.RegisterTransactionResponse, error) {
+	source, merkle_path, hash, height, err := s.store.RegisterTransaction(ctx, transaction)
```
boecklim, line 129: `merkle_path` should be `merklePath`

blocktx/store/sql/GetTransactionMerklePath.go
```diff
@@ -0,0 +1,40 @@
...
+	defer cancel()
+
+	q := `
+		SELECT
+		 t.merkle_path
+		FROM transactions t
+		WHERE t.hash = $1
+	`
+
+	var merkle_path sql.NullString
```
boecklim, line 30: Name of var `merkle_path` should be `merklePath`.

api/transactionHandler/metamorph.go
```diff
@@ -88,8 +89,24 @@ func (m *Metamorph) GetTransactionStatus(ctx context.Context, txID string) (stat
 		return nil, ErrTransactionNotFound
 	}
 
+	hash, err := chainhash.NewHashFromStr(txID)
+	if err != nil {
+		return nil, err
+	}
+
+	var merkle_path string
```
boecklim, line 97: var name `merkle_path` should be camel case: `merklePath`

blocktx/PeerHandler.go
```diff
@@ -352,21 +361,42 @@ func (bs *PeerHandler) insertBlock(blockHash *chainhash.Hash, merkleRoot *chainh
...
+			merklePath, err = bc.GetTxMerklePath(txIndex, merkleTree).String()
+			if err != nil {
+				return err
+			}
+
+			root, err := bc.GetTxMerklePath(txIndex, merkleTree).CalculateRoot(hash.String())
+			if err != nil {
+				return err
+			}
+			fmt.Println(root)
```
boecklim, line 390: Was this printing used for debugging?
If not, rather use the logger to print it.

blocktx/PeerHandler.go
```diff
@@ -352,21 +361,42 @@ func (bs *PeerHandler) insertBlock(blockHash *chainhash.Hash, merkleRoot *chainh
...
+
+		var merklePath string
+		var err error
+		if len(transactionHashes) != 1 {
+			merklePath, err = bc.GetTxMerklePath(txIndex, merkleTree).String()
+			if err != nil {
+				return err
+			}
+
+			root, err := bc.GetTxMerklePath(txIndex, merkleTree).CalculateRoot(hash.String())
```
boecklim, line 386: If we don't need to log the root, then we can also remove this line.

blocktx/PeerHandler.go
```diff
@@ -290,20 +290,29 @@ func (bs *PeerHandler) HandleBlock(wireMsg wire.Message, peer p2p.PeerI) error {
...
-		return fmt.Errorf("unable to mark block as mined %s: %v", blockHash.String(), err)
+	transactionHashes := make([]string, len(msg.TransactionHashes))
+	for i, hash := range msg.TransactionHashes {
+		transactionHashes[i] = hash.String()
 	}
 
-	transactionHashes := make([][]byte, len(msg.TransactionHashes))
-	for i, hash := range msg.TransactionHashes {
-		transactionHashes[i] = hash[:]
+	calculatedMerkleRoot, err := bc.BuildMerkleTreeStore(transactionHashes)
```
boecklim, line 298: I'd rename this to `calculatedMerkleTree` as only the last element is the root

blocktx/PeerHandler.go
```diff
@@ -290,20 +290,29 @@ func (bs *PeerHandler) HandleBlock(wireMsg wire.Message, peer p2p.PeerI) error {
...
+	if err != nil {
+		return err
 	}
 
-	calculatedMerkleRoot := blockchain.BuildMerkleTreeStore(transactionHashes)
-	if !bytes.Equal(calculatedMerkleRoot[len(calculatedMerkleRoot)-1], merkleRoot[:]) {
+	hash, err := chainhash.NewHashFromStr(calculatedMerkleRoot[len(calculatedMerkleRoot)-1])
```
boecklim, line 303: hash is not a very expressive name. Maybe `RootHash`?

blocktx/store/sql/InsertBlockTransactions.go
```diff
@@ -42,7 +42,7 @@ func (s *SQL) InsertBlockTransactions(ctx context.Context, blockId uint64, trans
 	for pos, tx := range transactions {
 		var txid uint64
 
-		if err = qTx.QueryRowContext(ctx, tx.Hash).Scan(&txid); err != nil {
+		if err = qTx.QueryRowContext(ctx, tx.Hash, merklePaths[pos]).Scan(&txid); err != nil {
```
boecklim, line 45: It's not your code but, this block looks very complicated to me. I think it can be simplified to this:

```
		err = qTx.QueryRowContext(ctx, tx.Hash, merklePaths[pos]).Scan(&txid)
		if err != nil {
			if err != sql.ErrNoRows {
				return err
			}

			err = s.db.QueryRowContext(ctx, `
					SELECT id
					FROM transactions
					WHERE hash = $1
				`, tx.Hash).Scan(&txid)
			if err != nil {
				return err
			}
		}
```

blocktx/store/sql/RegisterTransaction.go
```diff
@@ -69,30 +69,30 @@ func (s *SQL) RegisterTransaction(ctx context.Context, transaction *blocktx_api.
...
+			return "", "", nil, 0, err
 		}
 
-		rows, err := result.RowsAffected()
-		if err != nil {
-			if spanErr != nil {
-				spanErr.SetTag(string(ext.Error), true)
-				spanErr.LogFields(log.Error(err))
+		var rows int
+		var merkle_path string
```
boecklim, line 90: var `merkle_path` is in snake case but should be camel case `merklePath`

boecklim (changes requested):
Overall looks good. Some minor changes would be good to be done.

metamorph/metamorph_api/metamorph_api.proto
shotasilagadzetaal: Do we need to be compatible? if not it's working rn

blocktx/blocktx_api/blocktx_api.proto
shotasilagadzetaal: same as above

PR #78, opened 2023-10-02
Get tx status tests

nozim:
* Add initial unit test for `GetTransactionStatus` handler


boecklim (commented):
The functions which are tested here have been generated.
I’m not sure if we should create test cases for generated code.
Right now we have all generated code like arc.go excluded from the SonarQube analysis for test coverage
https://github.com/bitcoin-sv/arc/blob/d1a776878f45c0e5c493923da1590ca6462819b1/sonar-project.properties#L10

nozim:
@boecklim yes I suppose I'll change the scope of test to non generated part.

PR #80, opened 2023-10-03
Rename files to snake case

nozim:


blocktx/peer_handler.go
```diff
@@ -93,32 +94,38 @@ func init() {
...
-			if err := msg.AddInvVect(wire.NewInvVect(wire.InvTypeBlock, &hash)); err != nil {
-				logger.Errorf("EvictionFunc: could not create InvVect: %v", err)
-				return false
-			}
+		if err := msg.AddInvVect(wire.NewInvVect(wire.InvTypeBlock, &hash)); err != nil {
+			logger.Errorf("EvictionFunc: could not create InvVect: %v", err)
+			return false
+		}
+		// Select a random peer to send the request to
+		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(peers))))
```
boecklim, line 105: What is the rationale behind changing `peer := peers[rand.Intn(len(peers))]` to this?

nozim: it was recommendation from sonarcube gosec check. I can revert it to standard math rand or we can disable this check.

metamorph/processor_response_map.go
```diff
@@ -53,25 +53,26 @@ func NewProcessorResponseMap(expiry time.Duration) *ProcessorResponseMap {
...
-
-	f, err := os.OpenFile(m.logFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	err := os.MkdirAll(dir, 0750)
+	if err != nil {
+		log.Fatalf("failed to create folder for logging %s", err)
+	}
+	f, err := os.OpenFile(m.logFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
 	if err != nil {
-		log.Printf("error opening log file: %s", err.Error())
+		log.Fatalf("error opening log file: %s", err)
```
boecklim, line 62: I think `Fatalf()` is not to be used here. Because Fatalf is equivalent to Printf() followed by a call to os.Exit(1).
We shouldn't exit if an error occurs, but just log it.

Regarding the formatted string. We should print the error either like this `"... %s", err.String()` or `"... %v", err` so that the error message will be printed.

metamorph/processor_response_map.go
```diff
@@ -53,25 +53,26 @@ func NewProcessorResponseMap(expiry time.Duration) *ProcessorResponseMap {
 
 func (m *ProcessorResponseMap) logWriter() {
 	dir := path.Dir(m.logFile)
-	_ = os.MkdirAll(dir, 0777)
-
-	f, err := os.OpenFile(m.logFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	err := os.MkdirAll(dir, 0750)
+	if err != nil {
```
boecklim, line 57: Good to handle the error here 👍

metamorph/processor_response_map.go
```diff
@@ -53,25 +53,26 @@ func NewProcessorResponseMap(expiry time.Duration) *ProcessorResponseMap {
...
 	}
 	defer f.Close()
 
 	for prl := range m.logWorker {
-
 		var b []byte
 		b, err = json.Marshal(prl)
 		if err != nil {
-			log.Printf("error marshaling log data: %s", err.Error())
+			log.Fatalf("error marshaling log data: %s", err)
```
boecklim, line 70: I'd leave it at just logging without exit.

metamorph/processor_response_map.go
```diff
@@ -53,25 +53,26 @@ func NewProcessorResponseMap(expiry time.Duration) *ProcessorResponseMap {
...
 		}
 
 		_, err = f.WriteString(string(b) + "\n")
 		if err != nil {
-			log.Printf("error writing to log file: %s", err.Error())
+			log.Fatalf("error writing to log file: %s", err)
```
boecklim, line 75: I'd leave it at just logging without exit.

metamorph/processor_response_map.go
nozim: 1. I'm not sure how it would proceed with the rest if there is an error with opening file. 
2. I could not find `err.String` in standard errors package, probably you meant err.Error()?
3. In this particular case "%s" and "%v" can serve same purpose and .Error() method will be called.

metamorph/processor_response_map.go
nozim: same here, you can't proceed if there is error with parsing json. The `b` value would be useless.

metamorph/processor_response_map.go
boecklim: Oh yes, I meant `err.Error()`. Ok, didn't know that 👍

boecklim: 1. Ok, I see that point

metamorph/processor_response_map.go
boecklim: Ok, but then here we should then maybe do this
```
if err != nil {
  log.Printf("error marshaling log data: %s", err.Error())
  continue
}
```

nozim: yup, will amend

boecklim (approved):

PR #85, opened 2023-10-05
 Add database migrations

nozim:
- added database migrations and golang migrate library 
- added go migrate library installation script 
- added separate docker compose for local arc and postgres database

### DIFF
--- a/database/migrations/000001_create_blocks.up.sql
+++ b/database/migrations/000001_create_blocks.up.sql
@@ -1,13 +1,13 @@
 CREATE TABLE blocks (
-                        inserted_at timestamp with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
-                        id bigint NOT NULL,
-                        hash bytea NOT NULL,
-                        prevhash bytea NOT NULL,
-                        merkleroot bytea NOT NULL,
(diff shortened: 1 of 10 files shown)


database/migrations/000001_create_blocks.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE blocks (
+                        inserted_at timestamp with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
boecklim, line 2: IMO all sql keywords should be capitalised

database/schema.sql 
boecklim, line 1: What is this file used for? Is it the schema for sqlite? Couldn't find any reference to it.

nozim: this is in case if you need to run the current schema all at once. But can be removed as well :)

nozim:
@boecklim for sqlite migrations I was thinking if we could reuse the current ones. The should be universal.

scripts/install-migrate.sh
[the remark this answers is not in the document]
nozim: renamed in latest commit.

boecklim (approved):

PR #88, opened 2023-10-11
test cherry pick

farzan1990:


Makefile
```diff
@@ -92,3 +102,6 @@ gh-pages:
 .PHONY: api
 api:
 	oapi-codegen -config api/config.yaml api/arc.yml > api/arc.go
+
+.PHONY: clean_test
```
boecklim, line 106: clean test sounds like to clean up leftover from a test.
Maybe rename to something like: `clean_restart_e2e_test`

Makefile
```diff
@@ -11,14 +11,24 @@ deps:
 build:
 	go build ./...
 
+.PHONY: restart
```
boecklim, line 14: This does not restart anything
Rename to sth like `clean_e2e_tests`

test/utils.go
```diff
@@ -110,11 +110,11 @@ func generate(t *testing.T, amount uint64, address string) string {
 	return blockHash
 }
 
-func execCommandGenerate(t *testing.T, amount uint64, address string) string {
+func execCommandGenerate(t *testing.T, amount uint64) string {
```
boecklim, line 113: Let's rename all functions with prefix `exec` as we're now the `go-bitcoin` package functions

boecklim (approved):

PR #91, opened 2023-10-12
Add alternative integration tests flow with docker compose

nozim:
- moved integration test related dockerfiles to root folder 
- fixed dockerfile build for arc server 
- changed arc config file to bind to 0.0.0.0 instead of localhost


test/utils.go
```diff
@@ -80,17 +80,6 @@ func getNewWalletAddress(t *testing.T) (address, privateKey string) {
 	return
 }
 
-func sendToAddress(t *testing.T, address string, bsv float64) (txID string) {
```
boecklim, line 83: This function will be needed as soon as this is merged: https://github.com/bitcoin-sv/arc/pull/88

Dockerfile
```diff
@@ -1,10 +1,14 @@
-FROM ubuntu:latest
-
-COPY ./build/arc_linux_amd64 /service/arc
+FROM golang:1.21.1
```
boecklim, line 1: The docker image when using this base image is much larger: ~302.75 MB
With ubuntu base image it is just ~50MB. You can see that here: https://hub.docker.com/r/bsvb/arc/tags

.github/workflows/go.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         run: make api && git diff --exit-code
 
       - name: Run tests
-        run: go test -race -vet=off ./...
+        run: go test -race -vet=off $(go list ./... | grep -v github.com/bitcoin-sv/arc/test)
```
boecklim, line 46: I think the better solution to have separate execution of tests would be to run the e2e tests only if a certain env var is given e.g. `E2E_TESTS=1 go test -race -vet=off` as described here: https://konradreiche.com/blog/how-to-separate-integration-tests-in-go

test/utils.go
nozim: it's indicated by static check. Not sure we need to keep dead code in repo.

Dockerfile
nozim: I'm not sure we have to discuss image size. If that becomes a real problem we can optimize later. What do you think?

.github/workflows/go.yml
nozim: Either approach is fine.  I don't see any additional benefits of the aporoach you mention here. But open for suggestions as always.

boecklim: With this approach you used here this will always fail `go test ./...` unless I add this tail `$(go list ./... | grep -v github.com/bitcoin-sv/arc/test)`.

I'd rather have a solution where I have to add something if I want to run the e2e tests. Usually you'll not want to run them because they only make sense together with the respective docker-compose file

Dockerfile
boecklim: I don't see the benefit in this change. We just create a larger docker image which takes longer to pull and takes more storage.

test/utils.go
boecklim: This is now merged: https://github.com/bitcoin-sv/arc/pull/88
After a rebase that indication by staticcheck should be gone

Dockerfile
nozim: well 
1. with previous dockerfile I wasn't able to build anything 
2. using official golang docker image is somewhat simplest way to get things done as it's recommended here
https://docs.docker.com/language/golang/build-images/
3. I'd be happy to see alternative approach in building ci that works in every environment.

test/utils.go
nozim: I'm not sure what to do here as it's staticcheck that complains about dead code. 
If you are fine with keeping dead code and making it harder for us to maintain we can definitely remove staticcheck.

boecklim: No, I'm not saying I want to keep dead code. If you rebase this branch on `main` you'll see that this function will be used and will therefore not be dead code anymore.

Dockerfile
boecklim: Did you run `make build_release` before building the docker container?

On my mac I had no problem to run this locally with e.g.
```
make build_release
 docker build -t arc .
```

test/utils.go
nozim: yup, you are right. added back ;)

Dockerfile
nozim: We can definitely keep the original way however the Dockerfile I propose is reused for tests so we avoid some overhead in it's maintenance.
What do you think?

.github/workflows/go.yml
boecklim: Still running `go test ./...` would fail now. I prefer to have to add something like an env var in order to run the e2e tests over having to prevent running the e2e tests each time with this tail `$(go list ./... | grep -v github.com/bitcoin-sv/arc/test)`.
This commit shows how it can be done: https://github.com/bitcoin-sv/arc/commit/8c16a77f5a8d30dc6657802a03e348e21ed8a778

nozim: it's env variable vs folder exclusion in go test. I don't see any exceptional benefit but will do it the way you insist.

test/arc_txt_endpoint_test.go
```diff
@@ -41,199 +41,122 @@ type TxStatusResponse struct {
...
+		require.NoError(t, err)
+		defer statusResp.Body.Close()
 
-	// Print the response body for the GET request
-	fmt.Println("Transaction status response body:", string(statusBodyBytes))
+		var statusResponse TxStatusResponse
+		err = json.NewDecoder(statusResp.Body).Decode(&statusResponse)
+		require.NoError(t, err)
+		//TODO: follow up and make sure the tx is MINED
+		if statusResponse.TxStatus == "ANNOUNCED_TO_NETWORK" {
```
boecklim, line 150: Why did you change that?
We want to wait for the "MINED" status. That is why you need at least 2 nodes by the way.

Dockerfile.itest
boecklim, line 1: Now we have 3 docker files. We just need 2

They could be started like this
```
  arc:
    build:
      context: .
      dockerfile: Dockerfile.arc.itest
    ports:
      - "8011:8011"
      - "9090:9090"
      - "9999:9999"
    depends_on:
      - node0
    command: ["/service/arc"]

  tests:
    build:
      context: .
      dockerfile: Dockerfile.arc.itest
    environment:
      NODE_IP: "node0"
      ARC_URL: "http://arc:9090"
    depends_on:
      - arc
    healthcheck:
      test: ["CMD","curl", "http://arc:9090/v1/tx"]
    command: ["go", "test", "-v", "./test"]
```

test/arc_txt_endpoint_test.go
nozim: i had inconsistent test responses but investigating that. Added back 2 other nodes.

Dockerfile.itest
nozim: removed redundant dockerfile.

test/arc_txt_endpoint_test.go
boecklim: Ok, I think that would probably solve it.

nozim: it's not solved yet. What if we merge this and I can follow up with this in next PR?  I'm learning how metamorph internals work and will add some more testcases.  I will create additional documentation as a result.

nozim:
closing this for now. Will continue in other prs.

PR #92, opened 2023-10-13
Add sqlite migration

nozim:
- move postgres migrations to separate folder 
- add sqlite migrations
- add makefile targets to apply database migrations


database/migrations/sqlite/000001_create_blocks.up.sql
```diff
@@ -0,0 +1,15 @@
...
+                        merkleroot BLOB NOT NULL,
+                        height INTEGER NOT NULL,
+                        processed_at DATETIME,
+                        size INTEGER,
+                        tx_count INTEGER,
+                        orphanedyn BOOLEAN DEFAULT 0 NOT NULL,
+                        merkle_path TEXT DEFAULT ''
+);
+
+CREATE INDEX ix_blocks_hash ON blocks (hash);
```
boecklim, line 15: There is another index missing: https://github.com/bitcoin-sv/arc/blob/bbf2eab6ec8add99c2ffe0b3ff4aa7dc4096cae4/blocktx/store/sql/sql.go#L181
```
CREATE UNIQUE INDEX IF NOT EXISTS pux_blocks_height ON blocks(height) WHERE orphanedyn = FALSE;
```

boecklim (approved):

PR #94, opened 2023-10-17
Bpaas 1148 implement interface for dynamodb

shotasilagadzetaal:


metamorph/store/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,461 @@
...
+		}
+	}
+
+	// Using the Config value, create the DynamoDB client
+	return dynamodbClient, nil
+}
+
+func (ddb *DynamoDB) TableExists(tableName string) (bool, error) {
+	// Build the request with its input parameters
+	resp, err := ddb.dynamoCli.ListTables(context.TODO(), &dynamodb.ListTablesInput{})
```
boecklim, line 69: Alternatively here instead of listing the table could be described in order to check if it exists like this
```
resp, err := ddb.dynamoCli.DescribeTable(context.TODO(), &dynamodb.DescribeTableInput{TableName: &tableName})
```

Then it wouldn't be needed to iterate through them.
But this is also fine

metamorph/store/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,461 @@
...
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// CreateTransactionsTable creates a DynamoDB table for storing metamorph user transactions
+func (ddb *DynamoDB) CreateTransactionsTable() error {
+	// construct primary key and create table
+	if _, err := ddb.dynamoCli.CreateTable(context.TODO(), &dynamodb.CreateTableInput{
```
boecklim, line 86: This if clause is huge.
I personally find this nicer, but it is a matter of taste

```
err := ddb.dynamoCli.CreateTable(context.TODO(), &dynamodb.CreateTableInput{...}
if err != nil { ... }
```

metamorph/store/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,461 @@
...
+		gocore.NewStat("mtm_store_dynamodb").NewStat("Get").AddTime(startNanos)
+	}()
+	span, _ := opentracing.StartSpanFromContext(ctx, "dynamodb:Get")
+	defer span.Finish()
+
+	val := map[string]types.AttributeValue{
+		"tx_hash": &types.AttributeValueMemberB{Value: key},
+	}
+
+	response, err := ddb.dynamoCli.GetItem(context.TODO(), &dynamodb.GetItemInput{
```
boecklim, line 181: I think here the context from the parameters can be used `GetItem(ctx, ...)`

metamorph/store/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,461 @@
...
+	// marshal input value for new entry
+	item, err := attributevalue.MarshalMap(value)
+	if err != nil {
+		span.SetTag(string(ext.Error), true)
+		span.LogFields(log.Error(err))
+		return err
+	}
+
+	// put item into table
+	_, err = ddb.dynamoCli.PutItem(context.TODO(), &dynamodb.PutItemInput{
```
boecklim, line 223: Also here...

metamorph/store/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,461 @@
...
+		TableName: aws.String("transactions"), Item: item,
+	})
+
+	if err != nil {
+		span.SetTag(string(ext.Error), true)
+		span.LogFields(log.Error(err))
+		return nil
```
boecklim, line 230: Shouldn't we return the `err` here?

boecklim (approved):

metamorph/store/dynamodb/dynamodb.go
```diff
@@ -43,15 +43,21 @@ func New() (store.MetamorphStore, error) {
 	if err != nil {
 		return dynamodbClient, err
 	} else if !exists {
```
nozim, line 45: I think this if can be flattened this way:
```
if err != nil {
  return dynamo ...
]

if !exists {

}
``` without else. Looks slightly more readable to me.

PR #96, opened 2023-10-19
Bpaas 1155 config for centralised storage

shotasilagadzetaal:


metamorph/store/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,461 @@
...
+	"github.com/opentracing/opentracing-go/ext"
+	"github.com/opentracing/opentracing-go/log"
+	"github.com/ordishs/gocore"
+)
+
+type DynamoDB struct {
+	dynamoCli *dynamodb.Client
+}
+
+const ISO8601 = "2006-01-02T15:04:05.999Z"
```
nozim, line 27: there is already a constant for time format `time.RFC3339`. Maybe worth using that instead? What do you think?

shotasilagadzetaal: Not sure,

From code `RFC3339     = "2006-01-02T15:04:05Z07:00"` So are they really same ? 🤔

nozim: you are right. It's RFC3339Nano rather.  
see here https://play.golang.com/p/xJuEOoEtHVF

PR #106, opened 2023-10-23
no need to check source

shotasilagadzetaal:


metamorph/server.go
```diff
@@ -337,7 +338,7 @@ func (s *Server) putTransactionInit(ctx context.Context, req *metamorph_api.Tran
 		return 0, 0, nil, nil, err
 	}
 
-	if rtr.Source != s.source {
+	if _, ok := s.store.(*dynamodb.DynamoDB); !ok && rtr.Source != s.source {
```
icellan, line 341: @shotasilagadzetaal Why is the database hard-coded here? Would it not be possible to handle this through the interface?

shotasilagadzetaal: The only way to avoid this is to change interface which I didn't want. Even after changing interface a new function would look ugly. `IsCentralised` is like unrelated to all other function in the interface so I preferred checking db locally. If you want me to change the interface I can.

boecklim: The function would look very simple like
```
func (ddb *DynamoDB) IsCentralised() bool {
	return true
}
```

This way no implementation details of the storage leak into the business logic.
Later we can remove this again because we can remove the code in the if clause.

boecklim (approved):

PR #108, opened 2023-10-23
Add structured logging to blocktx client

boecklim:
- Add structured logging to blocktx client
- Add unit tests
- Move `LoadArcHandler()` as the function is only used in the `cmd` package and it is not testable


blocktx/client.go
```diff
@@ -57,9 +74,9 @@ func (btc *Client) Start(minedBlockChan chan *blocktx_api.Block) {
 	for {
 		stream, err := btc.client.GetBlockNotificationStream(context.Background(), &blocktx_api.Height{})
 		if err != nil {
-			btc.logger.Errorf("Error getting block notification stream: %v", err)
+			btc.logger.Error("Error getting block notification stream", slog.String("err", err.Error()))
```
nozim, line 77: I think Errorf would fit better here. 
Also you don't have to explicitly call .Error() on error when passed to formatter.

boecklim: `slog` does not provide `Errorf()` function

nozim: I see. no problem then.

blocktx/client.go
```diff
@@ -26,55 +27,100 @@ type ClientI interface {
...
+					btc.logger.Error("failed to get block notification stream", slog.String("err", err.Error()))
+					continue
 				}
 
-				btc.logger.Infof("Block %s\n", utils.ReverseAndHexEncodeSlice(block.Hash))
-				utils.SafeSend(minedBlockChan, block)
+				btc.logger.Info("Connected to block-tx server")
+
+				var block *blocktx_api.Block
+				for {
```
nozim, line 97: I'm not sure nested channel processing inside another select is good idea.

boecklim: Can you explain your concern in more detail?
The first select waits for either the ticker or the shutdown to be triggered
The second select waits for either the receiving of the block to end or also the shutdown to be triggered.

In this way the `Shutdown()` function waits until `block, err = stream.Recv()` is finished.

nozim: 1. I've refactored the client start test in this branch https://github.com/bitcoin-sv/arc/blob/b182300b6efe04c9d7a358fb39b5cdaaf0876a99/blocktx/client.go#L70 

with single select. The result is exact same and much less confusing than nested selects.

nozim (approved):

PR #110, opened 2023-10-24
Sql tests2

nozim:
- adding more tests for block tx sql functions


blocktx/store/sql/insert_block_transactions.go
```diff
@@ -7,25 +7,14 @@ import (
...
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	qTx, err := s.db.Prepare(`
-			INSERT INTO transactions (hash, merkle_path) VALUES ($1, $2)
-			ON CONFLICT(hash) DO UPDATE SET merkle_path=$2
-			RETURNING id
-			;
-		`)
+	qTx, err := s.db.Prepare(`INSERT INTO transactions (hash, merkle_path) VALUES ($1, $2);`)
```
boecklim, line 17: With this change the transaction can not be "upserted" any more. That may be the reason why the e2e test fails.

nozim: e2e tests don't use embedded db. They have completely different flow.

database/migrations/postgres/000001_create_blocks.up.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TABLE blocks (
                         inserted_at TIMESTAMP WITH TIME ZONE  DEFAULT CURRENT_TIMESTAMP NOT NULL,
-                        id BIGINT NOT NULL,
+                        id SERIAL PRIMARY KEY,
```
boecklim, line 3: I think the migrations should reflect what we have here i.e. the `id           BIGSERIAL PRIMARY KEY`
https://github.com/bitcoin-sv/arc/blob/31aa8190dff210b87c12378adb7a67cc286aeb4e/blocktx/store/sql/sql.go#L173

Which is the schema of the postgres production db

blocktx/store/sql/sql_test.go
```diff
@@ -23,75 +26,70 @@ func TestInOut(t *testing.T) {
 		Height:       1,
 	}
 
-	s, err := New("sqlite_memory")
```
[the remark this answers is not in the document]
nozim: I'm not sure how to keep tests for both sqlite and postgres. 
The tests should not depend on database implementation.

database/schema.sql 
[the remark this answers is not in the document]
nozim: removed this. The idea initially was to have an option to run entire schema instead of migrations one by one. Might be useful at some point but for now will remove it.

blocktx/store/model.go
```diff
@@ -31,3 +31,8 @@ type BlockTransactionMap struct {
 	Pos           int64     `db:"pos"`
 	InsertedAt    time.Time `db:"inserted_at"`
 }
+
+type BlockTransactions struct {
```
boecklim, line 35: It seems this type is not used anywhere

blocktx/store/sql/get_block_test.go
```diff
@@ -17,7 +17,7 @@ type GetBlockTestSuite struct {
 	DatabaseTestSuite
 }
 
-func (s *GetBlockByHeightTestSuite) TestGetBlock() {
+func (s *GetBlockTestSuite) Test() {
```
boecklim, line 20: Here it is called `Test()` but everywhere else it is called `Run()`

database/migrations/postgres/000003_create_transactions.up.sql
```diff
@@ -1,5 +1,5 @@
 CREATE TABLE transactions (
-                              id BIGINT NOT NULL,
+                              id SERIAL PRIMARY KEY,
```
boecklim, line 2: Also here the original type of this column is BIGSERIAL 
https://github.com/bitcoin-sv/arc/blob/31aa8190dff210b87c12378adb7a67cc286aeb4e/blocktx/store/sql/sql.go#L200

blocktx/store/model.go
nozim: removed.

database/migrations/postgres/000003_create_transactions.up.sql
nozim: thanks, amended.

blocktx/store/sql/get_block_test.go
nozim: renamed all the methods to `Test`.

database/migrations/postgres/000001_create_blocks.up.sql
nozim: changed to bigserial.

boecklim (approved):

PR #111, opened 2023-10-24
txt endpoint tests

farzan1990:


test/arc_txt_endpoint_test.go
```diff
@@ -237,3 +238,174 @@ func TestHttpPost(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Error sending HTTP request: %s", err)
+	}
+	defer resp.Body.Close()
+
+
+	if resp.StatusCode != http.StatusOK {
+		t.Errorf("Expected 200 OK but got: %d", resp.StatusCode)
+	}
+}
```
boecklim, line 320: @farzan1990 this closing bracket is the issue

test/arc_txt_endpoint_test.go
```diff
@@ -237,3 +238,174 @@ func TestHttpPost(t *testing.T) {
...
...1697c76638c7c587f77517f7c01007e817602fc00a06302fd00a063546752687f7c01007e81687f777c6876638c7c587f77517f7c01007e817602fc00a06302fd00a063546752687f7c01007e81687f777c6863587f77517f7c01007e817602fc00a06302fd00a063546752687f7c01007e81687f7768587f517f7801007e817602fc00a06302fd00a063546752687f7801007e81727e7b7b687f75537f7c0376a9148801147f775379645579887567726881766968789263556753687a76026c057f7701147f8263517f7c766301007e817f7c6775006877686b537992635379528763547a6b547a6b677c6b567a6b537a7c717c71716868547a587f7c81547a557964936755795187637c686b687c547f7701207f75748c7a7669765880748c7a76567a876457790376a9147e7c7e557967041976a9147c7e0288ac687e7e5579636c766976748c7a9d58807e6c0376a9147e748c7a7e6c7e7e676c766b8263828c007c80517e846864745aa0637c748c7a76697d937b7b58807e56790376a9147e748c7a7e55797e7e6868686c567a5187637500678263828c007c80517e846868647459a0637c748c7a76697d937b7b58807e55790376a9147e748c7a7e55797e7e687459a0637c748c7a76697d937b7b58807e55790376a9147e748c7a7e55797e7e68687c537a9d547963557958807e041976a91455797e0288ac7e7e68aa87726d77776a14704c123b485438548ed86b42271f16f59147083f01000a544f4b454e54455354530754455354494e4737030000000000001976a9145eb2169818a110149787ed09a918c842aad8835288ac00000000"
```
nozim, line 270: i think this is best to be loaded from fixture file to keep the test a bit more readable. What do you think?

test/arc_txt_endpoint_test.go
```diff
@@ -237,3 +238,174 @@ func TestHttpPost(t *testing.T) {
...
+	jsonPayload := fmt.Sprintf(`{"rawTx": "%s"}`, txHexString)
+	resp, err := postTx(t, jsonPayload, nil)
+	if err != nil {
```
nozim, line 273: It's probably easer to use https://pkg.go.dev/github.com/stretchr/testify/require
like this:
``` require.NoError(t, err)
```

test/arc_txt_endpoint_test.go
```diff
@@ -237,3 +238,174 @@ func TestHttpPost(t *testing.T) {
...
+	}
+}
+
+	if resp.StatusCode != 466 {
+		t.Errorf("Expected 466 Conflicting transaction found but got: %d", resp.StatusCode)
+	}
+}
+
+func postTx(t *testing.T, jsonPayload string, headers map[string]string) (*http.Response, error) {
+	url := "http://arc:9090/v1/tx"
```
nozim, line 328: maybe worth extracting the arc IP to env variable to be able to run it outside docker container?

test/arc_txt_endpoint_test.go
farzan1990: Agreed :)

test/arc_txt_endpoint_test.go
farzan1990: Thanks I will commit shortly

test/arc_txt_endpoint_test.go
farzan1990: Good idea. I will put that in a separate PR for enhanchments as the other tests are currently working in this way.

test/arc_txt_endpoint_test.go
farzan1990: Done

test/arc_txt_endpoint_test.go
boecklim: In this case you could use `require.NoErrorf(t, err, "Error sending HTTP request")`

test/arc_txt_endpoint_test.go
```diff
@@ -237,3 +237,177 @@ func TestHttpPost(t *testing.T) {
...
+func TestPostTx_Success(t *testing.T) {
+	txHexString := createTxHexStringExtended(t) // This is a placeholder for the method to create a valid transaction string.
+	jsonPayload := fmt.Sprintf(`{"rawTx": "%s"}`, txHexString)
+	resp, err := postTx(t, jsonPayload, nil) // no extra headers
+	if err != nil {
+		t.Fatalf("Error sending HTTP request: %s", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
```
boecklim, line 250: simpler would be `require.Equal(t, http.StatusOK, resp.StatusCode`

boecklim (commented):
I think we should make use of the `require` package instead of having a lot of `if` statements in the tests

test/arc_txt_endpoint_test.go
nozim: yup, that's most thoughtful way. Thanks @farzan1990

nozim (approved):

PR #112, opened 2023-10-26
dynamo option in config

shotasilagadzetaal:


config.yaml
```diff
@@ -39,17 +39,17 @@ metamorph:
...
       host: localhost
       port: 5432
       name: metamorph
       user: arc
       password: arc
     dynamodb:
-      aws_access_key_id:
-      aws_secret_access_key:
-      aws_session_token:
+      awsAccessKeyID:
```
boecklim, line 50: This cannon be changed only here. It also needs to be changed where it is used by viper: https://github.com/bitcoin-sv/arc/blob/76bf601b0e02f1403fc1bab0c480442b9e8a0563/metamorph/store/dynamodb/dynamodb.go#L32

boecklim (approved):

PR #114, opened 2023-10-26
Refactored blocktx client

nozim:


blocktx/client.go
```diff
@@ -75,48 +68,37 @@ func NewClient(client blocktx_api.BlockTxAPIClient, opts ...Option) ClientI {
...
-							btc.logger.Error("Failed to receive block", slog.String("err", err.Error()))
-							break
-						}
-						btc.logger.Info("Block", slog.String("hash", utils.ReverseAndHexEncodeSlice(block.Hash)))
-						utils.SafeSend(minedBlockChan, block)
-					}
-				}
 			case <-btc.shutdown:
 				return
+			default:
```
boecklim, line 88: In the PR https://github.com/bitcoin-sv/arc/pull/108 I didn't want to change the logic, but just add `slog` and make it testable.

This PR changes the logic of the `Start` function. 

The original logic first gets the stream every `btc.retryTicker` interval and then iterates over the stream

Here we don't wait for incoming interval ticks `btc.retryTicker.C`  to first get the stream `btc.client.GetBlockNotificationStream` but the the connection is open all the time.

The question is if we should change the logic just for better readability. Need to check that first

nozim:
closing this as it's served it's purpose for discussion topic :)

PR #118, opened 2023-10-28
Add structured logging to main.go

boecklim:
Add structured logging to main

### DIFF
--- a/tracing/init_test.go
+++ b/tracing/init_test.go
@@ -1,6 +1,7 @@
 package tracing
 
 import (
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -10,6 +11,7 @@ func TestInit(t *testing.T) {
 	tt := []struct {
 		name        string
 		serviceName string
+		setEnv      bool
 
 		expectedErrorStr string
 	}{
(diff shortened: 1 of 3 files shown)


tracing/init_test.go
```diff
@@ -0,0 +1,43 @@
...
+			name:        "failed to initialize jaeger tracer",
+			serviceName: "",
+
+			expectedErrorStr: "cannot initialize jaeger tracer",
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			tracer, closer, err := InitTracer("")
```
nozim, line 30: isn't it better to just 

```
struct { 
// ....
ExpectedErr error 
}
```
and in test body 
```
assert.Equal(t, tc.ExpectedErr, err) 
```

Seems slightly more readable to me.

boecklim: Then I have to have in each test case the full error message including the one from the `jaeger` package in the one case e.g. 
`cannot parse jaeger env vars: cannot parse env var JAEGER_RPC_METRICS=not a boolean: strconv.ParseBool: parsing \"not a boolean\": invalid syntax`

Using `require.ErrorContains()` I can have there a substring. That is why I find it more useful.

shotasilagadzetaal (approved):

PR #122, opened 2023-10-31
Delete txs from storage after period of time

boecklim:
On loading unmined transactions delete those from storage which have been stored more than 2 weeks earlier

### DIFF
--- a/metamorph/processor.go
+++ b/metamorph/processor.go
@@ -83,6 +83,7 @@ const (
 	logLevelDefault                         = slog.LevelInfo
 
 	failedToUpdateStatus = "Failed to update status"
+	txDeletionPeriod     = 336 * time.Hour // 14 days
 )
 
 func WithProcessExpiredSeenTxsInterval(d time.Duration) func(*Processor) {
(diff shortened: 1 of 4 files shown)


metamorph/processor.go
```diff
@@ -312,6 +313,17 @@ func (p *Processor) LoadUnmined() {
 	defer span.Finish()
 
 	err := p.store.GetUnmined(spanCtx, func(record *store.StoreData) {
+
+		if p.now().Sub(record.StoredAt) > txDeletionPeriod {
```
nozim, line 317: 1. do we really need `time.Now` as another field in processor?
2. would it be more readable to extract this condition as function:

```
func isExpired(record *store.StoreData) bool {
  return time.Now().Sub(record.StoredAt) > txDeletonPeriod
}
```

and then 

```
if isExpired(record) {
 // handle deletion
}
``` 
What do you think?

metamorph/processor.go
```diff
@@ -84,6 +84,7 @@ const (
 	logLevelDefault                         = slog.LevelInfo
 
 	failedToUpdateStatus = "Failed to update status"
+	txDeletionPeriod     = 336 * time.Hour // 14 days
```
nozim, line 87: `14 * 24 * time,Hour` seems much much more readable to me.

shotasilagadzetaal:
As I understand we are removing transactions that have status `seen` on network and never mined right? 🤔

metamorph/processor.go
boecklim: 1. The field was added as `now: func() time.Time` so that you can set this function. By default the function is simply `time.Now()` but when you do tests you may want to control what time is returned. Then you can set it to `func() time.Time { return time.Date(...)}`. That is why it is a field
2. Agree. Will see how to best put it in a function

boecklim:
> As I understand we are removing transactions that have status `seen` on network and never mined right? 🤔

@shotasilagadzetaal 
Currently at startup umined txs get loaded from the storage. This can be a huge number of txs, as there is currently no place where they ever get deleted. So very old transactions which for some reason didn't get mined will be retried here.

Depending on the status of each transaction different actions occur. For seen transactions, `blocktx` service is asked if a block is stored with this tx. That creates a great cpu load on the DB.

With this PR any transaction which is older than 14 days are removed from the storage. That should remove the load on the DB for now. 
We will later need a follow-up PR which periodically removes txs after this period of time.

I think we should actually turn the 14 days into a setting, because users of ARC may want to store them longer.
I hope that answered your question.

nozim (approved):
@boecklim  Would it make sense to extract this clean up in background jobs?

boecklim:
> @boecklim Would it make sense to extract this clean up in background jobs?

Hey @nozim 
Yes, we should have that as well => `We will later need a follow-up PR which periodically removes txs after this period of time.`

PR #129, opened 2023-11-06
Fix for submitting mined transactions

shotasilagadzetaal:
Insert transaction first to update it's status to mined.

### DIFF
--- a/metamorph/server.go
+++ b/metamorph/server.go
@@ -401,6 +401,25 @@ func (s *Server) putTransactionInit(ctx context.Context, req *metamorph_api.Tran
 	}
 
 	if rtr.BlockHash != nil {
+		// In case a transaction is submited to network outside of ARC and mined and now
+		// submitting the same transaction through arc endpoint we have problem.
+		// The transaction doesn't exist in metamorph so the call after this line
+		// (updating tx status to MINED) will fail as there is no transaction to update.
+		// In that case we take the transaction from blocktx (where we parsed it from network)
+		// and store it first in metamorph database. BPAAS-1150
+
+		// In case this is "normal" transaction - meaning that we already
+		// have it in the database because it was submitted to network through arc
+		// this store call will fail insertion and the old functionality will still work as expected.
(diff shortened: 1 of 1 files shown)


metamorph/server.go
```diff
@@ -401,6 +401,22 @@ func (s *Server) putTransactionInit(ctx context.Context, req *metamorph_api.Tran
...
+		// In that case we take the transaction and store it first in metamorph database.
+		if err := s.store.Set(ctx, hash[:], &store.StoreData{
+			Hash:          &hash,
+			Status:        status,
+			CallbackUrl:   req.CallbackUrl,
+			CallbackToken: req.CallbackToken,
+			MerkleProof:   req.MerkleProof,
+			RawTx:         req.RawTx,
+		}); err != nil {
+			s.logger.Warn("Transaction already exists in db", slog.String("warn", err.Error()))
```
nozim, line 417: 1. Isn't it suposed to be ".Error" instead of Warn?
2. If the store is dynamodb the errors in store.Set can be because of `MarshalMap` and `dynamoCli.PutItem` so not sure the message is correct. 
The error might be not because of existing record.

shotasilagadzetaal: If we have the transaction already there it will return error but it's not really error for us. 

I'll check if there is anything to be cautious in dynamodb implmenentation.

boecklim: I guess the error may be something else than transaction already existing, not?

I'd put there something like this: `s.logger.Warn("Failed to store transaction", slog.String("hash", hash.String()), slog.String("warn", err.Error()))`

nozim: yes but it's still error and not warn. Because you are handling error not warning.

boecklim: Ok, then rather this I guess
```go
s.logger.Error("Failed to store transaction", slog.String("hash", hash.String()), slog.String("err", err.Error()))
```

boecklim (approved):

PR #130, opened 2023-11-06
Use IRSA for metamorph

taal-martins-eglitis:
Use the default config via IRSA, update the Dockerfile, use context passed by args


metamorph/store/dynamodb/dynamodb.go
```diff
@@ -28,24 +26,11 @@ type DynamoDB struct {
 const ISO8601 = "2006-01-02T15:04:05.999Z"
 
 func New() (store.MetamorphStore, error) {
-	// export necessary parameters for aws dynamodb connection
-	err := os.Setenv("AWS_ACCESS_KEY_ID", viper.GetString("metamorph.db.dynamodb.awsAccessKeyID"))
```
boecklim, line 32: If we don't need this, we can also remove it from the default config file: https://github.com/bitcoin-sv/arc/blob/c64f83a63a82bc0a446661b929d522abc14be4d6/config.yaml#L50

taal-martins-eglitis: Yes, indeed. Will update.

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM ubuntu:latest
 
+RUN apt-get update && apt-get install ca-certificates -y && apt-get clean && rm -rf /var/lib/apt/lists/*
```
boecklim, line 3: Possibly we can do this using this package like this: https://github.com/breml/rootcerts
```go
import (
    _ "github.com/breml/rootcerts"
)
```

But we can also do that in a separate PR

boecklim (approved):

PR #131, opened 2023-11-06
Fix/callback blockinfo

boecklim:
- Store block infor in callbacker storage
- e2e test for checking that callback has expected body and authorization header with 2 iterations
  - 1st iteration fails => callbacker updates expiry and loads expired callbacks from storage
  - 2nd iteration succeeds


nozim (approved):
lgtm

PR #136, opened 2023-11-07
Remove irrelevant tests and files,  refactor processor request

nozim:
- replace manually created mock objects with moq library alternative 
- refactor metamorph server test
- remove context.Context from ProcessorRequest as per context usage best practices 
https://go.dev/blog/context-and-structs


metamorph/processor_request.go
boecklim, line 1: I think this type and function was in the right place here as the other file is called `interface.go`

metamorph/processor_request_test.go
boecklim, line 1: This test was not super useful, but also didn't hurt. Without it 1 line less is test covered

metamorph/processor_test.go
```diff
@@ -621,15 +621,14 @@ func BenchmarkProcessTransaction(b *testing.B) {
 
 		txs[txID] = &txHash
 
-		processor.ProcessTransaction(NewProcessorRequest(
-			context.Background(),
-			&store.StoreData{
+		processor.ProcessTransaction(&ProcessorRequest{
```
boecklim, line 624: Why not use the constructor as before `NewProcessorRequest`

nozim: the constructor does not actually do anything except initialising structure and the arguments seem less readable. 
Also you don't have to specify default value for channels for example.

nozim: also from 
https://pkg.go.dev/context:
"Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:"

metamorph/processor_mock.go
boecklim, line 1: As this file is generated it should not be edited by hand
To update this file use `make install_gen` and `make gen_go`

boecklim (approved):

PR #144, opened 2023-11-13
Feature/metamorph store lockedby

boecklim:
- New field to metamorph storage denoting which host is currently exclusively processing a record
- Function which to remove the lock from a record
- Unit tests for DynamoDB storage implementation
- Updated readme on how to run tests


cmd/metamorph.go
```diff
@@ -309,13 +319,47 @@ func StartMetamorph(logger utils.Logger) (func(), error) {
...
 		logger.Infof("Shutting down metamorph store")
+		metamorphProcessor.Shutdown()
 		err = s.Close(context.Background())
 		if err != nil {
 			logger.Errorf("Could not close store: %v", err)
 		}
 	}, nil
 }
 
+func NewStore(dbMode string, folder string) (s store.MetamorphStore, err error) {
```
nozim, line 330: it's not quite obvious that default storage type is SQL here. Maybe worth keeping all the options as an argument instead?
Or document each of existing dbModes?

metamorph/processor.go
```diff
@@ -209,6 +214,17 @@ func (p *Processor) Shutdown() {
 	}
 }
 
+func (p *Processor) unlockItems() error {
+	items := p.processorResponseMap.Items()
+	hashes := make([]*chainhash.Hash, len(items))
+	index := 0
+	for key := range items {
+		hashes[index] = &key
```
nozim, line 222: you are setting hashes[0] with new keys throughout the loop here. 
probably you meant 
```
for index, key := range items {
  hashes[index] = &key
}
``` 
?)

boecklim: This I can't do because it is a map, so the first parameter is not an index.

But you're right, I forgot thi `index++`

metamorph/store/dynamodb/dynamodb.go
```diff
@@ -19,62 +19,70 @@ import (
 	"github.com/ordishs/gocore"
 )
 
+const (
+	ISO8601      = "2006-01-02T15:04:05.999Z"
```
nozim, line 23: Why do we need to keep this custom format everywhere?
What is wrong with `RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"`

https://pkg.go.dev/time#:~:text=The%20RFC3339Nano%20format%20removes%20trailing,creating%20a%20custom%20layout%20string.

cmd/metamorph.go
boecklim: Will revisit that

metamorph/store/dynamodb/dynamodb_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	t.Cleanup(func() {
+		err := pool.Purge(resource)
+		require.NoError(t, err)
+	})
+
+	resolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+		return aws.Endpoint{
+			PartitionID:   "aws",
+			URL:           "http://localhost:" + resource.GetPort("8000/tcp"),
```
nozim, line 60: maybe worth extracting this as const? 
it will be hard to maintain this tests in docker setup this way.

metamorph/store/dynamodb/dynamodb.go
```diff
@@ -416,7 +473,7 @@ func (ddb *DynamoDB) GetBlockProcessed(ctx context.Context, blockHash *chainhash
 
 	var processedAtTime time.Time
 	if blockItem.ProcessedAt != "" {
-		processedAtTime, err = time.Parse(ISO8601, blockItem.ProcessedAt)
+		processedAtTime, err = time.Parse(time.RFC3339, blockItem.ProcessedAt)
```
nozim, line 476: time.RFC3339Nano is probably better fit if you need nano precision as per original format. But up to you ;)

nozim (approved):
🚢

shotasilagadzetaal (approved):

PR #153, opened 2023-11-16
Feature/unlock by name

boecklim:
New metamorph functionality which allows to unlock records in a centralised storage in case of a non-graceful shutdown where metamorph will not unlock the records by it's own.


nozim (approved):
lgtm

PR #154, opened 2023-11-20
Add e2e test for merkle path in BUMP format

boecklim:
New e2e test checks for correctness of the Merkle path after tx has been mined


sirdeggen (approved):
Let's merge it.

PR #155, opened 2023-11-20
Fix/update api docu

boecklim:
- update error type links
- add possible error responses for the different endpoints
- add missing error types

### DIFF
--- a/api/arc.go
+++ b/api/arc.go
@@ -160,6 +160,21 @@ type ErrorFrozenPolicy struct {
 	Type interface{} `json:"type"`
 }
 
+// ErrorGeneric defines model for ErrorGeneric.
+type ErrorGeneric struct {
+	Detail interface{} `json:"detail"`
+
+	// ExtraInfo Optional extra information about the error from the miner
+	ExtraInfo *string      `json:"extraInfo"`
+	Instance  *interface{} `json:"instance,omitempty"`
+	Status    interface{}  `json:"status"`
+	Title     interface{}  `json:"title"`
+
+	// Txid Transaction ID this error is referring to
+	Txid *string     `json:"txid"`
+	Type interface{} `json:"type"`
+}
+
 // ErrorInputs defines model for ErrorInputs.
 type ErrorInputs struct {
 	Detail interface{} `json:"detail"`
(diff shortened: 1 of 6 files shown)


api/arc.go
```diff
@@ -190,6 +190,51 @@ type ErrorMalformed struct {
 	Type interface{} `json:"type"`
 }
 
+// ErrorNotFound defines model for ErrorNotFound.
+type ErrorNotFound struct {
```
nozim, line 194: you can reuse fields from `ErrorGeneric` in `ErrorNotFound` and others simply by embedding to avoid duplicating exact same set of fields:

```
type ErrNotFound {
   ErrorGeneric 
}

type ErrorOutputs struct {
  ErrorGeneric 
}
```

or by type aliases:

```
type ErrorNotFound ErrorGeneric 
type ErrorOutputs ErrorGeneric 
// and so on ...
```
However, option 2 assumes they have exact same set of fields so I guess former example is more flexible.

api/arc.go
```diff
@@ -1416,26 +1476,61 @@ func ParsePOSTTransactionResponse(rsp *http.Response) (*POSTTransactionResponse,
 		}
 		response.JSON400 = &dest
 
+	case strings.Contains(rsp.Header.Get("Content-Type"), "json") && rsp.StatusCode == 409:
```
nozim, line 1479: I guess it's better to use constant `http.StatusConflict` instead of "magic number" here. What do you think ?

Ref: https://go.dev/src/net/http/status.go (seek StatusConflict)

api/arc.go
```diff
@@ -1416,26 +1476,61 @@ func ParsePOSTTransactionResponse(rsp *http.Response) (*POSTTransactionResponse,
...
-		var dest ErrorConflict
+	case strings.Contains(rsp.Header.Get("Content-Type"), "json") && rsp.StatusCode == 463:
+		var dest ErrorMalformed
 		if err := json.Unmarshal(bodyBytes, &dest); err != nil {
 			return nil, err
 		}
-		response.JSON466 = &dest
+		response.JSON463 = &dest
+
+	case strings.Contains(rsp.Header.Get("Content-Type"), "json") && rsp.StatusCode == 464:
```
nozim, line 1521: 1. why do we need to cover all the status codes, even those obscure not listed in net/http package? 
2. same magic number as in example above. Can you please define what exact status code you are trying to cover?

nozim (approved):
left some comments, the rest looks good.

api/arc.go
boecklim: This code I did not edit. `arc.go` is generated from `arc.yml`.

api/arc.go
boecklim: This code I did not edit. arc.go is generated from arc.yml.

api/arc.go
boecklim: What I basically did is just to add those errors which ARC in fact returns to the API definition to make it complete. The status codes were already there defined here `api/status.go`

PR #157, opened 2023-11-20
add new status and return if tx is orphaned

shotasilagadzetaal:


metamorph/metamorph_api/metamorph_api.proto
```diff
@@ -17,7 +17,8 @@ enum Status {
   SENT_TO_NETWORK = 6;
   ACCEPTED_BY_NETWORK = 7;
   SEEN_ON_NETWORK = 8;
-  MINED = 9;
+  ORPHANED = 9;
+  MINED = 10;
```
boecklim, line 21: This will break the existing tx statuses that are stored, but If we introduce a new status we won't get around it.

As we are already changing the statuses how about using a much higher number for MINED in case we'll have yet more statuses in the future?

e.g.  `MINED=50`

If this is the order of statuses then `ORPHANED` we maybe should move it before `ACCEPTED_BY_NETWORK` because an orphaned tx has been sent to network but was not accepted by network. Like `ACCEPTED_BY_NETWORK` status, the orphaned status is received by zmq.

So I'd put the order in this way.
```
ORPHANED=7;
ACCEPTED_BY_NETWORK = 8;
SEEN_ON_NETWORK = 9;

MINED=50;
```

What do you think?

sirdeggen: Hold on. Let's talk about this. Transactions cannot be orphaned. Only Blocks can be orphaned. No such thing as an orphan transaction.

sirdeggen: The prior numeric value is probably completely irrelevant since users don't see this anyway.

Perhaps we should use these numbers in the Response - since it's better than parsing `txStatus` on the client side, and allows users to set `if txStatusCode >= 7` in their codebase for example.

sirdeggen: Having said all that - can we define "ORPHANED"?

sirdeggen: > The transaction has been sent to at least 1 Bitcoin node but parent transaction was not found.

sirdeggen: Impossible situation - why are we broadcasting transactions for which there is no parent transaction?

README.md
```diff
@@ -159,9 +159,10 @@ available:
...
-| 9   | `MINED`                | The transaction has been mined into a block by a mining node.                                                                                                                                            |
+| 8   | `ACCEPTED_BY_NETWORK`  | The transaction has been accepted by a connected Bitcoin node on the ZMQ interface. If metamorph is not connected to ZMQ, this status will never by set.                                                 |
+| 9   | `SEEN_ON_NETWORK`      | The transaction has been seen on the Bitcoin network and propagated to other nodes. This status is set when metamorph receives an INV message for the transaction from another node than it was sent to. |
+| 10   | `SEEN_IN_ORPHAN_MEMPOOL`             | The transaction has been sent to at least 1 Bitcoin node but parent transaction was not found. |
+| 50   | `MINED`                | The transaction has been mined into a block by a mining node.                                                                                                                                            |
```
boecklim, line 165: These numbers are wrong. Should be like this
```
  ACCEPTED_BY_NETWORK = 7;
  SEEN_ON_NETWORK = 8;
  MINED = 9;
```

boecklim (approved):

PR #166, opened 2023-11-23
Test/chained txs e2e

boecklim:
- Do not interpret metamorph status integer as hierarchy
- Added e2e test to ensure batch of chained transactions get mined

### DIFF
--- a/metamorph/server.go
+++ b/metamorph/server.go
@@ -309,6 +309,26 @@ func (s *Server) PutTransactions(ctx context.Context, req *metamorph_api.Transac
 	return resp, nil
 }
 
+func returnStatus(status metamorph_api.Status, waitForStatus metamorph_api.Status) bool {
+	statusValueMap := map[metamorph_api.Status]int{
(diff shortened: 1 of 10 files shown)


metamorph/server.go
```diff
@@ -309,7 +309,27 @@ func (s *Server) PutTransactions(ctx context.Context, req *metamorph_api.Transac
 	return resp, nil
 }
 
-func (s *Server) processTransaction(ctx context.Context, waitForStatus metamorph_api.Status, data *store.StoreData, TxID string, latestStatus metamorph_api.Status) *metamorph_api.TransactionStatus {
+func hasWaitForStatusReached(status metamorph_api.Status, waitForStatus metamorph_api.Status) bool {
```
shotasilagadzetaal, line 312: I don't think those statuses are correct. We have new orphan status with 10.

boecklim: These values do not reflect the values of the metamorph api in any way. This function merely determines at which status the response can be returned to the user for that particular tx.
Does it make sense?

PR #169, opened 2023-11-24
Fix/dynamodb store

boecklim:
- Save Stored at date in DynamoDB
- More log entries when shutting down metamorph
- Listen for SIGTERM signal issued by K8s at pod termination to initiate graceful shutdown of pod

### DIFF
--- a/cmd/metamorph/main.go
+++ b/cmd/metamorph/main.go
@@ -6,6 +6,7 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"os/signal"
+	"syscall"
 
 	"github.com/bitcoin-sv/arc/cmd"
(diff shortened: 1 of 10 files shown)


cmd/metamorph/main.go
```diff
@@ -55,11 +56,11 @@ func main() {
 	// setup signal catching
 	signalChan := make(chan os.Signal, 1)
 	signal.Notify(signalChan, os.Interrupt)
```
shotasilagadzetaal, line 58: maybe for conformity instead of `os.Interrupt` we could have `syscall.SIGINT` (they are the same)

shotasilagadzetaal (approved):

PR #170, opened 2023-11-24
Feature/k8s pod watcher

boecklim:
- K8s-Watcher service for running along `metamorph` for unlocking records if `metamorph` terminates ungracefully
- Unit tests
- Documentation


api/transactionHandler/metamorph.go
```diff
@@ -29,7 +29,21 @@ type Metamorph struct {
 }
 
 // NewMetamorph creates a connection to a list of metamorph servers via gRPC
-func NewMetamorph(targets string, blockTxClient blocktx.ClientI, grpcMessageSize int, logger utils.Logger) (*Metamorph, error) {
+func NewMetamorph(address string, blockTxClient blocktx.ClientI, grpcMessageSize int, logger utils.Logger) (*Metamorph, error) {
+	conn, err := GetConnection(address, grpcMessageSize)
+	if err != nil {
+		return nil, err
```
nozim, line 35: maybe wrap this into additional message like "Unable to get connection due to ", err or something like that ? )

main.go
```diff
@@ -208,6 +212,16 @@ func main() {
 		}
 	}
 
+	if startK8sWatcher != nil && *startK8sWatcher {
+		if apiShutdown, err := cmd.StartK8sWatcher(logger); err != nil {
```
nozim, line 216: I think the execution should stop if the service is not running. Otherwise it will be hard to troubleshoot if it's ignored.

boecklim: You mean the service should always be running?
For local execution in one instance this will not be needed. It is only needed when running a decentralised storage and ARC running in a K8s cluster.

main.go
```diff
@@ -208,6 +212,16 @@ func main() {
...
+			logger.Error("Error starting k8s-watcher server", slog.String("err", err.Error()))
```
nozim, line 217: sorry for confusion, I meant that here it should return but it does not. If you don't look the logs carefully you may not know that K8sWatcher did not even started. I'm not sure that's what is actually intended.

nozim (approved):

PR #177, opened 2023-11-28
Blocktx centralisation

shotasilagadzetaal:


blocktx/store/interface.go
```diff
@@ -20,6 +20,8 @@ var (
 
 type Interface interface {
 	RegisterTransaction(ctx context.Context, transaction *blocktx_api.TransactionAndSource) (string, string, []byte, uint64, error)
+	TryToBecomePrimary(ctx context.Context, myHostName string) error
```
boecklim, line 23: Please add `//go:generate moq -out ./store/mock.go ./store Interface` to some where peer_handler_test.go so that in the future running `go generate ./...` is enough to recreate the mocks from a change in the interface

blocktx/peer_handler.go
```diff
@@ -275,6 +276,20 @@ func (bs *PeerHandler) HandleBlock(wireMsg wire.Message, peer p2p.PeerI) error {
...
+		return err
+	}
+
+	hostName, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+
+	if primaryBlocktx != hostName {
+		return nil
```
boecklim, line 290: Here an Info log could be added that this `blocktx` is not primary so it will not do anything.

blocktx/peer_handler.go
```diff
@@ -275,6 +276,20 @@ func (bs *PeerHandler) HandleBlock(wireMsg wire.Message, peer p2p.PeerI) error {
 		gocore.NewStat("blocktx").NewStat("HandleBlock").AddTime(start)
 	}()
 
+	primaryBlocktx, err := bs.store.PrimaryBlocktx(context.Background())
+	if err != nil {
+		return err
+	}
+
```
boecklim, line 283: The checking if the `blocktx` is primary should happen earlier I think in `HandleBlockAnnouncement`. This would then avoid that a block is even requested by this `utils.SafeSend(bs.workerCh, pair)`

shotasilagadzetaal: Yes but `HandleBlock` is called "from outside" right? So we need to have check there. No?

boecklim: I think `HandleBlock` will not be called for a particular when `blocktx` does not request the block. But yes, let's have it in both places.

blocktx/store/interface.go
shotasilagadzetaal: You mean to add this in Make file?

blocktx/peer_handler.go
```diff
@@ -238,6 +239,21 @@ func (bs *PeerHandler) HandleTransaction(msg *wire.MsgTx, peer p2p.PeerI) error
 }
 
 func (bs *PeerHandler) HandleBlockAnnouncement(msg *wire.InvVect, peer p2p.PeerI) error {
+	primaryBlocktx, err := bs.store.PrimaryBlocktx(context.TODO())
```
boecklim, line 242: If this whole block is in a function, then it can be reused in both `HandleBlockAnnouncement` and `HandleBlock`
```
primaryBlocktx, err := bs.store.PrimaryBlocktx(context.TODO())
	if err != nil {
		return err
	}

	hostName, err := os.Hostname()
	if err != nil {
		return err
	}

	if primaryBlocktx != hostName {
		bs.logger.Infof("Not primary, skipping block processing")
		return nil
	}
```

blocktx/store/sql/primary_blocktx.go
```diff
@@ -25,16 +25,18 @@ func (s *SQL) PrimaryBlocktx(ctx context.Context) (string, error) {
...
 	var hostName string
 
 	for rows.Next() {
 		err = rows.Scan(&hostName)
 		if err != nil {
 			return "", err
 		}
-
-		return hostName, nil
+		if ind == 0 {
```
nozim, line 36: ind is by default 0 so the loop will exit on first iteration. Is that intended?

shotasilagadzetaal: Yes, otherwise the linter complained 🤷

nozim (approved):
🚀

boecklim (approved):

PR #179, opened 2023-11-28
BAARC-44: check is flag passed

boecklim:
- Check if k8s-watcher flag is passed, so that k8s-watcher can be started individually


nozim (approved):
:ship ;)

PR #181, opened 2023-11-29
Fix/load unmined

boecklim:
- Do not delete expired with centralised storage, as it has an automatic mechanism for that (ttl)
- Set deletion log entry to loglevel Debug

### DIFF
--- a/metamorph/processor.go
+++ b/metamorph/processor.go
@@ -300,8 +300,12 @@ func (p *Processor) LoadUnmined() {
 	defer span.Finish()
 
 	err := p.store.GetUnmined(spanCtx, func(record *store.StoreData) {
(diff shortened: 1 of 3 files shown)


metamorph/processor.go
```diff
@@ -300,8 +300,12 @@ func (p *Processor) LoadUnmined() {
 	defer span.Finish()
 
 	err := p.store.GetUnmined(spanCtx, func(record *store.StoreData) {
-		if p.deleteExpired(record) {
-			return
+
+		if !p.store.IsCentralised() {
```
nozim, line 304:
```
if !p.store.IsCentralised() || p.deleteExpired(record) {
  return
}
``` ?

nozim (approved):

PR #183, opened 2023-11-30
add stats and garbage collection between runs. free up memory properly

shotasilagadzetaal:


blocktx/peer_handler.go
```diff
@@ -27,7 +28,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-const transactionStoringBatchsizeDefault = 16384 // power of 2 for easier memory allocation
+const transactionStoringBatchsizeDefault = 65536 // power of 2 for easier memory allocation
```
boecklim, line 31: The batch size has increased. Shouldn't it rather decrease so that the number of txs and merk paths held in memory will be smaller?

shotasilagadzetaal: I believe that this is not the reason of the failure. From that perspective, I would like to confirm that by increasing the number of transactions there. 

Even if that was the problem I would want to cause this failure much more often this way to quickly resolve it.

PR #191, opened 2023-12-01
fix: update go-bc solves BUMP related memory exhaustion

sirdeggen:
When BUMPs are being created, it was encoding the whole tree for every transaction. This small update ensures that only the relevant hashes are stored and the BUMP format returned is minimal. This should Massively improve performance issues associated with large memory usage.


sirdeggen:
https://github.com/libsv/go-bc/pull/81/files

For code change overview. Merged there already because the problem is obvious in hindsight - the function was left unfinished.

PR #195, opened 2023-12-04
Return error if setting for data record retention days is not found

boecklim:
- Return error if setting for data record retention days is not found

### DIFF
--- a/cmd/metamorph.go
+++ b/cmd/metamorph.go
@@ -38,9 +38,8 @@ import (
 )
 
 const (
-	DbModeBadger               = "badger"
-	DbModeDynamoDB             = "dynamodb"
-	recordRetentionDaysDefault = 14
+	DbModeBadger   = "badger"
+	DbModeDynamoDB = "dynamodb"
 )
 
 func StartMetamorph(logger utils.Logger) (func(), error) {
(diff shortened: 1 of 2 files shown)


cmd/metamorph.go
```diff
@@ -358,17 +362,17 @@ func NewStore(dbMode string, folder string) (s store.MetamorphStore, err error)
...
+			if err != nil {
+				return nil, err
 			}
 
 			tableNameSuffix := viper.GetString("metamorph.db.dynamoDB.tableNameSuffix")
 
 			s, err = dynamodb.New(
 				awsdynamodb.NewFromConfig(cfg),
 				hostname,
-				time.Duration(ttlDays)*24*time.Hour,
```
nozim, line 371: it does not have to be converted to time.Duration. You can just write 
`dataRetentionDays*24*time.Hour`. Seems much more readable to me.

nozim (approved):

cmd/metamorph.go
boecklim: It seems the conversion is needed
```
go build ./...
# github.com/bitcoin-sv/arc/cmd
cmd/metamorph.go:375:5: invalid operation: dataRetentionDays * 24 * time.Hour (mismatched types int and "time".Duration)
```

PR #196, opened 2023-12-04
List pods to check if k8s client is working

boecklim:
- Currently events are not received from pod watcher
- Need to ensure that client works
- List pods to check if k8s client is working


k8s_watcher/watcher.go
```diff
@@ -63,13 +65,26 @@ func (c *Watcher) Start() error {
 		return fmt.Errorf("failed to get pod watcher: %v", err)
 	}
 
+	getPodsTicker := time.NewTicker(time.Second * 10)
```
nozim, line 68: it's MUCH more readable when constant is before second like this:
10*time.Second not other way around.

nozim (approved):

PR #198, opened 2023-12-06
Fix/status updates

boecklim:
- Introduce status value map. The higher the value, the more information we have on whether a transaction with that status will be mined
- Do not overwrite a status with a lower or equal value status


nozim (approved):
Lgtm

PR #206, opened 2023-12-09
Db/metamorph postgres

boecklim:
- Separation of SqLite and Postgres implementation for metamorph
- integration tests for Postgres implementation
- e2e test using Postgres implementation
- index for deletion of records
- Remove duplicate setting of record


metamorph/server.go
```diff
@@ -279,10 +279,6 @@ func (s *Server) PutTransactions(ctx context.Context, req *metamorph_api.Transac
 			RawTx:         txReq.RawTx,
 		}
 
-		if err = s.processor.Set(ctx, &ProcessorRequest{Data: sReq}); err != nil {
```
shotasilagadzetaal, line 282: what is this all about ?

shotasilagadzetaal (approved):

metamorph/server.go
boecklim: I found that this setting is not needed as it already happens here: https://github.com/bitcoin-sv/arc/blob/f0c4abc631fcba488ba80df889496694f3bf8982/metamorph/processor.go#L529

With badger implementation the setting twice didn't raise an error, but postgres complained about the duplicate primary key, that is how I noticed it.

PR #207, opened 2023-12-10
New indexed field inserted_at_num in blocktx db

boecklim:
Adding an index on a timestamp of insertion is needed for eventual deletion of the records. Otherwise deletions will be very slow when a lot of records exist. An index on a timestamp turns out to have a penalty on performance, which can be seen when e2e tests fail in this case. Thus a new field was added which represents the hour when a record was added.

- New indexed field inserted_at_num in blocktx db which represents the hour when a record was added as an integer
- Update integration tests with new indexed field
- Change log entry `Status not updated for tx` to debug log level


background_worker/jobs/clear_block_transactions_map.go
```diff
@@ -9,23 +9,24 @@ import (
...
 	conn, err := sqlx.Open(params.Scheme(), params.String())
 	if err != nil {
 		Log(ERROR, fmt.Sprintf("unable to create connection %s", err))
 		return err
 	}
-	interval := fmt.Sprintf("%d days", params.RecordRetentionDays)
 
-	stmt, err := conn.Preparex("DELETE FROM block_transactions_map WHERE inserted_at <= (CURRENT_DATE - $1::interval)")
+	start := c.now()
+	datenumRetentiondDays := start.Add(-24 * time.Hour * time.Duration(params.RecordRetentionDays))
```
nozim, line 21: RecordRetentionDays is already in Days so this parameter seems redundant to me.

background_worker/jobs/clear_block_transactions_map_test.go
```diff
@@ -24,36 +25,30 @@ func (s *ClearBlockTransactionsMapSuite) Test() {
...
+		testfixtures.Directory("fixtures"), // The directory containing the YAML files
+	)
+	require.NoError(s.T(), err)
 
-	db, err := sqlx.Open("postgres", params.String())
+	err = fixtures.Load()
 	require.NoError(s.T(), err)
 
-	err = ClearBlockTransactionsMap(params)
+	now := time.Date(2023, 12, 22, 12, 0, 0, 0, time.UTC)
```
nozim, line 41: you don't need to manipulate date inside ClearJob. 
You simply insert "old" records into db and make sure it's deleted after job is complete.

background_worker/jobs/clear_blocks.go
```diff
@@ -10,29 +10,56 @@ import (
...
 		Log(ERROR, "unable to create connection")
 		return err
 	}
-	interval := fmt.Sprintf("%d days", params.RecordRetentionDays)
 
-	stmt, err := conn.Preparex("DELETE FROM blocks WHERE inserted_at <= (CURRENT_DATE - $1::interval)")
+	start := c.now()
+	datenumRetentiondDays := start.Add(-24 * time.Hour * time.Duration(params.RecordRetentionDays))
+
+	stmt, err := conn.Preparex("DELETE FROM blocks WHERE inserted_at_num <= $1::int")
```
nozim, line 56: what is inserted_at_num and how does it differ from inserted_at? Why we can't use inserted_at?

database/migrations/blocktx/postgres/000005_add_inserted_at.up.sql
```diff
@@ -1,5 +1,12 @@
+ALTER TABLE blocks
+ADD COLUMN inserted_at_num INTEGER DEFAULT TO_NUMBER(TO_CHAR((NOW()) AT TIME ZONE 'UTC', 'yyyymmddhh24'), '9999999999') NOT NULL;
```
nozim, line 2: Why do we need this column?

shotasilagadzetaal:
Probably too late but we could add block num there and remove blocks below specific height.

database/migrations/blocktx/postgres/000005_add_inserted_at.up.sql
shotasilagadzetaal: Will this default value correctly convert to hours?

shotasilagadzetaal (approved):
We could also use timestamp (by changing the format to unix time, same as integer and index it) and not add a new column

database/migrations/blocktx/postgres/000005_add_inserted_at.up.sql
boecklim: In my test it converted to hours. When I do this for example:
`SELECT TO_NUMBER(TO_CHAR((NOW()) AT TIME ZONE 'UTC', 'yyyymmddhh24'), '9999999999');`
I get
`2023121112`

PR #209, opened 2023-12-11
Add 2 minute period unmined txs processing

nozim:
- add periodic unmined transactions processing


cmd/metamorph.go
```diff
@@ -42,6 +42,10 @@ const (
 	DbModeDynamoDB = "dynamodb"
 )
 
+var (
+	unminedTxsPeriod = 2 * time.Minute
```
boecklim, line 46: This could be a constant like here: https://github.com/bitcoin-sv/arc/blob/f2dc16a3b1700d3ba179537957e269c72608132d/blocktx/client.go#L35

nozim: done

boecklim (approved):

PR #211, opened 2023-12-12
fix/get tx blocks for accepted txs

boecklim:
- Include txs with status ACCEPTED_BY_NETWORK in filter for checking against transaction blocks


boecklim:
Closing this PR and will create a new one with cleaned commits

PR #217, opened 2023-12-17
Add metamorph records clearing

nozim:
* add background job for clearing expired transactions in metamorph tables


database/migrations/metamorph/postgres/000001_create_transactions.up.sql
boecklim, line 1: If you copy these files from this path: metamorph/store/postgresql/migrations
then they should be exactly the same as they are the migration files which are actually used in the deployment. Right now they're not the same

boecklim: The migrations are still not exaclty the same. This schema is missing
https://github.com/bitcoin-sv/arc/blob/50a824542e1e9eade7357080e3be9f00f05af3a5/metamorph/store/postgresql/migrations/000001_create_transactions.up.sql#L1

background_worker/jobs/clear_metamorph.go
```diff
@@ -0,0 +1,42 @@
...
+		Log(ERROR, "unable to delete block rows")
+		return err
+	}
+	rows, _ := res.RowsAffected()
+	Log(INFO, fmt.Sprintf("Successfully deleted %d rows", rows))
+	return nil
+}
+
+func ClearMetamorph(params ClearRecordsParams) error {
+	if err := runDelete("blocks", params); err != nil {
```
boecklim, line 34: After adding the schema the from clause would be `metamorph.blocks` like here
https://github.com/bitcoin-sv/arc/blob/fbe226552c2e042b825bd50b90e0a4f3704ab837/metamorph/store/postgresql/postgres.go#L603

background_worker/jobs/clear_metamorph.go
```diff
@@ -0,0 +1,42 @@
...
+		return err
+	}
+	if err := runDelete("transactions", params); err != nil {
```
boecklim, line 37: The same here
https://github.com/bitcoin-sv/arc/blob/fbe226552c2e042b825bd50b90e0a4f3704ab837/metamorph/store/postgresql/postgres.go#L74

database/migrations/metamorph/postgres/000001_create_transactions.up.sql
nozim: why does it have to have custom schema?

background_worker/jobs/clear_metamorph.go
nozim: I'm not sure it's a good idea to hardcode database name inside the query. Because if the name inside db connection can be different from the one inside the query.

background_worker/jobs/clear_metamorph.go
nozim: same here, why do we need to hardcode the database name?

PR #229, opened 2024-01-03
fix parameters for update

shotasilagadzetaal:


boecklim (commented):
Looks good. 

Could you add here a integration test for that function?
`metamorph/store/postgresql/postgres_test.go`

boecklim (approved):

PR #251, opened 2024-01-12
Fix/resubmit tx

boecklim:
- Check if tx already exists in store
  - If yes, return that status
- Cover fix by e2e test


sirdeggen (approved):
lgtm 

[This line](https://github.com/bitcoin-sv/arc/blob/95cc851334a166a672f8b3d03e22d1f99b6ddf92/metamorph/processor.go#L462) is where I suspect it's tripping, so checking before we get there I think would indeed work.

PR #254, opened 2024-01-15
Refactor/blocktx enrich block info

boecklim:
- Register transaction in blocktx
- Do not insert transactions in blocktx but just update previously registered transactions


shotasilagadzetaal (approved):
Searching every new transaction in the database to see if it's ours (registered) might be too slow after some time. Not critical for now

PR #259, opened 2024-01-18
Refactor/rm unused blocktx funcs

boecklim:
- Remove unused functions from blocktx client
- Remove block notification stream and block channel from blocktx
- Move remaining logic from block notifier to peer handler


blocktx/peer_handler_test.go
```diff
@@ -388,7 +386,9 @@ func TestFillGaps(t *testing.T) {
 			}
 
 			logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo}))
-			peerHandler := NewPeerHandler(logger, storeMock, bockChannel, 100)
+			batchSize := 4
```
shotasilagadzetaal, line 389: why 4? Can we have it as const somewhere above?

blocktx/peer_handler.go
```diff
@@ -29,6 +29,8 @@ import (
 const (
 	transactionStoringBatchsizeDefault = 2048 // power of 2 for easier memory allocation
 	maxRequestBlocks                   = 5
+	fillGapsInterval                   = 15 * time.Minute
+	maximumBlockSize                   = 4000000000 // 4Gb
```
shotasilagadzetaal, line 33: maybe `4294967296` would be better?

blocktx/peer_handler_test.go
boecklim: The batch size of 4 was there before and as it is a unit test I didn't create a separate const for it.

It is 4 in this case so that the test will also cover this if clause: https://github.com/bitcoin-sv/arc/blob/441a5793da093f357732cfc8549e65a33ef31394/blocktx/peer_handler.go#L507

shotasilagadzetaal (approved):

PR #268, opened 2024-01-25
check not found err

shotasilagadzetaal:


api/transaction_handler/metamorph.go
```diff
@@ -70,7 +71,7 @@ func (m *Metamorph) GetTransactionStatus(ctx context.Context, txID string) (stat
 	tx, err = m.Client.GetTransactionStatus(ctx, &metamorph_api.TransactionStatusRequest{
 		Txid: txID,
 	})
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "key could not be found") {
```
boecklim, line 74: I think that will do the job.
However the string of the error could change. In order that this keeps working if the string changes you could  have an error var in `metamorph/server.go` which you can compare with e.g. like this
```
var ErrNotFound = errors.New("key could not be found")

func (s *Server) GetTransactionStatus(ctx context.Context, req *metamorph_api.TransactionStatusRequest) (*metamorph_api.TransactionStatus, error) {
	data, announcedAt, minedAt, storedAt, err := s.getTransactionData(ctx, req)
	if err != nil {
		
		if errors.Is(err, store.ErrNotFound){
			return nil, ErrNotFound
		}
   ...
```

then you could do here
```
if err != nil && !errors.Is(err, metamorph.ErrNotFound) { ... }
```

shotasilagadzetaal: The error is in api package that I'm comparing against. In that case I had to import api from metamorph (wrong direction). That's why I didn't do that. Should I declare that specific error in metamorph you say?

boecklim: When I serach for "key could not be found" - error, I find it not in `api`, but here: https://github.com/bitcoin-sv/arc/blob/1e03dcc801e275f23275152e25511523a31ff520/metamorph/store/Interface.go#L205

Yes, I'd declare another error in`metamorph/server.go`. And in `api/transaction_handler/metamorph.go` you compare against that error. Then the direction of dependencies is correct.

You could also reference `store.ErrNotFound`, but then you'll have a reference to the `metamorph`-store in `api` which is not nice.

shotasilagadzetaal: I need that error to be defined in one place. Otherwise how would I compare it? In that case I should keep those two errors in sync.

api/transaction_handler/metamorph.go
```diff
@@ -70,7 +72,7 @@ func (m *Metamorph) GetTransactionStatus(ctx context.Context, txID string) (stat
 	tx, err = m.Client.GetTransactionStatus(ctx, &metamorph_api.TransactionStatusRequest{
 		Txid: txID,
 	})
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), metamorph.ErrNotFound.Error()) {
```
boecklim, line 75: IMO it is better to compare using the `errors.Is()` function.

```
if err != nil && !errors.Is(err, metamorph.ErrNotFound.Error) { ... }
```

shotasilagadzetaal: The error isn't `key could not be found` the error is `code = Unknown desc = key could not be found` so I still need `contains` or I need to define that error as `code = Unknown desc = key could not be found`

metamorph/server.go
```diff
@@ -37,11 +36,18 @@ func init() {
...
+func PtrTo[T any](v T) *T {
+	return &v
+}
+
 const (
 	responseTimeout = 5 * time.Second
 	blocktxTimeout  = 1 * time.Second
 )
 
+var ErrNotFound = errors.New("key could not be found")
```
boecklim, line 49: I think you'll have to add this, otherwise this error has no effect.

```
func (s *Server) GetTransactionStatus(ctx context.Context, req *metamorph_api.TransactionStatusRequest) (*metamorph_api.TransactionStatus, error) {
	data, announcedAt, minedAt, storedAt, err := s.getTransactionData(ctx, req)
	if err != nil {
		
		if errors.Is(err, store.ErrNotFound){
			return nil, ErrNotFound
		}
           ...
```

boecklim:
This PR is superseded by https://github.com/bitcoin-sv/arc/pull/278

PR #272, opened 2024-01-26
Fix/unknown status

boecklim:
-  As soon as ARC processes the tx, it has status RECEIVED and not UNKNOWN
- Skip separate RECEIVED step in processor
- Default status of a stored tx is STORED
-  If waiting for <= received status, then return immediately


sirdeggen (approved):
What was the original purpose of delineating between UNKNOWN and RECEIVED?

So long as that is understood, approved. I agree UNKNOWN is not a helpful response in terms of what it implies.

PR #281, opened 2024-02-08
Refactor/transaction handler

boecklim:
- Move transaction handler to metamorph service. Reverse any dependency to api in metamorph
- Replace error package with std library
- Interface for maintainer functionality used in background-worker and k8s-watcher


metamorph/client.go
```diff
@@ -1,20 +1,50 @@
...
-	"github.com/bitcoin-sv/arc/metamorph"
 	"github.com/bitcoin-sv/arc/metamorph/metamorph_api"
 	"github.com/bitcoin-sv/arc/tracing"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+var (
+	// ErrTransactionNotFound is returned when a transaction is not found.
```
kameshraj23, line 18: Do we still need this comment as the error is self explanatory?

metamorph/client.go
```diff
@@ -1,20 +1,50 @@
...
+	ErrTransactionNotFound = errors.New("transaction not found")
+
+	// ErrParentTransactionNotFound is returned when a parent transaction is not found.
```
kameshraj23, line 21: Do we still need this comment as the error is self explanatory?

kameshraj23 (approved):

PR #282, opened 2024-02-08
Feature/nats mq

boecklim:
- Publish/subscribe for register transaction messages using NATS mq
- Peer handler in blocktx bulk inserts batches of transactions


blocktx/async/nats_mq/consumer.go
```diff
@@ -0,0 +1,74 @@
...
+	txChannel    chan []byte
+	subscription *nats.Subscription
+}
+
+func NewNatsMQConsumer(txChannel chan []byte, logger *slog.Logger, natsURL string) (*Consumer, error) {
+
+	var nc *nats.Conn
+
+	var err error
+	for i := 0; i < 5; i++ {
```
kameshraj23, line 28: Are we trying five times for it to connect?

boecklim: Yes, this tries 5 times and once that has worked it exits. But I think it should return an error if it didn't work after 5 times. Need to add that.

kameshraj23 (approved):

PR #295, opened 2024-02-16
Fix/blocktx parallel proc

boecklim:
- Blocktx processes blocks exclusively -> only one blocktx can request and process a certain block
- Blocktx processes blocks in parallel -> each blocktx can request a block
- Do not send block request automatically after 10min as it can take longer to process a large block
- Gaps also include blocks where the processing has not finished (processed_at = null)


blocktx/store/interface.go
```diff
@@ -28,5 +27,10 @@ type Interface interface {
 	MarkBlockAsDone(ctx context.Context, hash *chainhash.Hash, size uint64, txCount uint64) error
 	GetBlockGaps(ctx context.Context, heightRange int) ([]*BlockGap, error)
 	ClearBlocktxTable(ctx context.Context, retentionDays int32, table string) (*blocktx_api.ClearDataResponse, error)
+
+	SetBlockProcessing(ctx context.Context, hash *chainhash.Hash, processedBy string) (string, error)
+	GetBlockHashesProcessingInProgress(ctx context.Context, processedBy string) ([]*chainhash.Hash, error)
```
shotasilagadzetaal, line 32: How can we be processing many blocks at the same time? Isn't that dangerous? Shouldn't we be processing a new block if we have prev_block_hash already processed?

blocktx/store/sql/get_block_gaps.go
```diff
@@ -17,7 +17,9 @@ func (s *SQL) GetBlockGaps(ctx context.Context, blockHeightRange int) ([]*store.
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	q := `SELECT missing_blocks.missing_block_height, prevhash AS hash FROM blocks
+	q := `
```
shotasilagadzetaal, line 20: I still don't understand why cannot we tie blocks to the previous one? Let's not process block we don't already have processed previous block so it guarantees that we don't have gaps? Leaving gaps is very dangerous and uncomfortable for users. Imagine not giving information that something is already spent.

As a user I would prefer the most up to date state to the state that might be missing deposits or withdrawals.

blocktx/store/interface.go
boecklim: If a block announcement comes in [here](https://github.com/bitcoin-sv/arc/blob/ec2b718b940d268fdfc68464bbf05a23faf042f0/blocktx/peer_handler.go#L426) then only one `blocktx` instance can process it, because only the first instance which tries to set the block processing [here](https://github.com/bitcoin-sv/arc/blob/ec2b718b940d268fdfc68464bbf05a23faf042f0/blocktx/peer_handler.go#L218) succeeds. Every other `blocktx` instance will fail here with error `store.ErrBlockProcessingDuplicateKey`. That ensures that any `blocktx` can process a block, but only one `blocktx` will process a certain block.

There is no danger in processing a block if the previous one is not processed yet. There is the `FillGaps` [go routine](https://github.com/bitcoin-sv/arc/blob/ec2b718b940d268fdfc68464bbf05a23faf042f0/blocktx/peer_handler.go#L203) which ensures that all blocks are eventually getting processed

blocktx/peer_handler.go
```diff
@@ -534,14 +514,6 @@ const (
 )
 
 func (ph *PeerHandler) FillGaps(peer p2p.PeerI) error {
-	primary, err := ph.CheckPrimary()
```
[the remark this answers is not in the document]
boecklim: Yes. All `blocktx` instances are allowed to process blocks in parallel and exclusively i.e. each block can only be processed by one `blocktx`.

blocktx/store/sql/get_block_gaps.go
boecklim: The gap filling mechanism in place has proved working well. There are no gaps anymore. The mechanism in place which follows requests all previous blocks was not good enough: https://github.com/bitcoin-sv/arc/blob/56b20194990d79c5cf38cd8d42b94953b0cd4cca/blocktx/peer_handler.go#L587
because if for example `blocktx` is down for a while and missed some blocks it can well happen that there are gaps.

The changed query now includes not only the gaps, but also those blocks without a `processed_at` date. Blocks without this date mean, that the block processing was started, but never finished. These blocks should also be considered as gaps. This can also always happen, when for example `blocktx` gets a new deployment while it is processing a block and can't finish the processing.

shotasilagadzetaal (approved):

PR #297, opened 2024-02-19
add health check startup

shotasilagadzetaal:


config.yaml
```diff
@@ -59,7 +59,8 @@ metamorph:
   statsKeypress: false # enable stats keypress. If enabled pressing any key will print stats to stdout
   profilerAddr: localhost:9992 # address to start profiler server on
   blocktxTimeout: 1s # timeout for blocktx service
-  healthServerDialAddr: localhost:8005
+  metamorphHealthServerDialAddr: localhost:8005
+  blocktxHealthServerDialAddr: localhost:8006
```
boecklim, line 63: I think this setting `blocktxHealthServerDialAddr` belongs to `blocktx` so it should be moved e.g. line 87 without the prefix `blocktx`

cmd/blocktx.go
```diff
@@ -144,3 +155,29 @@ func StartBlockTx(logger *slog.Logger) (func(), error) {
...
+
+func StartHealthServerBlocktx(serv *blocktx.Server) error {
+	gs := grpc.NewServer()
+	defer gs.Stop()
+
+	grpc_health_v1.RegisterHealthServer(gs, serv) // registration
+	// register your own services
+	reflection.Register(gs)
+
+	address, err := config.GetString("blocktx.blocktxHealthServerDialAddr") //"localhost:8005"
```
boecklim, line 167: According to the other comment https://github.com/bitcoin-sv/arc/pull/297/files#r1494276927
this could then be changed to `blocktx.healthServerDialAddr`

boecklim (commented):
Can you please add some unit tests?
Basically you can copy them from here: metamorph/health_check_test.go

boecklim (approved):

PR #303, opened 2024-02-21
Allow blocktx to process at most 1 block at the time

boecklim:
- Allow blocktx to process at most 1 block at the time
- Quit startPeerWorker() gracefully
- e2e test with multiple `blocktx` containers
- e2e tests in github actions shows only output from tests


blocktx/peer_handler.go
```diff
@@ -232,15 +212,23 @@ func (ph *PeerHandler) startPeerWorker() {
 
 				ctx := context.Background()
 
+				bhs, err := ph.store.GetBlockHashesProcessingInProgress(ctx, ph.hostname)
+				if err == nil {
```
shotasilagadzetaal, line 216: Shouldn't we also check fo err not nil?

blocktx/peer_handler.go
```diff
@@ -232,15 +212,23 @@ func (ph *PeerHandler) startPeerWorker() {
...
+					if len(bhs) >= maxBlocksInProgress {
```
shotasilagadzetaal, line 217: Btw, if blocktx instances cannot process blocks in parallel why do we have them at all? As a backup instances?

cmd/blocktx.go
```diff
@@ -94,17 +81,50 @@ func StartBlockTx(logger *slog.Logger) (func(), error) {
...
+		peer, err := p2p.NewPeer(logger, peerURL, peerHandler, network, p2p.WithMaximumMessageSize(maximumBlockSize))
+		if err != nil {
+			return nil, fmt.Errorf("error creating peer %s: %v", peerURL, err)
+		}
+		err = pm.AddPeer(peer)
+		if err != nil {
+			return nil, fmt.Errorf("error adding peer: %v", err)
+		}
+
+		peers[i] = peer
```
shotasilagadzetaal, line 122: Would it be beneficial in the future to have the `peers` array saved in peerHandler? For instance I needed it to check that each peer was healthy and connected.

blocktx/peer_handler.go
boecklim: No, `blocktx` can process blocks in parallel since this PR: https://github.com/bitcoin-sv/arc/pull/295

blocktx/peer_handler.go
```diff
@@ -589,12 +592,27 @@ func (ph *PeerHandler) markTransactionsAsMined(blockId uint64, merkleTree []*cha
...
+			break
+		}
+	}
+
+	step := int(math.Ceil(float64(totalSize) / 5))
+	progressIndices := map[int]int{step: 20, step * 2: 40, step * 3: 60, step * 4: 80}
+
 	for txIndex, hash := range leaves {
 		// Everything to the right of the first nil will also be nil, as this is just padding upto the next PoT.
 		if hash == nil {
```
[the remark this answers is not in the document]
boecklim: Yes, but we need `txIndex` and `hash` in the loop

cmd/blocktx.go
boecklim: The reason I took `peers` out of `peerHandler` is that `peerHandler` is passed as a parameter to each peer when it is created. 
https://github.com/bitcoin-sv/arc/blob/fb75bb464d40f8ad3084175107ffc3c29369627e/cmd/blocktx.go#L113

So I found it strange if `peerHandler` is inside each `peer` and in turn all `peers` are inside `peerHandler`.

The way it is done now is that `peers` is passed to the server where they can be used for the health check instead of going via the `peerHandler`.

https://github.com/bitcoin-sv/arc/blob/fb75bb464d40f8ad3084175107ffc3c29369627e/blocktx/health_check.go#L37

Does that make sense to you?

blocktx/peer_handler.go
shotasilagadzetaal: Yes, but they do it sequentially. So there is no advantage in processing speed right?

cmd/blocktx.go
shotasilagadzetaal: yeap

shotasilagadzetaal (approved):

PR #304, opened 2024-02-22
Health api for metamorph

shotasilagadzetaal:


api/handler/default.go
```diff
@@ -80,6 +80,29 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) error {
...
+	healthy := true
+
+	err := m.TransactionHandler.Health(ct)
+
+	if err == nil {
+		return ctx.JSON(http.StatusOK, api.Health{
+			Healthy: &healthy,
+			Reason:  nil,
+		})
+	} else {
```
boecklim, line 96: The `else` is not needed here

api/handler/default.go
```diff
@@ -80,6 +80,29 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) error {
...
+		healthy = false
+		reason := err.Error()
+		return ctx.JSON(http.StatusOK, api.Health{
+			Healthy: &healthy,
```
boecklim, line 100: Consider using `PtrTo(false)` and `PtrTo(err.Error())` here

api/handler/default.go
```diff
@@ -80,6 +80,29 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) error {
 	})
 }
 
+func (m ArcDefaultHandler) GETHealth(ctx echo.Context) error {
```
boecklim, line 83: Can we have a unit test for this function?

api/handler/default.go
shotasilagadzetaal: `PtrTo(err.Error())` doesn't work :/ I fixed `false`

boecklim (approved):

PR #305, opened 2024-02-22
Refactor/bulk update status

boecklim:
- Status updates are as bulk updates in a go routine consuming a channel
- Ensure no duplicate hashes and only latest status when bulk updating status
- Remove unused update function


boecklim:
Replaced by https://github.com/bitcoin-sv/arc/pull/306

PR #308, opened 2024-02-23
Fix/Start prometheus server

boecklim:
Start prometheus server


boecklim:
Will reopen when https://github.com/bitcoin-sv/arc/pull/309 is merged

PR #314, opened 2024-03-01
Reduce size of processor response

boecklim:
- Reduce the size of processor response, in order to save memory per transaction

### DIFF
--- a/metamorph/processor.go
+++ b/metamorph/processor.go
@@ -374,7 +374,6 @@ func (p *Processor) LoadUnmined() {
 		return
 	}
 
-	p.logger.Info("loading unmined transactions", slog.Int64("limit", limit))
 	getUnminedSince := p.now().Add(-1 * p.mapExpiryTime)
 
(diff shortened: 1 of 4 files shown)


metamorph/processor.go
```diff
@@ -352,7 +351,7 @@ func (p *Processor) processExpiredTransactions() {
 					p.pm.RequestTransaction(item.Hash)
 				} else {
 					p.logger.Debug("Re-announcing expired tx", slog.String("hash", txID.String()))
-					p.pm.AnnounceTransaction(item.Hash, item.AnnouncedPeers)
+					p.pm.AnnounceTransaction(item.Hash, nil)
```
shotasilagadzetaal, line 354: Looks like we don't need second parameter anymore? 🤔

boecklim: Yes, the 2nd parameter is not needed here. But that function is defined in the p2p package: https://github.com/libsv/go-p2p/blob/d26b92c11b36216da12748833b580b41496c4dd5/PeerManager.go#L72

PR #316, opened 2024-03-04
Feature/broadcaster cli paypack

boecklim:
Changes to broadcaster-cli
- use cobra package for commands
- command to prepare utxos (not finished)
  - payback flag to send funds from receiving address back to funding address
- command to show funding and receiving address

Out of scope for this PR:
- Command for preparation of utxo set
- Command for submitting transaction at specified rate


broadcaster/utxo_preparer.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+}
+
+func NewUTXOPreparer(logger *slog.Logger, client ArcClient, fromKeySet *keyset.KeySet, toKeyset *keyset.KeySet, utxoClient UtxoClient, opts ...func(p *UTXOPreparer)) *UTXOPreparer {
+	utxoPreparer := &UTXOPreparer{
+		logger:     logger,
+		Client:     client,
+		FromKeySet: fromKeySet,
+		ToKeySet:   toKeyset,
+		IsTestnet:  true,
```
shotasilagadzetaal, line 77: Should that be hardcoded? or even const?

shotasilagadzetaal (approved):

broadcaster/utxo_preparer.go
boecklim: That is just the default value. But I'll put it in a const

PR #318, opened 2024-03-06
remove objects from map after client connection

shotasilagadzetaal:
Removed most of stats from code.

Response map doesn't have element after client connection (after ~5 sec)

Getting unmined transactions from database now.

Simplifying some callback hell.

### DIFF
--- a/metamorph/processor.go
+++ b/metamorph/processor.go
@@ -3,7 +3,6 @@ package metamorph
 import (
 	"context"
 	"errors"
-	"fmt"
 	"log/slog"
 	"net/http"
 	"os"
@@ -140,7 +139,7 @@ func NewProcessor(s store.MetamorphStore, pm p2p.PeerManagerI, opts ...Option) (
 		opt(p)
 	}
 
-	p.ProcessorResponseMap = NewProcessorResponseMap(s, p.mapExpiryTime, WithNowResponseMap(p.now))
+	p.ProcessorResponseMap = NewProcessorResponseMap(p.mapExpiryTime, WithNowResponseMap(p.now))
 
 	p.logger.Info("Starting processor", slog.String("cacheExpiryTime", p.mapExpiryTime.String()))
 
@@ -256,7 +255,6 @@ func (p *Processor) processStatusUpdates() {
 				if !found || (found && statusValueMap[foundStatusUpdate.Status] < statusValueMap[statusUpdate.Status]) {
 					statusUpdatesMap[statusUpdate.Hash] = statusUpdate
 				}
-				fmt.Println("AAAAA")
 				p.CheckAndUpdate(&statusUpdates, &statusUpdatesMap, true)
 			case <-ticker.C:
 				p.CheckAndUpdate(&statusUpdates, &statusUpdatesMap, false)
--- a/metamorph/processor.go
+++ b/metamorph/processor.go
@@ -216,21 +216,21 @@ func (p *Processor) processMinedCallbacks() {
 	}()
 }
 
-func (p *Processor) CheckAndUpdate(statusUpdates *[]store.UpdateStatus, statusUpdatesMap *map[chainhash.Hash]store.UpdateStatus, skipCheck bool) {
-	if skipCheck && len(*statusUpdatesMap) < p.processStatusUpdatesBatchSize {
+func (p *Processor) CheckAndUpdate(statusUpdatesMap *map[chainhash.Hash]store.UpdateStatus, skipCheck bool) {
+	if (skipCheck && len(*statusUpdatesMap) < p.processStatusUpdatesBatchSize) || len(*statusUpdatesMap) == 0 {
 		return
 	}
 
+	statusUpdates := make([]store.UpdateStatus, 0, p.processStatusUpdatesBatchSize)
 	for _, distinctStatusUpdate := range *statusUpdatesMap {
-		*statusUpdates = append(*statusUpdates, distinctStatusUpdate)
+		statusUpdates = append(statusUpdates, distinctStatusUpdate)
 	}
 
-	err := p.statusUpdateWithCallback(*statusUpdates)
+	err := p.statusUpdateWithCallback(statusUpdates)
 	if err != nil {
 		p.logger.Error("failed to bulk update statuses", slog.String("err", err.Error()))
 	}
 
-	*statusUpdates = make([]store.UpdateStatus, 0, p.processStatusUpdatesBatchSize)
 	*statusUpdatesMap = map[chainhash.Hash]store.UpdateStatus{}
 }
 
@@ -242,7 +242,6 @@ func (p *Processor) processStatusUpdates() {
 			p.quitListenStatusUpdateChComplete <- struct{}{}
 		}()
 
-		statusUpdates := make([]store.UpdateStatus, 0, p.processStatusUpdatesBatchSize)
 		statusUpdatesMap := map[chainhash.Hash]store.UpdateStatus{}
 
 		for {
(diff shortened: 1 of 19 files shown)


metamorph/processor.go
```diff
@@ -310,35 +278,51 @@ func (p *Processor) statusUpdateWithCallback(statusUpdates []store.UpdateStatus)
...
 
-		if len(expiredTransactionItems) > 0 {
-			p.logger.Info("Resending expired transactions", slog.Int("number", len(expiredTransactionItems)))
-			for txID, item := range expiredTransactionItems {
-				startTime := time.Now()
-				retries := item.GetRetries()
-				item.IncrementRetry()
+		for {
+			// get all transactions since then chunk by chunk
+			fmt.Println(loadUnminedLimit, offset)
```
boecklim, line 293: I'd suggest either remove this line or use the logger for this

metamorph/processor.go
```diff
@@ -358,42 +342,42 @@ func (p *Processor) GetPeers() ([]string, []string) {
 	return peersConnected, peersDisconnected
 }
 
-func (p *Processor) LoadUnmined() {
-	span, spanCtx := opentracing.StartSpanFromContext(context.Background(), "Processor:LoadUnmined")
-	defer span.Finish()
+// func (p *Processor) LoadUnmined() {
```
boecklim, line 345: Remove lines instead of commenting them

metamorph/store/postgresql/postgres.go
```diff
@@ -360,14 +371,15 @@ func (p *PostgreSQL) GetUnmined(ctx context.Context, since time.Time, limit int6
 		,full_status_updates
 		,raw_tx
 		,locked_by
+		,retries
 		FROM metamorph.transactions
 		WHERE locked_by = 'NONE'
```
boecklim, line 376: I think we have to include here also those which are locked by this `metamorph`-pod, not?
```
WHERE (locked_by = 'NONE' OR locked_by = $6)
```
Where `$6` would be `p.hostname` on line 382 

That will ensure that only 1 metamorph pod is loading and retrying these unmined txs as they are set locked on line 485.

metamorph/processor.go
```diff
@@ -494,12 +436,13 @@ func (p *Processor) ProcessTransaction(ctx context.Context, req *ProcessorReques
...
 	data, err := p.store.Get(ctx, req.Data.Hash[:])
-
 	if err == nil {
+		/*
+			When transaction is re-submitted we make inserted_at_num to be now()
+			to make sure it will be loaded and re-broadcasted if needed.
+		*/
 		insertedAtNum, _ := strconv.Atoi(p.now().Format("2006010215"))
 		data.InsertedAtNum = insertedAtNum
 		if set_err := p.store.Set(spanCtx, req.Data.Hash[:], data); set_err != nil {
```
boecklim, line 448: Not related to the PR, but can you please rename `set_err` to `setErr`?

metamorph/processor.go
```diff
@@ -412,73 +396,31 @@ var statusValueMap = map[metamorph_api.Status]int{
...
-			}
-		},
+	p.storageStatusUpdateCh <- store.UpdateStatus{
+		Hash:         *hash,
+		Status:       status,
+		RejectReason: rejectReason,
 	}
 
-	processorResponse.UpdateStatus(statusUpdate)
+	// if we recieve new update check if we have client connection waiting for status and send it
```
boecklim, line 414: *receive

metamorph/processor_response/processor_response.go
```diff
@@ -19,126 +17,68 @@ type StatusAndError struct {
...
 
-	statKey := fmt.Sprintf("%d: %s", statusUpdate.Status, statusUpdate.Status.String())
-	r.LastStatusUpdateNanos.Store(gocore.NewStat("processorResponse").NewStat(statKey).AddTime(r.LastStatusUpdateNanos.Load()))
-
-	if !statusUpdate.IgnoreCallback {
-		statusUpdate.Callback(nil)
-	}
 }
 
 func (r *ProcessorResponse) String() string {
```
boecklim, line 60: I think we can remove this function as it is only used here: `metamorph/processor_response/processor_response_test.go`

```

func TestString(t *testing.T) {
	t.Run("String", func(t *testing.T) {
		response := NewProcessorResponse(testdata.TX1Hash)
		assert.IsType(t, "string", response.String())
	})
}
```

metamorph/processor.go
```diff
@@ -310,35 +278,51 @@ func (p *Processor) statusUpdateWithCallback(statusUpdates []store.UpdateStatus)
...
+			fmt.Println(loadUnminedLimit, offset)
+			unminedTxs, err := p.store.GetUnmined(dbctx, getUnminedSince, loadUnminedLimit, offset)
+			if err != nil {
+				p.logger.Error("Failed to get unmined transactions", slog.String("err", err.Error()))
+				continue
+			}
+
+			offset += loadUnminedLimit
+			p.logger.Info("loaded unmined transactions", slog.Int("number", len(unminedTxs)), slog.Int64("limit", loadUnminedLimit))
+			if len(unminedTxs) == 0 {
```
boecklim, line 302: The whole for-loop of line 291 will only stop once there are no more unmined txs to be loaded and retried, right?

metamorph/processor.go
```diff
@@ -225,62 +216,48 @@ func (p *Processor) processMinedCallbacks() {
 	}()
 }
 
-func (p *Processor) processStatusUpdates() {
+func (p *Processor) CheckAndUpdate(statusUpdatesMap *map[chainhash.Hash]store.UpdateStatus, skipCheck bool) {
+	if (skipCheck && len(*statusUpdatesMap) < p.processStatusUpdatesBatchSize) || len(*statusUpdatesMap) == 0 {
```
boecklim, line 220: I find it a good idea to extract the duplicate code into this function. But I think this part I would keep out of it. Because that is a bit hard to read.
```
if (skipCheck && len(*statusUpdatesMap) < p.processStatusUpdatesBatchSize) || len(*statusUpdatesMap) == 0 {}
```

Then you don't need the `skipCheck` parameter. 

With this I mean I would keep this code before calling `CheckAndUpdate`
```
if len(statusUpdatesMap) < p.processStatusUpdatesBatchSize {
					continue
				}
```
... and this one respectively
```
				if len(statusUpdatesMap) == 0 {
					continue
				}
```

metamorph/store/postgresql/postgres.go
shotasilagadzetaal: Yes, that was my question. So that will guarantee to retry stalled transactions.

metamorph/processor.go
shotasilagadzetaal: Yeap, exactly

metamorph/store/postgresql/postgres.go
boecklim: In that case you can then on line 485 first check if the locked by field is equal to "NONE" in order to save unnecessary updates on the DB.

```
		if data.LockedBy == "NONE" {
			err = p.setLockedBy(ctx, data.Hash, p.hostname)
			if err != nil {
				return nil, err
			}
		}

```

metamorph/processor.go
```diff
@@ -225,62 +215,50 @@ func (p *Processor) processMinedCallbacks() {
 	}()
 }
 
-func (p *Processor) processStatusUpdates() {
+func (p *Processor) CheckAndUpdate(statusUpdatesMap *map[chainhash.Hash]store.UpdateStatus) {
```
boecklim, line 218: I think the map does not need to be passed as a pointer becauce [Map types are reference types, like pointers or slices](https://go.dev/blog/maps)

metamorph/store/postgresql/postgres.go
```diff
@@ -464,6 +478,10 @@ func (p *PostgreSQL) GetUnmined(ctx context.Context, since time.Time, limit int6
 			data.LockedBy = lockedBy.String
 		}
 
+		if retries.Valid {
+			data.Retries = int(retries.Int32)
+		}
+
 		err = p.setLockedBy(ctx, data.Hash, p.hostname)
```
boecklim, line 485: We should set the locked by field only to the hostname if it is not set yet. That may save us some unnecessary updates on the DB
```
		if data.LockedBy == "NONE" {
			err = p.setLockedBy(ctx, data.Hash, p.hostname)
			if err != nil {
				return nil, err
			}
		}
```

metamorph/processor.go
shotasilagadzetaal: Screenshot 2024-03-08 at 11 01 12

boecklim (approved):

PR #328, opened 2024-03-12
Project layout

shotasilagadzetaal:


assets/LICENSE
boecklim, line 1: I think this file has to be in the root folder so that github will find it

internal/fees/fess.go
boecklim, line 1: Please rename this file to `fees.go`

internal/fees/fess_test.go
boecklim, line 1: Please rename this file to `fees_test.go`

config/settings.conf
boecklim, line 1: We don't use `settings.conf` anymore. Please remove.

Makefile
boecklim, line 1: There are some broken paths in the Makefile. Please see this commit: https://github.com/bitcoin-sv/arc/commit/2e4145e4e27627de8d45919b7ac07382c61841bd

config/.pre-commit-config.yaml
boecklim, line 1: I think that file needs to be in the root folder

Makefile
boecklim: Also this path: https://github.com/bitcoin-sv/arc/commit/8a7a914fdf9610e6ceb0271e48861b9182c0821d

PR #333, opened 2024-03-14
cause deadlock in pq

shotasilagadzetaal:


internal/metamorph/store/postgresql/postgres.go
```diff
@@ -464,12 +464,38 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
 		;
     `
 
-	rows, err := p.db.QueryContext(ctx, qBulk, pq.Array(txHashes), pq.Array(statuses), pq.Array(rejectReasons), p.hostname, metamorph_api.Status_SEEN_ON_NETWORK, metamorph_api.Status_MINED)
+	tx, err := p.db.Begin()
 	if err != nil {
+		panic(err)
```
boecklim, line 469: panic stops the whole program from execution. I think we should simply return an error here.

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -464,12 +464,38 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+	}
+
+	_, err = tx.Exec("LOCK TABLE metamorph.transactions IN EXCLUSIVE MODE")
+	if err != nil {
+		if err := tx.Rollback(); err != nil {
+			panic(err)
```
boecklim, line 475: panic stops the whole program from execution. I think we should simply return an error here.

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -464,12 +464,38 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+		}
 		return nil, err
 	}
 
-	return p.getStoreDataFromRows(rows)
+	rows, err := tx.QueryContext(ctx, qBulk, pq.Array(txHashes), pq.Array(statuses), pq.Array(rejectReasons), metamorph_api.Status_SEEN_ON_NETWORK, metamorph_api.Status_MINED)
+	if err != nil {
+		if err := tx.Rollback(); err != nil {
+			panic(err)
```
boecklim, line 483: panic stops the whole program from execution. I think we should simply return an error here.

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -525,19 +552,44 @@ func (p *PostgreSQL) UpdateMined(ctx context.Context, txsBlocks *blocktx_api.Tra
...
 `
-	rows, err := p.db.QueryContext(ctx, qBulkUpdate, metamorph_api.Status_MINED, p.now(), pq.Array(txHashes), pq.Array(blockHashes), pq.Array(blockHeights), pq.Array(merklePaths))
+	tx, err := p.db.Begin()
 	if err != nil {
-		return nil, fmt.Errorf("failed to execute transaction update query: %v", err)
+		panic(err)
 	}
 
-	return p.getStoreDataFromRows(rows)
+	_, err = tx.Exec("LOCK TABLE metamorph.transactions IN EXCLUSIVE MODE")
```
boecklim, line 560: We haven't seen a deadlock issue in `UpdateMined`. I would suggest to only do the change in `UpdateStatusBulk` for now and see what happens.

internal/metamorph/store/postgresql/postgres.go
shotasilagadzetaal: I left it because the error is so unexpected here and also if there is error it means the application cannot even run anymore.Db is locked so IDK, doesn't change much removing this

internal/metamorph/store/postgresql/postgres.go
shotasilagadzetaal: I think it's just because it was less probably. It does the same thing - updates bunch of rows in the database. So it would be better to protect ourselves.

internal/metamorph/store/postgresql/postgres.go
boecklim: I think it is better to return the error as usual, instead of shutting down the whole service in an uncontrolled way using `panic`. IMO there should be no panics anywhere, except maybe in examples.

If the goal is to shutdown the service completely, then waiting for the particular error and shutting down the service gracefully would be the better option.

internal/metamorph/store/postgresql/postgres.go
boecklim: OK. Let's see what is the performance penalty of locking the table for each call.

boecklim (approved):

PR #351, opened 2024-04-03
lock rows in order to avoid deadlock

shotasilagadzetaal:
Instead of locking the whole table, lock modified rows in order to avoid deadlock.

### DIFF
--- a/internal/metamorph/store/postgresql/postgres.go
+++ b/internal/metamorph/store/postgresql/postgres.go
@@ -440,7 +440,7 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
 		return nil, err
 	}
 
-	_, err = tx.Exec("LOCK TABLE metamorph.transactions IN EXCLUSIVE MODE")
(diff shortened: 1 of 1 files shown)


internal/metamorph/store/postgresql/postgres.go
```diff
@@ -440,7 +440,7 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
 		return nil, err
 	}
 
-	_, err = tx.Exec("LOCK TABLE metamorph.transactions IN EXCLUSIVE MODE")
+	_, err = tx.Exec(`SELECT * FROM metamorph.transactions WHERE hash in (SELECT UNNEST($1::BYTEA[])) ORDER BY hash FOR UPDATE`, pq.Array(txHashes))
```
boecklim, line 443: What does this query do?
It looks to me like there is just a query which has no effect as the result is not used anywhere

shotasilagadzetaal: So when we have it in transaction it picks those specific rows and locks them in order. Having it in transaction is what makes difference.

boecklim (approved):

PR #371, opened 2024-04-15
change bulk update size, run updates in separate routines to speed up process

shotasilagadzetaal:
As bulk updates don't have intersection in transactions they can be running separately (using db connection pool) to speed up block processing. Also increasing bulk size.

### DIFF
--- a/internal/blocktx/store/postgresql/postgres.go
+++ b/internal/blocktx/store/postgresql/postgres.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	postgresDriverName        = "postgres"
-	maxPostgresBulkInsertRows = 1000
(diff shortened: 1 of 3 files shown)


internal/blocktx/store/postgresql/postgres.go
```diff
@@ -14,7 +14,7 @@ import (
 
 const (
 	postgresDriverName        = "postgres"
-	maxPostgresBulkInsertRows = 1000
+	maxPostgresBulkInsertRows = 8192
```
boecklim, line 17: Where does this number come from?

boecklim:
Change of batch size has increased processing speed enough for the time being

PR #385, opened 2024-04-22
fix query for seen on network transactions

shotasilagadzetaal:


internal/metamorph/store/postgresql/postgres.go
```diff
@@ -377,18 +389,36 @@ func (p *PostgreSQL) GetSeenOnNetwork(ctx context.Context, since time.Time, limi
...
-		LIMIT $3 OFFSET $4;`
+		,retries;`
+
+	tx, err := p.db.Begin()
+	if err != nil {
+		return nil, err
+	}
 
-	rows, err := p.db.QueryContext(ctx, q, metamorph_api.Status_SEEN_ON_NETWORK, since.Format(numericalDateHourLayout), limit, offset, p.hostname)
+	_, err = tx.Exec(`SELECT hash
```
boecklim, line 399: Is this query intended to lock the rows?

I think you could do all in one query like this

```sql
`UPDATE metamorph.transactions
		SET locked_by=$5
		FROM (
			SELECT hash
			FROM metamorph.transactions
			WHERE (locked_by = $5 OR locked_by = 'NONE')
			AND status = $1
			AND inserted_at_num > $2
			ORDER BY inserted_at_num DESC
			LIMIT $3 OFFSET $4
                        FOR UPDATE SKIP LOCKED
                 ) as t
		WHERE metamorph.transactions.hash = t.hash
		RETURNING
	     stored_at
		,announced_at
		,mined_at
		,t.hash
		,status
		,block_height
		,block_hash
	@@ -377,18 +389,36 @@ func (p *PostgreSQL) GetSeenOnNetwork(ctx context.Context, since time.Time, limi
		,raw_tx
		,locked_by
     	,merkle_path
		,retries;`
```

As it is done here: https://github.com/bitcoin-sv/arc/blob/eb27e0619cb2955589bdfead40ec838ab3c4c89d/internal/metamorph/store/postgresql/postgres.go#L317

internal/metamorph/store/postgresql/postgres_test.go
```diff
@@ -539,27 +539,27 @@ func TestPostgresDB(t *testing.T) {
...
+			InsertedAtNum: 2024010101,
 		}
 
 		err = postgresDB.Set(ctx, testdata.TX6Hash[:], tx6Data)
 		require.NoError(t, err)
 
 		records, err := postgresDB.GetSeenOnNetwork(ctx, time.Date(2023, 1, 1, 1, 0, 0, 0, time.UTC), 2, 0)
 		require.NoError(t, err)
-		require.Equal(t, len(records), 1)
+		require.Equal(t, 1, len(records))
```
boecklim, line 563: Here it is just checked the number of records returned.

Can you check in this test that field `locked_by` was updated by the query?

internal/metamorph/store/postgresql/postgres.go
shotasilagadzetaal: I'm not sure. That's how we did earlier and it caused deadlock right? 

What if we are updating the same rows? I guess that would be less likely 🤔

internal/metamorph/store/postgresql/postgres_test.go
shotasilagadzetaal: Done

internal/metamorph/store/postgresql/postgres.go
boecklim: Ok, let's try it like this

boecklim (approved):

PR #392, opened 2024-04-24
don't process latest transactions

shotasilagadzetaal:


internal/metamorph/store/postgresql/postgres.go
```diff
@@ -17,6 +17,7 @@ import (
 const (
 	postgresDriverName      = "postgres"
 	numericalDateHourLayout = "2006010215"
+	until                   = -1 * 2 * time.Hour
```
boecklim, line 20: nit: simplification `-2*time.Hour`

boecklim (approved):

PR #395, opened 2024-04-26
return err if callback is localhost

shotasilagadzetaal:


internal/metamorph/server.go
```diff
@@ -165,6 +165,13 @@ func ValidateCallbackURL(callbackURL string) error {
 		return nil
 	}
 
+	rejectedCallbackUrlSubstrings := viper.GetStringSlice("metamorph.rejectCallbackContaining")
```
boecklim, line 168: I think this validation should happen in `api` service.

We already have here one validation: https://github.com/bitcoin-sv/arc/blob/277ef146f0210d093daeb33438050bdf082657d8/pkg/api/handler/default.go#L338

pkg/api/handler/default.go
```diff
@@ -332,12 +334,26 @@ func getTransactionOptions(params api.POSTTransactionParams) (*metamorph.Transac
 	return getTransactionsOptions(api.POSTTransactionsParams(params))
 }
 
+func ValidateCallbackURL(callbackURL string) error {
+	_, err := url.ParseRequestURI(callbackURL)
+	if err != nil {
+		return fmt.Errorf("invalid callback URL [%w]", err)
+	}
+
+	rejectedCallbackUrlSubstrings := viper.GetStringSlice("metamorph.rejectCallbackContaining")
```
boecklim, line 343: We shouldn't use viper here. Like in the other services this setting should be passed as an option https://github.com/bitcoin-sv/arc/blob/277ef146f0210d093daeb33438050bdf082657d8/pkg/api/handler/default.go#L42

when a new default handler is created https://github.com/bitcoin-sv/arc/blob/277ef146f0210d093daeb33438050bdf082657d8/cmd/arc/services/api.go#L109

This should be called `viper.GetStringSlice("metamorph.rejectCallbackContaining")` in `api.go`

pkg/api/handler/default_test.go
```diff
@@ -136,6 +136,43 @@ func TestGETHealth(t *testing.T) {
...
+		{
+			name:        "valid callback URL",
+			callbackURL: "http://api.callback.com",
+		},
+		{
+			name:             "blocked url",
+			callbackURL:      "http://localhost",
+			expectedErrorStr: "callback url not acceptable",
+		},
+	}
```
boecklim, line 160: Add a test case which covers this code

```
	for _, substring := range rejectedCallbackUrlSubstrings {
		if strings.Contains(callbackURL, substring) {
			return fmt.Errorf("callback url not acceptable %s", callbackURL)
		}
	}
```

shotasilagadzetaal: I think it's covered with test case `expectedErrorStr: "callback url not acceptable",`

pkg/api/handler/default.go
shotasilagadzetaal: For now no option is passed here `handler.NewDefault(logger, metamorphClient, policy)` How would it work? Would I pass function option which, after running, would set local slice of unacceptable callback url strings(read from viper)?

boecklim: Yes. Options are already defined here: https://github.com/bitcoin-sv/arc/blob/277ef146f0210d093daeb33438050bdf082657d8/pkg/api/handler/default.go#L42

Slice of unacceptable callback url strings should then be a member of the `*ArcDefaultHandler` - type and passed in as option read from Viper

PR #400, opened 2024-04-30
fix response data

shotasilagadzetaal:


pkg/api/handler/default.go
```diff
@@ -187,13 +186,16 @@ func (m ArcDefaultHandler) GETTransactionStatus(ctx echo.Context, id string) err
...
+	emptyString := ""
+
 	return ctx.JSON(http.StatusOK, api.TransactionStatus{
 		BlockHash:   &tx.BlockHash,
 		BlockHeight: &tx.BlockHeight,
 		TxStatus:    &tx.Status,
 		Timestamp:   m.now(),
 		Txid:        tx.TxID,
 		MerklePath:  &tx.MerklePath,
+		ExtraInfo:   &emptyString,
```
boecklim, line 198: You could use this instead of creating that empty string variable `PtrTo(""),`

pkg/api/arc.yml
```diff
@@ -551,21 +538,8 @@ components:
     TransactionSubmitStatus:
```
boecklim, line 538: I think you can now remove the whole `TransactionSubmitStatus` schema as it does not contain anything anymore

pkg/api/status.go
```diff
@@ -25,8 +25,10 @@ const (
 )
 
 func NewErrorFields(status StatusCode, extraInfo string) *ErrorFields {
+	emptyString := ""
 	errFields := ErrorFields{
-		Status: int(status),
+		Status:    int(status),
+		ExtraInfo: &emptyString,
```
boecklim, line 31: You could use this instead of creating that empty string variable PtrTo(""),

shotasilagadzetaal: yeah, I tried that but `PtrTo` is undefined there and didn't want to copy function.

pkg/api/handler/default.go
shotasilagadzetaal: same here, `PtrTo` is undefined.

boecklim: Have you checked? 
It is defined

```
	return ctx.JSON(http.StatusOK, api.TransactionStatus{
		BlockHash:   &tx.BlockHash,
		BlockHeight: &tx.BlockHeight,
		TxStatus:    &tx.Status,
		Timestamp:   m.now(),
		Txid:        tx.TxID,
		MerklePath:  &tx.MerklePath,
		ExtraInfo:   PtrTo(""),
	})
```

shotasilagadzetaal: Oh, it should work here

PR #413, opened 2024-05-07
distribute callbacks over time, increase timeout

shotasilagadzetaal:


internal/metamorph/callbacker.go
```diff
@@ -95,7 +96,9 @@ func (p *Callbacker) SendCallback(logger *slog.Logger, tx *store.StoreData) {
 		logger.Debug("Callback response status code not ok", slog.String("url", tx.CallbackUrl), slog.String("token", tx.CallbackToken), slog.String("hash", tx.Hash.String()), slog.Int("status", response.StatusCode))
 
 		// sleep before trying again
-		time.Sleep(time.Duration(sleepDuration) * time.Second)
+		base := time.Duration(sleepDuration) * time.Millisecond
+		time.Sleep(time.Duration(rand.Intn(1000)) * base)
```
boecklim, line 100: So at each iteration the values of the sleep time could be
1. iteration: Between `5 ms` and `5 s`
2. iteration: Between `7 ms` and `7 s`
3. iteration: Between `9 ms` and `9 s`
4. iteration: Between `11 ms` and `11 s`
5. iteration: Between `13 ms` and `13 s`

Correct?

shotasilagadzetaal: yes, so the interval in which the sleep duration is chosen randomly increases.

boecklim (approved):

PR #423, opened 2024-05-17
Revert logic

boecklim:
Revert logic: Announce expired transactions first, then request them


boecklim:
Closing as change is not needed

PR #424, opened 2024-05-21
feat(ARCO-21): BEEF parsing and structural validation

kuba-4chain:


pkg/api/handler/parsers.go
```diff
@@ -0,0 +1,96 @@
...
+	case "application/json":
+		var txBody api.POSTTransactionsJSONRequestBody
+		if err = json.Unmarshal(body, &txBody); err != nil {
+			return nil, api.NewErrorFields(api.ErrStatusBadRequest, err.Error())
+		}
+
+		for _, tx := range txBody {
+			partialHex, err := hex.DecodeString(tx.RawTx)
+			if err != nil {
+				return nil, api.NewErrorFields(api.ErrStatusBadRequest, err.Error())
```
arkadiuszos4chain, line 85: Move `api.NewErrorFields(...)` to handler implementation and return vanila errors from parsers

internal/beef/beef.go
```diff
@@ -0,0 +1,277 @@
...
+	if len(txHex) < versionBytesCount+markerBytesCount {
+		return false
+	}
+
+	// removes version bytes
+	txHex = txHex[versionBytesCount:]
+
+	if txHex[0] != BEEFMarkerPart1 || txHex[1] != BEEFMarkerPart2 {
+		return false
+	}
```
arkadiuszos4chain, line 50:
```suggestion
	if txHex[2] != BEEFMarkerPart1 || txHex[3] != BEEFMarkerPart2 {
		return false
	}
```

internal/beef/beef.go
```diff
@@ -0,0 +1,277 @@
...
+	txID string
+}
+
+type BEEF struct {
+	BUMPs        BUMPs
+	Transactions []*TxData
+}
+
+func CheckBeefFormat(txHex []byte) bool {
+	if len(txHex) < versionBytesCount+markerBytesCount {
```
arkadiuszos4chain, line 41:
```suggestion
	if len(txHex) < 4 {
```
It's known from the BRC

internal/beef/beef.go
```diff
@@ -0,0 +1,277 @@
...
+			Transaction: tx,
+			BumpIndex:   pathIndex,
+		})
+	}
+
+	return transactions, bytes, nil
+}
+
+func extractBytesWithoutVersionAndMarker(beefBytes []byte) ([]byte, error) {
+	if len(beefBytes) < 4 {
```
arkadiuszos4chain, line 254: I believe we can assume that we already have the hex validated and do not need to repeat the validation

internal/beef/beef.go
```diff
@@ -0,0 +1,277 @@
...
+	maxTreeHeight     = 64
+)
+
+const (
+	BEEFMarkerPart1 = 0xBE
+	BEEFMarkerPart2 = 0xEF
+)
+
+const (
+	HasNoBump = 0x00
```
arkadiuszos4chain, line 24: Should be unexported

internal/beef/beef.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	return decodedBeef.GetLatestTx(), decodedBeef, remainingBytes, nil
+}
+
+func (d *BEEF) GetLatestTx() *bt.Tx {
+	return d.Transactions[len(d.Transactions)-1].Transaction // get the last transaction as the processed transaction - it should be the last one because of khan's ordering
+}
+
+func decodeBUMPs(beefBytes []byte) ([]*BUMP, []byte, error) {
```
arkadiuszos4chain, line 83: I would try to use bump (and its decoding) implementation from `go-bc` repository

arkadiuszos4chain (changes requested):

pkg/api/handler/parsers.go
```diff
@@ -0,0 +1,96 @@
...
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/bitcoin-sv/arc/pkg/api"
+)
+
+func parseTransactionFromRequest(request *http.Request) ([]byte, *api.ErrorFields) {
+	requestBody := request.Body
+	contentType := request.Header.Get("Content-Type")
```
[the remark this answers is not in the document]
kuba-4chain: Done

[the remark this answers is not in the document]
kuba-4chain: Done

pkg/api/handler/parsers.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if len(body) == 0 {
+		return nil, api.NewErrorFields(api.ErrStatusBadRequest, "no transaction found - empty request body")
+	}
+
+	var txHex []byte
+
+	switch contentType {
+	case "text/plain":
```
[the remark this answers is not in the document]
kuba-4chain: Done for all content-types

pkg/api/handler/parsers.go
```diff
@@ -0,0 +1,96 @@
...
+		strBody := strings.ReplaceAll(string(body), "\n", "")
```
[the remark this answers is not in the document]
kuba-4chain: I switched the parsing, for a more robust approach. Please check it now :)

pkg/api/handler/parsers.go
kuba-4chain: Good point, done

internal/beef/beef.go
kuba-4chain: Done

internal/beef/beef.go
kuba-4chain: Yes, but I hate magic numbers. If there's a random number in the code, it should be a named-variable so you know what it means. 

But I refactored it and removed these two variables in favour of one.

internal/beef/beef.go
kuba-4chain: I'm not sure about that. `CheckBeefFormat()` is currently used for checking the format, because both formats come on the same endpoint. If there will be a situation when we expect BEEF, we may just use DecodeBEEF() without checking the format first, and it should still throw an error and inform us that the tx we passed is not in fact BEEF. 

So IMO, it should stay there. But I refactored it and used the already existing `CheckBeefFormat()`, so we don't repeat ourselves :D

internal/beef/beef.go
kuba-4chain: Done

pkg/api/handler/parsers.go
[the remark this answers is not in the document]
kuba-4chain: Changed in parsers_test.go as well

internal/beef/beef.go
kuba-4chain: Done

internal/beef/beef.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if nBump == 0 {
+		return nil, nil, errors.New("invalid BEEF - lack of BUMPs")
+	}
+
+	beefBytes = beefBytes[bytesUsed:]
+
+	bumps := make([]*bc.BUMP, 0, uint64(nBump))
+	for i := uint64(0); i < uint64(nBump); i++ {
+		if len(beefBytes) == 0 {
```
arkadiuszos4chain, line 92: This check can be removed

kuba-4chain: Done

arkadiuszos4chain (approved):

internal/beef/beef_test.go
```diff
@@ -0,0 +1,284 @@
...
+	"testing"
+
+	"github.com/libsv/go-bc"
+	"github.com/libsv/go-bt/v2"
+	"github.com/libsv/go-bt/v2/bscript"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckBeefFormat(t *testing.T) {
```
boecklim, line 15: nit: You could also make this test table based like the others

boecklim (approved):

internal/beef/beef_test.go
kuba-4chain: Done :)

PR #425, opened 2024-05-23
fix cancellation mechanism with contexts

shotasilagadzetaal:


cmd/arc/services/metamorph.go
```diff
@@ -148,14 +148,40 @@ func StartMetamorph(logger *slog.Logger) (func(), error) {
...
+	metamorphProcessor.StartLockTransactions(ctx)
 	time.Sleep(200 * time.Millisecond) // wait a short time so that process expired transactions will start shortly after lock transactions go routine
 
-	metamorphProcessor.StartProcessExpiredTransactions()
-	metamorphProcessor.StartRequestingSeenOnNetworkTxs()
-	metamorphProcessor.StartProcessStatusUpdatesInStorage()
-	metamorphProcessor.StartProcessMinedCallbacks()
-	err = metamorphProcessor.StartCollectStats()
+	ctx, cancel = context.WithCancel(context.Background())
+	metamorphProcessor.CancelProcessExpiredTransactions = cancel
```
arkadiuszos4chain, line 161: The cancel functions and wait group look like implementation details. Can't they be hidden from the client inside the processor?

internal/k8s_watcher/watcher.go
```diff
@@ -24,17 +24,16 @@ type K8sClient interface {
...
-	shutdownBlocktx           chan struct{}
-	shutdownBlocktxComplete   chan struct{}
+	metamorphClient   metamorph.TransactionMaintainer
+	blocktxClient     blocktx.BlocktxClient
+	k8sClient         K8sClient
+	logger            *slog.Logger
+	tickerMetamorph   Ticker
+	tickerBlocktx     Ticker
+	namespace         string
+	WaitGroup         *sync.WaitGroup
```
arkadiuszos4chain, line 34: Shouldn't be exported

internal/k8s_watcher/watcher.go
```diff
@@ -98,26 +94,28 @@ func WithBlocktxTicker(t Ticker) func(*Watcher) {
 }
 
 func (c *Watcher) Start() error {
-	c.watchMetamorph()
+	ctx, cancel := context.WithCancel(context.Background())
```
arkadiuszos4chain, line 97: Maybe creation of context, it's cancel function and incrementing waitgroup counter could be moved to the methods that use them?
Like  in `internal/metamorph/processor.go line 215` from PR #432

internal/k8s_watcher/watcher.go
```diff
@@ -98,26 +94,28 @@ func WithBlocktxTicker(t Ticker) func(*Watcher) {
...
-			}
-			c.shutdownBlocktxComplete <- struct{}{}
-		}()
-
 		var runningPods map[string]struct{}
 
 		for {
 			select {
+			case <-ctx.Done():
+				c.WaitGroup.Done()
```
arkadiuszos4chain, line 117: Please use `defer` for the call to `c.WaitGroup.Done()`

arkadiuszos4chain: And move it outside the loop

arkadiuszos4chain (changes requested):

cmd/arc/services/metamorph.go
shotasilagadzetaal: In some cases we start those functions from `cmd` directory, that's why I needed to have those exported not to change existing code. For metamorph, I did the same to have the same style.

internal/k8s_watcher/watcher.go
shotasilagadzetaal: Same as above.

internal/k8s_watcher/watcher.go
shotasilagadzetaal: Basically that wouldn't be the correct style. We pass the context to the function from outside to pass cancellation and whatever data we have. If we create those ctx and cancel function inside it won't make sense. I'm fine having those inside but just doesn't seem right.

internal/k8s_watcher/watcher.go
shotasilagadzetaal: Fixed.

cmd/arc/services/metamorph.go
boecklim: IMO it is more readable when all logic concerning the go routine life cycle is in processor and not in `metamorph.go`

e.g. like this


```
func (p *Processor) StartLockTransactions(ctx context.Context) {

	ctx, cancel := context.WithCancel(context.Background())
	p.cancelLockTransactions = cancel
	p.waitGroup.Add(1)

	ticker := time.NewTicker(p.lockTransactionsInterval)
	go func() {
		defer p.waitGroup.Done()
		for {
			select {
			case <-ctx.Done():
				return
			case <-ticker.C:
				expiredSince := p.now().Add(-1 * p.mapExpiryTime)
				err := p.store.SetLocked(ctx, expiredSince, loadUnminedLimit)
				if err != nil {
					p.logger.Error("Failed to set transactions locked", slog.String("err", err.Error()))
				}
			}
		}
	}()
}

```

arkadiuszos4chain (approved):

internal/metamorph/server.go
```diff
@@ -136,7 +136,6 @@ func (s *Server) StartGRPCServer(address string, grpcMessageSize int, prometheus
 
 func (s *Server) Shutdown() {
 	s.logger.Info("Shutting down")
-	s.processor.Shutdown()
```
boecklim, line 139: Don't we need this?

shotasilagadzetaal: I think not, I found out we are calling `processor.Shutdown()` twice and caused race conditions in tests.

arkadiuszos4chain: I believe we don't call it twice now. I have changed it in #429

PR #427, opened 2024-05-27
fix watcher cancellation

shotasilagadzetaal:


internal/k8s_watcher/watcher.go
```diff
@@ -146,27 +138,19 @@ func (c *Watcher) watchBlocktx() {
...
 			}
 		}
 	}()
 }
 
-func (c *Watcher) watchMetamorph() {
+func (c *Watcher) watchMetamorph(ctx context.Context) {
 	go func() {
-		defer func() {
-			if r := recover(); r != nil {
```
boecklim, line 160: Why remove the recovering with stack trace log?

internal/k8s_watcher/watcher.go
```diff
@@ -194,21 +178,18 @@ func (c *Watcher) watchMetamorph() {
...
-				return
 			}
 		}
 	}()
 }
 
 func (c *Watcher) Shutdown() {
-	c.tickerMetamorph.Stop()
-	c.shutdownMetamorph <- struct{}{}
-	<-c.shutdownMetamorphComplete
```
boecklim, line 210: If we don't wait for a signal when the go routine is finished, then I think the shutdown is not graceful anymore i.e. the logic from line 172 to 196 may be interrupted, not?

It could also be done with a wait group instead of a channel

PR #429, opened 2024-05-27
ARCO-26: fix lock issue on metamorph shutdown

arkadiuszos4chain:


cmd/arc/services/metamorph.go
```diff
@@ -280,7 +280,6 @@ func StartMetamorph(logger *slog.Logger) (func(), error) {
 			logger.Error("failed to shutdown mqClient", slog.String("err", err.Error()))
 		}
 
-		metamorphProcessor.Shutdown()
```
boecklim, line 283: What is the rationale behind removing the shutting down of the processor?

arkadiuszos4chain: It's called inside `server.Shutdown()`

boecklim: Yes, makes sense 👍

boecklim (approved):

PR #436, opened 2024-05-30
fix cancellation mechanism

shotasilagadzetaal:


internal/metamorph/stats_collector.go
```diff
@@ -127,30 +127,30 @@ func (p *Processor) StartCollectStats() error {
 		p.stats.healthyPeerConnections,
 	)
 	if err != nil {
+		p.waitGroup.Done()
 		return err
 	}
 	go func() {
 		defer func() {
 			if r := recover(); r != nil {
```
arkadiuszos4chain, line 135: You removed the logging stack from recover in other parts of the code. Shouldn't this one be removed too?

internal/metamorph/zmq.go
```diff
@@ -92,11 +91,6 @@ func (z *ZMQ) Start(zmqi ZMQI) error {
 	const invalidTxTopic = "invalidtx"
 	const discardedFromMempoolTopic = "discardedfrommempool"
 	go func() {
-		defer func() {
```
arkadiuszos4chain, line 95: Sorry, I missed it. Why don't we need these logs anymore?

internal/metamorph/stats_collector.go
shotasilagadzetaal: After removing this some tests failed. Still having the discussion about this with Michi.

internal/metamorph/zmq.go
shotasilagadzetaal: We simply tried to capture stack trace and log it FROM THE CODE. and we have 24 recovers. It's better to simply capture stderr if the application panics and log it from there.

arkadiuszos4chain (approved):

internal/metamorph/zmq.go
arkadiuszos4chain: Thanks

PR #440, opened 2024-06-04
BEEF Validation

kuba-4chain:


internal/beef/beef.go
```diff
@@ -27,6 +27,19 @@ const (
 type TxData struct {
 	Transaction *bt.Tx
 	BumpIndex   *bt.VarInt
+	txID        string
+}
+
+func (td *TxData) Unmined() bool {
+	return td.BumpIndex == nil
```
boecklim, line 34: One suggestion. How about calling this `Mined()` or `IsMined()`

```
func (td *TxData) Mined() bool {
	return td.BumpIndex != nil
}
```

Then this would become simpler to read
```
	if !parent.Unmined() {
		*minedAncestors = append(*minedAncestors, parent)
		return nil
	}
```

kuba-4chain: Right, done :)

internal/blocktx/store/postgresql/verify_merkle_roots.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	qMerkleRoot := `
+		SELECT b.height FROM blocks b WHERE b.merkleroot = $1 AND b.height = $2 AND b.orphanedyn = false
+	`
+
+	var unverifiedBlockHeights []uint64
+
+	for _, mr := range merkleRoots {
+		var hash []byte
```
boecklim, line 39: It seems this hash is not used anywhere. Can we remove it?

pkg/blocktx/client.go
```diff
@@ -15,6 +15,18 @@ type BlocktxClient interface {
 	ClearBlocks(ctx context.Context, retentionDays int32) (int64, error)
 	ClearBlockTransactionsMap(ctx context.Context, retentionDays int32) (int64, error)
 	DelUnfinishedBlockProcessing(ctx context.Context, processedBy string) error
+	VerifyMerkleRoots(ctx context.Context, merkleRootVerificationRequest []MerkleRootVerificationRequest) ([]uint64, error)
```
arkadiuszos4chain, line 18: `BlocktxClient` interface doesn't need `VerifyMerkleRoots` method

pkg/blocktx/client.go
```diff
@@ -15,6 +15,18 @@ type BlocktxClient interface {
...
+}
+
+// MerkleRootsVerificator verifies the merkle roots existance in blocktx db and returns unverified block heights.
+type MerkleRootsVerificator interface {
```
arkadiuszos4chain, line 22: I saw in other parts of the code that some interfaces end with the 'I' character. Shouldn't it be called `MerkleRootsVerificatorI`? @boecklim

pkg/blocktx/client.go
kuba-4chain: Unfortunately it does, because the method `blocktx.NewClient()` returns this interface and we want to pass blocktx as `MerkleRootsVerificator` into API handler, for example in `/cmd/arc/services/api.go` line `126`. So it needs to implement the method `VerifyMerkleRoots` from `MerkleRootsVerificator` interface.

pkg/blocktx/client.go
kuba-4chain: `BlocktxClient`, `TransactionHandler`, `TransactionMaintainer` are all interfaces without `I` character, so I decided to make `MerkleRootsVerificator` without this character as well, because it's in the same category as those

internal/blocktx/store/postgresql/verify_merkle_roots.go
kuba-4chain: Done

internal/validator/default/default_validator.go
```diff
@@ -33,20 +33,62 @@ func New(policy *bitcoin.Settings) validator.Validator {
...
+		return err
+	}
+
+	// everything checks out
+	return nil
+}
+
+func (v *DefaultValidator) ValidateBeef(beefTx *beef.BEEF, skipFeeValidation, skipScriptValidation bool) error {
+	for _, btx := range beefTx.Transactions {
+		if !btx.IsMined() {
```
boecklim, line 55: You can save one indentations if you do check `IsMined()`.
I find fewer indentations usually a bit easier to read.

```
	for _, btx := range beefTx.Transactions {
		if btx.IsMined() {
			continue
		}
		
		tx := btx.Transaction

		// needs to be calculated here, because txs in beef are not in EF format
		if !skipFeeValidation {
			if err := v.validateBeefFees(tx, beefTx); err != nil {
				return err
			}
		}

		// needs to be calculated here, because txs in beef are not in EF format
		if !skipScriptValidation {
			if err := beef.ValidateScripts(tx, beefTx.Transactions); err != nil {
				return validator.NewError(err, api.ErrStatusUnlockingScripts)
			}
		}

		// purposefully skip the fee and scripts validation, because it's done above
		if err := v.validateTransaction(tx, true, true); err != nil {
			return err
		}
	}
```

boecklim (approved):

pkg/blocktx/client.go
boecklim: Yes, I admit it's not completely consistent everywhere.
For now I suggest to name it without the `I`.

internal/validator/default/default_validator.go
kuba-4chain: Sure, done

pkg/blocktx/client.go
arkadiuszos4chain: ehh, who needs interface segregation...

arkadiuszos4chain (approved):

pkg/api/handler/default.go
```diff
@@ -48,12 +50,13 @@ func WithCallbackUrlRestrictions(rejectedCallbackUrlSubstrings []string) func(*A
 
 type Option func(f *ArcDefaultHandler)
 
-func NewDefault(logger *slog.Logger, transactionHandler metamorph.TransactionHandler, policy *bitcoin.Settings, opts ...Option) (api.ServerInterface, error) {
+func NewDefault(logger *slog.Logger, transactionHandler metamorph.TransactionHandler, merkleRootsVerificator blocktx.MerkleRootsVerificator, policy *bitcoin.Settings, opts ...Option) (api.ServerInterface, error) {
```
boecklim, line 53: Just saw now. I think "verificator" is not the correct word.
Shouldn't it be "verifier"?

boecklim (approved):

pkg/api/handler/default.go
kuba-4chain: Correct, done

PR #442, opened 2024-06-05
add stats for callbacks

shotasilagadzetaal:


internal/metamorph/callbacker.go
```diff
@@ -30,15 +31,17 @@ type Callback struct {
...
 }
 
-func NewCallbacker(httpClient HttpClient) *Callbacker {
+func NewCallbacker(httpClient HttpClient, processorStats *ProcessorStatsCollector) *Callbacker {
 	return &Callbacker{
-		httpClient: httpClient,
+		httpClient:     httpClient,
+		processorStats: processorStats,
 	}
 }
```
boecklim, line 46: I think these metrics could be added in `callbacker.go` to another struct like `callbackerStats`. Then there is no need to change `stats_collector.go` and making that type public

They can be registered in the constructor with `prometheus.Register()` and unregistered in `Shutdown()`

```
	CallbackSeenOnNetworkCount       prometheus.Gauge
	CallbackSeenInOrphanMempoolCount prometheus.Gauge
	CallbackRejectedCount            prometheus.Gauge
	CallbackMinedCount               prometheus.Gauge
	CallbackFailedCount              prometheus.Gauge
```

Also the type can be `prometheus.Counter`, as the value can only go up and never down.

internal/metamorph/callbacker.go
```diff
@@ -30,16 +40,52 @@ type Callback struct {
...
-	mu         sync.Mutex
-	disposed   bool
+	httpClient      HttpClient
+	wg              sync.WaitGroup
+	mu              sync.Mutex
+	disposed        bool
+	callbackerStats *CallbackerStats
 }
 
 func NewCallbacker(httpClient HttpClient) *Callbacker {
```
boecklim, line 50: If there is a possible error in this function, then it should return that error

```
func NewCallbacker(httpClient HttpClient) (*Callbacker, error) {...}
```

internal/metamorph/callbacker.go
```diff
@@ -30,16 +40,52 @@ type Callback struct {
...
+
+	err := registerStats(
+		callbacker.callbackerStats.callbackSeenOnNetworkCount,
+		callbacker.callbackerStats.callbackSeenInOrphanMempoolCount,
+		callbacker.callbackerStats.callbackRejectedCount,
+		callbacker.callbackerStats.callbackMinedCount,
+		callbacker.callbackerStats.callbackFailedCount,
+	)
+	if err != nil {
+		return nil
```
boecklim, line 85: `return nil, err`

PR #444, opened 2024-06-06
ARCO-71: inserted at num refactor

arkadiuszos4chain:
* rename `metamorph.transaction.inserted_at_num` -> `last_submitted_at` 
* change column type from INT -> TIMESTAMPZ


internal/metamorph/processor.go
```diff
@@ -615,3 +605,20 @@ func (p *Processor) Health() error {
 
 	return nil
 }
+
+func (p *Processor) storeData(ctx context.Context, data *store.StoreData) error {
```
boecklim, line 609: This function seems to be never returning an error

How about this as an alternative?

```
func (p *Processor) storeData(ctx context.Context, data *store.StoreData) error {
	nowNum := p.now()
	data.LastSubmittedAt = nowNum

	err := p.store.Set(ctx, data.Hash[:], data)
	if err != nil {
		return err
	}

	return nil
}
```

and when it is used on L 524

```
err = p.storeData(ctx, req.Data)
if err != nil {
     p.logger.Error("Failed to store transaction", slog.String("hash", data.Hash.String()), slog.String("err", err.Error()))
}

```

And on L 550 the function has to return with an error, because that is when the tx is stored initially. 
I see that this is a different logic than before, but I think it should be corrected like that.

```
err = p.storeData(ctx, req.Data)
if err != nil {
     return fmt.Errorf("failed to store transaction: %v", err)
}
```

arkadiuszos4chain: I believe it returns the error; the error is simply logged inside the function if it appears. If you are sure I should change the logic in line 550, I'll change it according to your suggestion

internal/metamorph/store/postgresql/migrations/000012_last_submitted_at_num.up.sql
boecklim, line 1: Wouldn't it be easier to do all the steps in one migration? Then you'll have to create and drop the index just once.

internal/metamorph/processor.go
boecklim: Yes, when on L 550 storing the data fails, then we should return there with an error. So that the user knows something is wrong with ARC in case of the DB not being available. Otherwise the user thinks it's all fine and keeps submitting transactions.

internal/metamorph/store/postgresql/migrations/000012_last_submitted_at_num.up.sql
arkadiuszos4chain: Good point, I've merged migrations

internal/metamorph/processor.go
arkadiuszos4chain: Changed

boecklim (approved):

PR #460, opened 2024-06-11
Additional test of BEEF validation

sirdeggen:
I created an example here of a BEEF tx where the most recent transaction is totally fine, but its parent hasn't been mined and doesn't pay any fees at all. Just like the handcash scenario we had earlier this year.

Leaving this un-merged until we can handle this situation with an error. Will continue to add examples here.

FAO @kuba-4chain


kuba-4chain:
I fixed the tests, in case of my test, it was the test name that was wrong, not the expected error message.

As for your test, I saw you include the expected error message
```
Fees paid by 5ef36123528b73da1bba4f329ef615188a9d8bedeadfd379c387a6d41c5042e2 are insufficient.
```
The `error.Detail` field is more like a short, one-sentence error, such as `Fees are too low`. 

So, do you want Arc to include which rawTx transaction the error applies to?

sirdeggen:
I approve, but can't approve my own PR - thank you! please proceed

boecklim (approved):

PR #465, opened 2024-06-13
feat(ARCO-90): new config flow

kuba-4chain:
### What has been done
1. Default config is specified in code, loaded in `main.go` and passed down to components that need it,
    -   this removes the need for getting the values out of viper manually,
    -   this removes the viper reference from any file other than those in `config/` dir,
    -   this allows for overriding only the specified values from a custom `config.yaml` (look at changes in README),
    -   this allows for removal of entire `internal/cfg/` directory, in favour of `config/utils.go`.
2. `main.go` and all services in `arc/service/` were refactored and cleaned up.
3. A feature of verifying that all values from config in code are presented in `config/example_config.yaml` was added to pipeline.
4. A flag to dump config was added.
5. Optional values were marked as such in `example_config.go` and removed (or set to 0 / "") from default config.
6. Readme was updated with an example.

### New Config Flow
The new flow of adding a new config variable looks like this now:
1. Add new variable to `config/config.go` (remember about `mapstruct` annotation)
2. Specify a default value (or nil if optional) in `config/defaults.go`
3. Take that value from `arcConfig` in code whenever you need it.
4. Add that value to `config/example_config.yaml` with an informative description.


config/config.go
boecklim, line 1: What are the json tags used for?

scripts/compare_yamls.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	newConfigFile, err := os.ReadFile("config/dumped_config.yaml")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// unmarshal ymal
```
boecklim, line 22: nit: yaml

scripts/compare_yamls.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	exampleConfig = convertToLowercase(exampleConfig)
+
+	newConfig := make(map[string]interface{})
+	if err := yaml.Unmarshal(newConfigFile, &newConfig); err != nil {
+		log.Fatal(err)
+	}
+
+	// iterate key in dumped config
+	for new_key := range newConfig {
```
boecklim, line 35: nit: use camel case => `newKey`

scripts/compare_yamls.go
```diff
@@ -0,0 +1,50 @@
...
+		log.Fatal(err)
+	}
+	exampleConfig = convertToLowercase(exampleConfig)
+
+	newConfig := make(map[string]interface{})
+	if err := yaml.Unmarshal(newConfigFile, &newConfig); err != nil {
+		log.Fatal(err)
+	}
+
+	// iterate key in dumped config
```
boecklim, line 34: I think comparison could be done more simple using e.g. `DeepEqual()`
https://www.geeksforgeeks.org/comparing-maps-in-golang/

scripts/compare_yamls.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// unmarshal ymal
+	exampleConfig := make(map[string]interface{})
+	if err := yaml.Unmarshal(exampleConfigFile, &exampleConfig); err != nil {
+		log.Fatal(err)
+	}
+	exampleConfig = convertToLowercase(exampleConfig)
```
boecklim, line 27: What happens if example config is not converted to lowercase?
I think that should not be necessary, not?

config/config.go
kuba-4chain: For configuration in json format, but since we're using yaml only, I removed them

scripts/compare_yamls.go
kuba-4chain: Thanks, fixed

scripts/compare_yamls.go
kuba-4chain: Right, done

scripts/compare_yamls.go
kuba-4chain: It could, but then we will not see exactly which key is missing in the example_config, as `DeepEqual()` only return true/false. But I can change it to DeepEqual if you think it's a better fit.

scripts/compare_yamls.go
kuba-4chain: Viper is case-insensitive. It treats all values as lowercase, and therefore dumps the config in all lowercase (try to run `go run ./cmd/main/arc -dump_config=./dumped_config.yaml` and look at the keys).

Maps in go, on the other hand, are case-sensitive. So, if we use cammelCase in `example_config.go`, it will be reflected in the unmarshalled map. Therefore, the 2 configs will not be equal in terms of maps in go:
```yaml
logLevel: INFO
loglevel: INFO
```
but they will be perfectly equal in viper, which is what we want to check here.

That's why I'm converting the example config to lowercase :)

config/load.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+func Load(configFileDirs ...string) (*ArcConfig, error) {
+	arcConfig := getDefaultArcConfig()
+
+	err := setDefaults(arcConfig)
+	if err != nil {
+		return nil, err
+	}
+
```
boecklim, line 19: If we add here this

```
	for _, configFileDir := range configFileDirs {
		viper.AddConfigPath(configFileDir)
	}
```

and the default value to "." on `cmd/arc/main.go:161`

```
configDir := flag.String("config", ".", "path to configuration yaml file")
```

Then I think we wouldn't need the change in `test/docker-compose.yml`, right? 🤔 

I think that may be better, because otherwise the `-config=...` parameter will be always required to start ARC and that may be a breaking change for operators of ARC and is not compatible with the instructions here: https://hub.docker.com/r/bsvb/arc/

kuba-4chain: The beauty of having config in code is that we don't need the `config.yaml` at all. We can just run Arc without any config and it will run with the defaults. So no breaking changes here, except the fact that we can simplify that command and run in without passing any config to the docker (but I think it should stay that way to show how you can override the default config using docker)

There are 2 issues with having `.` as the default value for the config path:
1. The default config will always be overridden by the config in the root directory, even if the user didn't intentionally specify the `-config` flag.
2. If a user specifies the flag `-config`, but the path that he provided doesn't have any config file, we want to show user the error of e.g. `config not found in directory ...`.

The problem is that if we leave `.` as the default value, and there will be no config.yaml, because the user did not specify the `-config` flag, arc will return an error, because it will look for that config file. But Arc can start with no config.yaml (defaults are in code), so it shouldn't return an error when the flag is not passed.

One thing that this is possible to solve with `.` as the default value is to NOT inform the user when the config.yaml file does not exist and just run with the defaults when we can't find a config file. But I don't think that's a correct approach, because it may confuse the user a lot.

So, I decided that this semantic would be the best, having the current solution:
1. If the `-config` flag is not provided, assume user wants to use default config - don't search for any file in any directory.
2. If the `-config` flag is provided, assume user wants to override the default values with his values - look for that config file and throw an error is it can't be found.

In case of tests, the `-config=...` flag is needed, because tests have a config file that is a bit different from the defaults, that's why I added it to the docker-compose.

boecklim (approved):

arkadiuszos4chain (approved):

PR #468, opened 2024-06-17
ARCO-101: Store all txs of a mined block in blocktx

arkadiuszos4chain:


internal/blocktx/store/postgresql/postgres_test.go
```diff
@@ -611,3 +444,221 @@ func TestPostgresDB(t *testing.T) {
...
+	}
+
+	// common setup for test cases
+	ctx, _, sut := setupPostgresTest(t)
+	defer sut.Close()
+	sut.maxPostgresBulkInsertRows = 5
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			prepareDb(t, sut.db, "fixtures/update_block_transactions")
```
boecklim, line 531: I think we should rename the folder `fixtures/update_block_transactions` to `fixtures/upsert_block_transactions`

boecklim (approved):

internal/blocktx/store/postgresql/postgres_test.go
arkadiuszos4chain: Right

PR #485, opened 2024-06-27
feat(ARCO-139): submit unmined parent txs in beef with callbacks

kuba-4chain:
When sending BEEF to Arc with unmined ancestors of a transaction, ARC will return a status of the last transaction in BEEF, but will submit and register callbacks for all unmined transactions in BEEF.

### DIFF
--- a/internal/beef/beef.go
+++ b/internal/beef/beef.go
@@ -59,20 +59,20 @@ func CheckBeefFormat(txHex []byte) bool {
 	return true
 }
 
-func DecodeBEEF(beefHex []byte) (*bt.Tx, *BEEF, []byte, error) {
+func DecodeBEEF(beefHex []byte) (*BEEF, []byte, error) {
 	remainingBytes, err := extractBytesWithoutVersionAndMarker(beefHex)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	bumps, remainingBytes, err := decodeBUMPs(remainingBytes)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	transactions, remainingBytes, err := decodeTransactionsWithPathIndexes(remainingBytes)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	decodedBeef := &BEEF{
@@ -80,7 +80,7 @@ func DecodeBEEF(beefHex []byte) (*bt.Tx, *BEEF, []byte, error) {
 		Transactions: transactions,
 	}
 
-	return decodedBeef.GetLatestTx(), decodedBeef, remainingBytes, nil
+	return decodedBeef, remainingBytes, nil
 }
 
 func (d *BEEF) GetLatestTx() *bt.Tx {
--- a/pkg/api/handler/helpers.go
+++ b/pkg/api/handler/helpers.go
@@ -123,3 +123,28 @@ func convertMerkleRootsRequest(beefMerkleRoots []beef.MerkleRootVerificationRequ
 
 	return merkleRoots
 }
+
+func findLastStatus(lastTxId string, statuses []*metamorph.TransactionStatus) *metamorph.TransactionStatus {
+	for _, status := range statuses {
+		if status.TxID == lastTxId {
+			return status
+		}
+	}
+	return nil
+}
+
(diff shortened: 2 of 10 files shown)


internal/beef/beef.go
```diff
@@ -59,28 +59,28 @@ func CheckBeefFormat(txHex []byte) bool {
...
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	decodedBeef := &BEEF{
 		BUMPs:        bumps,
 		Transactions: transactions,
 	}
 
-	return decodedBeef.GetLatestTx(), decodedBeef, remainingBytes, nil
```
arkadiuszos4chain, line 83: <3

test/beef_test.go
```diff
@@ -34,17 +34,27 @@ func TestBeef(t *testing.T) {
...
+			beef, middleTx, tx := prepareBeef(t, txID, hash, address, dstAddress, privateKey)
 
 			body := api.POSTTransactionJSONRequestBody{
 				RawTx: beef,
 			}
 
+			callbackReceivedChan := make(chan *api.TransactionStatus)
+			callbackErrChan := make(chan error)
+
+			callbackUrl, token, shutdown := startCallbackSrv(t, callbackReceivedChan, callbackErrChan, nil)
```
arkadiuszos4chain, line 49: Please move `startCallbackSrv(...)` to `utilis.go`

arkadiuszos4chain (changes requested):

pkg/api/handler/helpers.go
```diff
@@ -123,3 +123,28 @@ func convertMerkleRootsRequest(beefMerkleRoots []beef.MerkleRootVerificationRequ
...
+func findLastStatus(lastTxId string, statuses []*metamorph.TransactionStatus) *metamorph.TransactionStatus {
+	for _, status := range statuses {
+		if status.TxID == lastTxId {
+			return status
+		}
+	}
+	return nil
+}
+
+func filterStatusesOfInterest(txIds []string, allStatuses []*metamorph.TransactionStatus) []*metamorph.TransactionStatus {
```
boecklim, line 136: I'd rename this function to `filterStatusesByTxIDs(txIds []string, txStatuses []*metamorph.TransactionStatus)`

test/beef_test.go
```diff
@@ -62,6 +72,32 @@ func TestBeef(t *testing.T) {
 			statusResponse, err := arcClient.GETTransactionStatusWithResponse(context.Background(), tx.TxID())
 			require.NoError(t, err)
 			require.Equal(t, metamorph_api.Status_MINED.String(), *statusResponse.JSON200.TxStatus)
+
+			// verify callbacks for both unmined txs in BEEF
+			lastTxCallbackReceived := false
+			middleTxCallbackReceived := false
+
+			for i := 0; i < 2; i++ {
```
boecklim, line 80: Are 2 callbacks expected?
Maybe add a variable there like `expectedCallbacks := 2` to make it more explicit

pkg/api/handler/helpers.go
```diff
@@ -123,3 +123,28 @@ func convertMerkleRootsRequest(beefMerkleRoots []beef.MerkleRootVerificationRequ
 
 	return merkleRoots
 }
+
+func findLastStatus(lastTxId string, statuses []*metamorph.TransactionStatus) *metamorph.TransactionStatus {
```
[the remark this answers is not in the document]
kuba-4chain: Good point, done

test/beef_test.go
kuba-4chain: Done

test/beef_test.go
kuba-4chain: Yes, 2 callbacks are expected, one for the last tx in beef and one for the unmined parent, as per this feature request.

Thanks for pointing that out, magic numbers are the worst. Fixed

boecklim (approved):

test/beef_test.go
arkadiuszos4chain: IMO, making this magic number a variable doesn't make the code more readable or clear; moreover, it is misleading.
1. Making it a variable lets other developers think it is configurable via the variable, which is not true.
2. The expected callback number is hardcoded in the prepareBeef() function.

To make it truly readable, we should:
1. Change the prepareBeef() function implementation to make it possible to configure how many transactions it should contain (mined and unmined) – IMO, overengineering.
2. Return the number of "unmined" transactions from prepareBeef() – this way, we are hiding all the mess from the test code and move it to the setup code.

PR #491, opened 2024-07-03
Refactor/bulk set

boecklim:
- Do not pass context to processor => timeout doesn't stop storing of txs
- Set bulk db function
- Process txs submitted through message queue in batches

### DIFF
--- a/internal/metamorph/processor.go
+++ b/internal/metamorph/processor.go
@@ -37,8 +37,8 @@ const (
 	processStatusUpdatesIntervalDefault  = 500 * time.Millisecond
 	processStatusUpdatesBatchSizeDefault = 1000
 
-	bulkProcessBatchSizeDefault = 200
-	bulkProcessIntervalDefault  = 2 * time.Second
+	processTransactionsBatchSizeDefault = 200
+	processTransactionsIntervalDefault  = 2 * time.Second
 )
 
 type Processor struct {
@@ -79,7 +79,8 @@ type Processor struct {
 	processExpiredTxsInterval       time.Duration
 	processSeenOnNetworkTxsInterval time.Duration
 
-	bulkProcessInterval time.Duration
+	processTransactionsInterval  time.Duration
+	processTransactionsBatchSize int
 }
 
 type Option func(f *Processor)
@@ -125,8 +126,9 @@ func NewProcessor(s store.MetamorphStore, pm p2p.PeerManagerI, statusMessageChan
 		stats:                         newProcessorStats(),
 		waitGroup:                     &sync.WaitGroup{},
 
-		statCollectionInterval: statCollectionIntervalDefault,
-		bulkProcessInterval:    bulkProcessIntervalDefault,
+		statCollectionInterval:       statCollectionIntervalDefault,
+		processTransactionsInterval:  processTransactionsIntervalDefault,
+		processTransactionsBatchSize: processTransactionsBatchSizeDefault,
 	}
 
 	p.logger = slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: LogLevelDefault})).With(slog.String("service", "mtm"))
@@ -216,7 +218,7 @@ func (p *Processor) StartProcessMinedCallbacks() {
 
 func (p *Processor) StartProcessSubmittedTxs() {
 	p.waitGroup.Add(1)
-	ticker := time.NewTicker(p.bulkProcessInterval)
+	ticker := time.NewTicker(p.processTransactionsInterval)
 	go func() {
 		defer p.waitGroup.Done()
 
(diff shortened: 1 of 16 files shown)


internal/metamorph/processor.go
```diff
@@ -615,6 +628,46 @@ func (p *Processor) ProcessTransaction(ctx context.Context, req *ProcessorReques
 	}
 }
 
+func (p *Processor) BulkProcessTransaction(reqs []*ProcessorRequest) {
```
kuba-4chain, line 631: I'm not sure about this, but maybe we could change the name of this function to simply `ProcessTransactions`? Since we already have `ProcessTransaction`, it would be consistent with other methods in Arc, such as:

`SubmitTransaction`
`SubmitTransactions`
`PutTransaction`
`PutTransactions`

internal/metamorph/processor_options.go
```diff
@@ -69,6 +69,12 @@ func WithProcessStatusUpdatesInterval(d time.Duration) func(*Processor) {
 	}
 }
 
+func WithBulkProcessInterval(d time.Duration) func(*Processor) {
+	return func(p *Processor) {
+		p.bulkProcessInterval = d
+	}
+}
+
```
kuba-4chain, line 77: I'm thinking if we're allowing to configure the interval, maybe we should also make the bulk size configurable, instead of having it always set to `200`?

Also, I was thinking that `WithProcessTransactionsInterval` would sound good as well, but it's always hard with naming 😄

kuba-4chain (approved):

internal/metamorph/processor.go
```diff
@@ -210,24 +216,29 @@ func (p *Processor) StartProcessMinedCallbacks() {
...
+				if len(reqs) > 0 {
+					p.BulkProcessTransaction(reqs)
+					reqs = []*ProcessorRequest{}
+				}
 			case submittedTx := <-p.submittedTxsChan:
 				if submittedTx == nil {
 					continue
 				}
-
 				hash := PtrTo(chainhash.DoubleHashH(submittedTx.GetRawTx()))
```
arkadiuszos4chain, line 239: Can be moved to line 240

internal/metamorph/processor.go
```diff
@@ -210,24 +216,29 @@ func (p *Processor) StartProcessMinedCallbacks() {
 
 func (p *Processor) StartProcessSubmittedTxs() {
 	p.waitGroup.Add(1)
-
+	ticker := time.NewTicker(p.bulkProcessInterval)
 	go func() {
 		defer p.waitGroup.Done()
+
+		var reqs []*ProcessorRequest
```
arkadiuszos4chain, line 225: I suggest to use a slice with some capacity declared already to avoid multiple reallocation
`reqs:= make([]*ProcessorRequest, 0, 100)`

internal/metamorph/processor.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	processStatusUpdatesIntervalDefault  = 500 * time.Millisecond
 	processStatusUpdatesBatchSizeDefault = 1000
+
+	processTransactionsBatchSizeDefault = 200
+	processTransactionsIntervalDefault  = 2 * time.Second
```
arkadiuszos4chain, line 41: Are you sure it is not to long interval?

internal/metamorph/processor.go
```diff
@@ -615,6 +631,46 @@ func (p *Processor) ProcessTransaction(ctx context.Context, req *ProcessorReques
...
+		err = p.mqClient.PublishRegisterTxs(req.Data.Hash[:])
+		if err != nil {
+			p.logger.Error("Failed to register tx in blocktx", slog.String("hash", req.Data.Hash.String()), slog.String("err", err.Error()))
+		}
+
+		// Announce transaction to network and save peers
+		peers := p.pm.AnnounceTransaction(req.Data.Hash, nil)
+		if len(peers) == 0 {
+			p.logger.Warn("transaction was not announced to any peer", slog.String("hash", req.Data.Hash.String()))
+			return
```
arkadiuszos4chain, line 662: Please change to `continue`

internal/metamorph/processor.go
```diff
@@ -615,6 +631,46 @@ func (p *Processor) ProcessTransaction(ctx context.Context, req *ProcessorReques
 	}
 }
 
+func (p *Processor) ProcessTransactions(reqs []*ProcessorRequest) {
```
arkadiuszos4chain, line 634: We can use `[]*store.StoreData`

internal/metamorph/processor.go
```diff
@@ -239,7 +252,12 @@ func (p *Processor) StartProcessSubmittedTxs() {
 					Timeout: time.Duration(submittedTx.MaxTimeout) * time.Second,
```
arkadiuszos4chain, line 252: We do not use `Timeout`, please remove

internal/metamorph/processor.go
boecklim: I changed it to 1s for now. We can later make a setting of it and tweak the value

arkadiuszos4chain (approved):

PR #498, opened 2024-07-05
Refactor/keyset topup

kameshraj23:
Refactor for keyset topup

Screenshot 2024-07-05 at 09 27 34

### DIFF
--- a/cmd/broadcaster-cli/app/keyset/address/address.go
+++ b/cmd/broadcaster-cli/app/keyset/address/address.go
@@ -1,13 +1,10 @@
 package address
 
 import (
-	"errors"
-	"fmt"
 	"log/slog"
 	"strings"
 
 	"github.com/bitcoin-sv/arc/cmd/broadcaster-cli/helper"
-	"github.com/bitcoin-sv/arc/pkg/keyset"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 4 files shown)


cmd/broadcaster-cli/app/keyset/topup/topup.go
```diff
@@ -30,19 +27,60 @@ var Cmd = &cobra.Command{
...
+		var keys map[string]string
+		err = viper.UnmarshalKey("privateKeys", &keys)
 		if err != nil {
 			return err
 		}
-		logger.Info("top up complete", slog.String("address", fundingKeySet.Address(!isTestnet)))
+
+		var keySets []*keyset.KeySet
+
+		if len(keys) > 0 {
```
boecklim, line 46: Could you move this logic into a separate helper function here `cmd/broadcaster-cli/helper/functions.go` so that it doesn't get repeated in each command?

kameshraj23: yes makes sense will move it

cmd/broadcaster-cli/app/keyset/topup/topup.go
```diff
@@ -30,19 +24,36 @@ var Cmd = &cobra.Command{
...
 		if err != nil {
-			return fmt.Errorf("failed to get key sets: %v", err)
+			return err
 		}
 
-		err = wocClient.TopUp(context.Background(), !isTestnet, fundingKeySet.Address(!isTestnet))
+		keySets, err := helper.GetKeySets(keys, selectedKeys)
 		if err != nil {
 			return err
 		}
```
arkadiuszos4chain, line 43: Just loud thinking: maybe it could be one function too? I mean we could have helper functions like:
`func GetPrivateKeys()`
`func GetKeySetsFor(keys map[string]string (...))`
`func GetKeySets()`

The first two would be for future flexibility, and the last one would be used in the currently existing implementation. What do you think?

cmd/broadcaster-cli/app/keyset/address/address.go
```diff
@@ -27,38 +24,14 @@ var (
...
-							return fmt.Errorf("failed to get key sets: %v", err)
-						}
-						keySets = append(keySets, fundingKeySet)
-					}
-				}
-			} else {
-				return errors.New("no keys given in configuration")
+			keySets, err := helper.GetKeySets(keys, selectedKeys)
+			if err != nil {
+				return err
```
arkadiuszos4chain, line 34: Just loud thinking: maybe it could be one function too? I mean we could have helper functions like:
`func GetPrivateKeys()`
`func GetKeySetsFor(keys map[string]string (...))`
`func GetKeySets()`

The first two would be for future flexibility, and the last one would be used in the currently existing implementation. What do you think?

kameshraj23: I have them as getKeySets and GetPrivateKeys , yeah if need be i can fragment and add one more function  GetKeySetsFor(keys map[string]string (...))

kameshraj23: we also have this function GetKeySetsKeyFile() in the helper which will resolve the issue ?

boecklim (approved):

PR #499, opened 2024-07-05
feat(ARCO-148): get txs data for peers in bulk

kuba-4chain:
This PR is an adaptation for this change in go-p2p: [feat!: improve performance of HandleTransactionGet](https://github.com/libsv/go-p2p/pull/25)


internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
...
+		callback_token,
+		full_status_updates,
+		reject_reason,
+		raw_tx,
+		locked_by,
+		merkle_path,
+		retries
+		FROM metamorph.transactions
+		WHERE hash IN (%s);`, strings.Join(placeholders, ", "))
+	return query, args
```
kuba-4chain, line 364: I'm still thinking whether this function `GetBulk()` makes much sense. It is getting exactly the same data that `Get()` but for multiple rows.

But in reality, a peer only needs rawTxs and this function is not used anywhere else. So, we could change this to `GetBulk()` -> `GetRawTxsByHashes()` or something like that and make the query much less complicated:
```suggestion
	query := fmt.Sprintf(`SELECT raw_tx FROM metamorph.transactions WHERE hash IN (%s);`, strings.Join(placeholders, ", "))
	return query, args
```

WDYT @boecklim @arkadiuszos4chain @shotasilagadzetaal ?

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
...
+				&callbackToken,
+				&fullStatusUpdates,
+				&rejectReason,
+				&rawTx,
+				&lockedBy,
+				&merklePath,
+				&retries,
+			)
+			if err != nil {
+				if errors.Is(err, sql.ErrNoRows) {
```
arkadiuszos4chain, line 253: Remove. `rows.Scan()` won't return `ErrNoNows` error

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
...
+			if retries.Valid {
+				data.Retries = int(retries.Int32)
+			}
+
+			retData = append(retData, data)
+		}
+
+		// Check for any errors encountered during iteration
+		if err := rows.Err(); err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
```
arkadiuszos4chain, line 322: Remove, `rows.Err()` won't return `sql.ErrNoRows`

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
...
+		callback_url,
+		callback_token,
+		full_status_updates,
+		reject_reason,
+		raw_tx,
+		locked_by,
+		merkle_path,
+		retries
+		FROM metamorph.transactions
+		WHERE hash IN (%s);`, strings.Join(placeholders, ", "))
```
arkadiuszos4chain, line 363: If it is not for performance reasons (if it is, please share some data with me ;-)), please change the `WHERE` clause to:
```
WHERE hash in (SELECT UNNEST($1::BYTEA[]))

File: internal/metamorph/store/postgresql/postgres.go [hunk 5]
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
 	return data, nil
 }
 
+// GetBulk implements the MetamorphStore interface. It attempts to get values for given keys.
+// If the keys do not exist an error is returned, otherwise the retrieved values.
+func (p *PostgreSQL) GetBulk(ctx context.Context, hashes [][]byte) ([]*store.StoreData, error) {
```
arkadiuszos4chain, line 197: IMO `GetMany` is a little better name ;-)

internal/metamorph/store/postgresql/postgres.go
arkadiuszos4chain: As I see it ,in this situation, we can opt for a specialized function and query the database only for the data we really need.
Firstly, as you yourself have noted, so far this is the only call to this method.
Secondly, we are trying to solve the bottleneck problem, and this could positively impact performance.
Thirdly, it will definitely improve code readability ;-)

Therefore, I suggest changing the function name to `GetManyRawTxs` or simply `GetRawTxs` (I am not a fan of adding `ByHashes` to the name - it would be inconsistent with the rest of the code).

arkadiuszos4chain (changes requested):

internal/metamorph/store/postgresql/postgres.go
kuba-4chain: You convinced me. I refactored the solution completely to make it much simpler, and hopefully faster.

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/libsv/go-bc v0.1.29
 	github.com/libsv/go-bk v0.1.6
 	github.com/libsv/go-bt/v2 v2.2.5
-	github.com/libsv/go-p2p v0.2.6
+	github.com/libsv/go-p2p v0.2.7
```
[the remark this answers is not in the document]
kuba-4chain: You're right.

I'll switch once the [new PR](https://github.com/libsv/go-p2p/pull/27) is merged and new version released.

internal/metamorph/peer_handler.go
```diff
@@ -69,21 +66,31 @@ func (m *PeerHandler) HandleTransactionRejection(rejMsg *wire.MsgReject, peer p2
...
-		Status: metamorph_api.Status_REQUESTED_BY_NETWORK,
-		Peer:   peer.String(),
+	for i, msg := range msgs {
+		m.messageCh <- &PeerTxMessage{
+			Hash:   &msg.Hash,
+			Status: metamorph_api.Status_REQUESTED_BY_NETWORK,
+			Peer:   peer.String(),
+		}
+
+		hashes[i] = msg.Hash.CloneBytes()
```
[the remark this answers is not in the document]
kuba-4chain: Perfect, thanks

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
...
+	retData := make([]*store.StoreData, 0)
+
+	// Batch process hashes to avoid SQL limits
```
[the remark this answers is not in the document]
kuba-4chain: > By default, the maximum query length in PostgreSQL is 1 GB. However, this value can be changed by modifying the max_query_size parameter in the postgresql.conf file.
[Source](https://www.squash.io/does-postgresql-have-a-maximum-sql-query-length/)

Anyways, the default 1GB should be enough for even the biggest inv msgs. I refactored this solution using `UNNEST` operator and removed the batching, so it should be fine for now.

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
...
+	batchSize := 1000
+	for i := 0; i < len(hashes); i += batchSize {
+		end := i + batchSize
+		if end > len(hashes) {
+			end = len(hashes)
+		}
+		batch := hashes[i:end]
```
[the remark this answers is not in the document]
kuba-4chain: Refactored to a new solution using `UNNEST`

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +192,178 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
 	return data, nil
 }
 
+// GetBulk implements the MetamorphStore interface. It attempts to get values for given keys.
+// If the keys do not exist an error is returned, otherwise the retrieved values.
+func (p *PostgreSQL) GetBulk(ctx context.Context, hashes [][]byte) ([]*store.StoreData, error) {
+	retData := make([]*store.StoreData, 0)
+
+	// Batch process hashes to avoid SQL limits
+	batchSize := 1000
```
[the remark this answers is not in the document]
kuba-4chain: Refactored to a new solution using `UNNEST`

internal/metamorph/store/postgresql/postgres.go
kuba-4chain: Refactored

internal/metamorph/store/postgresql/postgres.go
kuba-4chain: Changed, thanks :)

internal/metamorph/store/postgresql/postgres.go
kuba-4chain: Refactored

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -191,6 +191,38 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
...
+
+	q := `SELECT raw_tx
+		FROM metamorph.transactions
+		WHERE hash in (SELECT UNNEST($1::BYTEA[]))`
+
+	rows, err := p.db.QueryContext(ctx, q, pq.Array(hashes))
+	if err != nil {
+		return nil, err
+	}
+
```
arkadiuszos4chain, line 209:
```suggestion
defer rows.Close()
```

The `sql.Rows` returned by `sql.QueryContext()` should always be disposed after usage. I've noticed a few other instances in the file where this isn't done – could you please address those as well?

Since I am not an expert in Golang, I'm uncertain about when they should be closed in case of an error in a transaction. So please follow the same pattern as in the `GetSeenOnNetwork()` function (just add `defer rows.Close()` after `Rollback()`)."

Places to fix:
1. line 209
2. line 564
3. line 649
4. line 779

kuba-4chain: You're right, added, thanks for noticing.

arkadiuszos4chain (approved):
Well, now it is a performance improvement!

boecklim (approved):

PR #500, opened 2024-07-08
ARCO-155: (PART 1) separate tx validators

arkadiuszos4chain:
* divide `validator.Validator` interface into two specialized interfaces: `DefaultValidator` and `BeefValidator`
* move all BEEF related logic from `deafulvalidator.DefaultValidator` to the new dedicated struct `beef.BeefValidator`
* move all BEEF validation related logic from `beef/...` to `validator/beef/beef_validator.go`
* move and rename `beef/bump_validation.go` -> `validator/beef/beef_validator_test.go`
* common validation logic move  from `default_validator.go` to `validator/validation.go`
* add enums for validation configuration
* adjust default handler to use two validators

Please ignore `TODO` comments as they'll be solved in the follow- up PR


internal/validator/validator.go
kuba-4chain, line 1: Can we maybe change this to `interface.go`? Because we already have `validation.go` and it can become confusing. Or change `validation.go` -> `common_validation.go`

internal/validator/validator.go
```diff
@@ -15,10 +17,57 @@ type OutpointData struct {
...
+
+type DefaultValidator interface {
+	ValidateTransaction(ctx context.Context, tx *bt.Tx, feeValidation FeeValidation, scriptValidation ScriptValidation) error
 	IsExtended(tx *bt.Tx) bool
-	IsBeef(txHex []byte) bool
+}
+
+type BeefValidator interface {
+	ValidateTransaction(ctx context.Context, beef *beef.BEEF, feeValidation FeeValidation, scriptValidation ScriptValidation) (*bt.Tx, error)
+}
```
kuba-4chain, line 41: Why not make it one interface and use a strategy design pattern here?

internal/validator/default/default_validator.go
```diff
@@ -1,147 +1,53 @@
...
-	if !skipFeeValidation {
-		if err := checkFees(tx, api.FeesToBtFeeQuote(v.policy.MinMiningTxFee)); err != nil {
-			return validator.NewError(err, api.ErrStatusFees)
+	switch feeValidation {
+	case validator.StandardFeeValidation:
+		if err := standardCheckFees(tx, api.FeesToBtFeeQuote(v.policy.MinMiningTxFee)); err != nil {
+			return err
 		}
+	case validator.NoneFeeValidation:
+		fallthrough
```
kuba-4chain, line 45: Maybe we can remove this lines?

internal/validator/validation.go
```diff
@@ -0,0 +1,237 @@
...
+	// 	}
+	// default:
+	// }
+
+	// // 12) The unlocking scripts for each input must validate against the corresponding output locking scripts
+	// if scriptValidation == StandardScriptValidation {
+	// 	if err := checkScripts(tx); err != nil {
+	// 		return NewError(err, api.ErrStatusUnlockingScripts)
+	// 	}
+	// }
```
kuba-4chain, line 115: Remove this comments

internal/validator/validation.go
```diff
@@ -0,0 +1,237 @@
...
+	for index, input := range tx.Inputs {
+		if hex.EncodeToString(input.PreviousTxID()) == coinbaseTxID {
+			return NewError(fmt.Errorf("transaction input %d is a coinbase input", index), api.ErrStatusInputs)
+		}
+		/* lots of our valid test transactions have this sequence number, is this not allowed?
+		if input.SequenceNumber == 0xffffffff {
+			fmt.Printf("input %d has sequence number 0xffffffff, txid = %s", index, tx.TxID())
+			return validator.NewError(fmt.Errorf("transaction input %d sequence number is invalid", index), arc.ErrStatusInputs)
+		}
+		*/
```
kuba-4chain, line 165: Maybe we can remove this as well?

pkg/api/handler/default.go
```diff
@@ -388,12 +390,10 @@ func (m ArcDefaultHandler) processTransactions(ctx context.Context, transactions
...
 	txErrors := make([]interface{}, 0)
 
-	txValidator := defaultValidator.New(m.NodePolicy)
-
 	for len(transactionsHexes) != 0 {
-		isBeefFormat := txValidator.IsBeef(transactionsHexes)
+		hexFormat := validator.GetHexFormat(transactionsHexes)
 
-		if isBeefFormat {
+		if hexFormat == validator.Beef {
```
kuba-4chain, line 396: Maybe we should change `validator.Beef` to `validator.FormatBeef` or `validator.BeefFormat`, to avoid confusion

internal/validator/validator.go
arkadiuszos4chain: I really don't like files named 'interface.go', but you have a point. I'll try to change the name of the 'validation' file.

internal/validator/validator.go
arkadiuszos4chain: Let me break the answer into a few points:

1. This is a big PR with a significant number of changes (specifically moving the code around), and I wanted to make it as reviewable as possible. Simply dividing the existing interface into two was a way to limit changes in the client code.
2. I considered creating a unified validation interface, but this would lead to the creation of another input structure and the need to validate this input sturcture inside our specialized validators, which feels like "validation inception" – something I don't like. That said, I believe we can unify them on the client side using the `Adapter` pattern.

pkg/api/handler/default.go
arkadiuszos4chain: `BEEF` and `EF` are acronyms where F stands for `Format`, and I feel a bit anxious when I see names like `BeefFormat`
I understand that the simple `Beef` here can be misleading since we have a package with the same name and a `BeefData` structure. 
In this context, it's about hex formats, so what do you think about names like: `RawHex`, `EfHex`, `BeefHex`?

internal/validator/validator.go
kuba-4chain: Makes sense, approved ✅

pkg/api/handler/default.go
kuba-4chain: Okay, let's make this compromise and have `RawHex`, `EfHex` and `BeefHex` :)

internal/validator/validation.go
arkadiuszos4chain: removed

pkg/api/handler/default.go
arkadiuszos4chain: Enum names changed

kuba-4chain (approved):

internal/validator/validator.go
```diff
@@ -15,10 +17,57 @@ type OutpointData struct {
...
+	}
+
+	if isEf(hex) {
+		return EfHex
+	}
+
+	return RawHex
+}
+
+func isEf(hex []byte) bool {
```
boecklim, line 63: Now we have 2 functions to check if a tx is EF. 
The other one is this one: https://github.com/bitcoin-sv/arc/blob/60a3c1b587647645aa8e601a05fc0fd8457bb481/internal/validator/default/default_validator.go#L59:L71

Wouldn't it be better to have just one? 🤔

arkadiuszos4chain: The one you mentioned checks if `bt.Tx` structure contains all the necessary information to be treated as extended. This one checks the hex format sent by the client.
So on this stage we need both

boecklim (approved):

PR #502, opened 2024-07-09
ARCO-155: (PART 3) encapsulate all validation logic inside validators

arkadiuszos4chain:
* Move validation processes into validators (e.g., extending rawTx, verifying Merkle roots).
* Implement a new method in WoC Client for getting transactions in batches.
* Align some parts of the code to the Interface Segregation Principle.


pkg/api/handler/tx_finder.go
```diff
@@ -0,0 +1,97 @@
...
+	pc         *config.PeerRpcConfig
+	l          *slog.Logger
+	w          *woc_client.WocClient
+	useMainnet bool
+}
+
+func (f txFinder) GetRawTxs(ctx context.Context, ids []string) ([]validator.RawTx, error) {
+	// NOTE: we can ignore ALL errors from providers, if one returns err we go to another
+
+	// TODO: discuss if it's worth to have implementation for len(ids) == 1
```
kuba-4chain, line 25: I don't think it's worth it. I would leave it like that

arkadiuszos4chain: Yeah, you may be right

internal/validator/common_validation.go
```diff
@@ -47,7 +47,7 @@ func CalculateMiningFeesRequired(size *bt.TxSize, fees *bt.FeeQuote) (uint64, er
 	return feesRequiredRounded, nil
 }
 
-func CommonValidateTransaction(policy *bitcoin.Settings, tx *bt.Tx) error {
+func CommonValidateTransaction(policy *bitcoin.Settings, tx *bt.Tx) *Error {
```
kuba-4chain, line 50: I really really don't like that this returns `*Error`. I know this is defined in the same package, so it's internal in theory, but it's exported and looks like it can be used in other packages and it shouldn't IMO, because it's an error made just for validator.

Can we revert the name back to `ValidationError` / `ValidatorError`? 🙏

internal/validator/default/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/bitcoin-sv/arc/internal/validator"
+	"github.com/libsv/go-bt/v2"
+)
+
+var errParentNotFound = errors.New("parent transaction not found")
+
+func extendTx(ctx context.Context, w validator.TxFinderI, rawTx *bt.Tx) error {
```
kuba-4chain, line 14: The same as for `BeefValidator`, why not make it the method of `DefaultValidator` and access its `TxFinder`?

internal/validator/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+
+type RawTx struct {
+	TxID    string
+	Bytes   []byte
+	IsMined bool
+}
+
+type MerkleVerifierI interface {
+	Verify(ctx context.Context, request []beef.MerkleRootVerificationRequest) ([]uint64, error)
+}
```
kuba-4chain, line 29: I think these are important definitions of important structs, consts and interfaces. I would move it to `validator.go` with other important definitions, and leave just helper functions `GetHexFormat()` and `ifEf()` here. WDYT?

pkg/api/handler/default.go
```diff
@@ -63,15 +64,34 @@ func NewDefault(
...
-		logger:              logger,
-		now:                 time.Now,
-		peerRpcConfig:       peerRpcConfig,
-		apiConfig:           apiConfig,
+		TransactionHandler: transactionHandler,
+		NodePolicy:         policy,
+		logger:             logger,
+		now:                time.Now,
+		txFinder:           &finder,
+		mrVerifier:         mr,
```
kuba-4chain, line 94: Why the pointer `&finder` here?

pkg/api/handler/merkle_verifier_adapter.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/bitcoin-sv/arc/internal/beef"
+	"github.com/bitcoin-sv/arc/pkg/blocktx"
+)
+
+type merkleVerifier struct {
+	v blocktx.MerkleRootsVerifier
+}
+
+func (a merkleVerifier) Verify(ctx context.Context, request []beef.MerkleRootVerificationRequest) ([]uint64, error) {
```
kuba-4chain, line 14: `a` like `api`? Maybe `m` or `mv` will be better?

arkadiuszos4chain: `a` like adapter

examples/customhandler/handler.go
```diff
@@ -36,11 +36,18 @@ func NewCustomHandler() (api.ServerInterface, error) {
 	// add blocktx, header service or custom implementation of merkle roots verifier
 	merkleRootVerifier := &CustomMerkleRootsVerifier{}
 
+	// create default handler
```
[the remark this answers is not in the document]
arkadiuszos4chain: IMO, it's not a relatable comment. Custom Handle is a wrapper/adapter on the default handler, and in line 40 we create the default handler instance.

Maybe better option is adding another comment at line 48?

internal/validator/beef/beef_validator.go
```diff
@@ -121,18 +124,30 @@ func validateScripts(tx *bt.Tx, inputTxs []*beef.TxData) error {
 	return nil
 }
 
-// TODO move to common
+func verifyMerkleRoots(ctx context.Context, v validator.MerkleVerifierI, beefTx *beef.BEEF) *validator.Error {
```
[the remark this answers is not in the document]
arkadiuszos4chain: There is no need to copy whole validator structure

internal/validator/common_validation.go
boecklim: @arkadiuszos4chain What is the difference between returning `error` or `*Error` here? 🤔 
I think the logic would be the same.

internal/validator/default/default_validator.go
```diff
@@ -56,7 +60,17 @@ func (v *DefaultValidator) ValidateTransaction(ctx context.Context, tx *bt.Tx, f
 	return nil
 }
 
-func (v *DefaultValidator) IsExtended(tx *bt.Tx) bool {
+func needsExtention(tx *bt.Tx, fv validator.FeeValidation, sv validator.ScriptValidation) bool {
```
boecklim, line 63: Typo: `needsExtension`

internal/validator/default/default_validator.go
```diff
@@ -7,25 +7,29 @@ import (
...
+		txFinder: finder,
 	}
 }
 
 func (v *DefaultValidator) ValidateTransaction(ctx context.Context, tx *bt.Tx, feeValidation validator.FeeValidation, scriptValidation validator.ScriptValidation) error { //nolint:funlen - mostly comments
 	// 0) Check whether we have a complete transaction in extended format, with all input information
 	//    we cannot check the satoshi input, OP_RETURN is allowed 0 satoshis
-	if !v.IsExtended(tx) {
-		return validator.NewError(fmt.Errorf("transaction is not in extended format"), api.ErrStatusTxFormat)
+	if needsExtention(tx, feeValidation, scriptValidation) {
```
boecklim, line 28: If a tx will only be extended if either a fee or script validation happens it may be good to have a small e2e test which ensures that sending a **standard** format tx while **skipping script and fee validation** still works.

A template for that can be found here: https://github.com/bitcoin-sv/arc/blob/2341e4ae688b05e32f1d05d16aeae2e4f712b707/test/endpoint_test.go#L317:L376

internal/validator/default/testdata/fixture.go
```diff
@@ -0,0 +1,41 @@
...
+	"encoding/hex"
+
+	"github.com/bitcoin-sv/arc/internal/validator"
+)
+
+const (
+	ValidTxID     = "900d18f42034586544fb6876e712b2621f14ca3f7e639018ef2e4ed6e22caf25"
+	ValidTxRawHex = "01000000023df7585811e737e1607b95f32f01574c5a9ce49bb872528a88cb2c7a372b6dec000000006b483045022100809f8409850ed8e1a4ad0abee60fe6e60707eff40af08f55cbb5af80ec363b8e02203204193b7a563463f8844be198228d1356d1ee0a25a73662c7488f45f046936dc321033df74ab4534eef6cb47badf56c115f627e87fd6be47f426ee63a3e94920d38ccffffffffeb48bd7fb162a94b41728b3d29d5aec5f2bc56eac46c026ee053642fd6ff1acc010000006a4730440220154a0a8b4096b7137dcfde373d262f007fd43bc6a127ffc82c1b851020cdde7b02202dbefd508892de0b82a967b0e95c415981b661dc07aab86929cba07405d5d5adc32103128c33a42ed19cf84fb7342d86f9344ab966e4d182bd678ecf9adbe63b162591ffffffff0201000000000000001976a914b44bcdac5ad9657b4558da182bf0ec2ba761291588ac803f0f00000000001976a914747a796c43d2953f9fde6ccbc302f8e8a5e1417088ac00000000"
+
+	ParentTx_1_ID     = "cc1affd62f6453e06e026cc4ea56bcf2c5aed5293d8b72414ba962b17fbd48eb"
```
boecklim, line 13: I'd prefer keeping variable names camel case. Maybe: `ParentTxID1`

kuba-4chain:
Let's remove the file `internal/validator/default/script_debugger.go`. It's long unused. It's about debugging the script validator, which is another lib in itself and debugging it should not be part of Arc.

internal/validator/default/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		inputs = append(inputs, in)
+		parentInputMap[prevTxID] = inputs
+	}
+
+	// get parents
+	parentsTxs, err := w.GetRawTxs(ctx, parentsIDs)
+	if err != nil {
+		return fmt.Errorf("cannot extend tx: %w", err)
```
boecklim, line 39: I think the error message is not correct
Maybe this "Failed to get raw transactions for parent IDs: %w" ?

internal/woc_client/woc_client.go
```diff
@@ -276,3 +286,61 @@ func (w *WocClient) TopUp(ctx context.Context, mainnet bool, address string) err
...
+		result = append(result, txs...)
+		ids = ids[bsize:] // move batch window
+	}
+
+	return result, nil
+}
+
+func (w *WocClient) getRawTxs(ctx context.Context, net string, batch []string) ([]*wocRawTx, error) {
+	payload := map[string][]string{"txs": batch}
+	body, _ := json.Marshal(payload)
```
boecklim, line 318: There is an unhandled error

internal/woc_client/woc_client.go
```diff
@@ -276,3 +286,61 @@ func (w *WocClient) TopUp(ctx context.Context, mainnet bool, address string) err
...
+
+		result = append(result, txs...)
+		ids = ids[bsize:] // move batch window
+	}
+
+	return result, nil
+}
+
+func (w *WocClient) getRawTxs(ctx context.Context, net string, batch []string) ([]*wocRawTx, error) {
+	payload := map[string][]string{"txs": batch}
```
boecklim, line 317: The payload has key "txids"

```
payload := map[string][]string{"txids": batch}
```

https://docs.taal.com/core-products/whatsonchain/transaction#bulk-raw-transaction-data

internal/validator/helpers.go
arkadiuszos4chain: These interfaces are on a different abstraction level; they represent a deeper abstraction. That's why I don't want to mix them with the content from `validator.go`. Ideally, I would put them in the `default` and `beef` packages, but right now I cannot due to cyclic dependencies in the mocks.

If you have a better name for the file than `helper.go`, I will change it, but I'm against moving them to the `validator.go`.

pkg/api/handler/default.go
arkadiuszos4chain: Because it's a 41-byte struct, note that `ArcDefaultHandler` is used directly in its method, so I thought it would be beneficial to use a pointer here.

internal/woc_client/woc_client.go
arkadiuszos4chain: Please correct me if I'm wrong, but `json.Marshal` would return an error only if the passed object is invalid JSON (e.g., circular references, invalid string, etc.). Here, the JSON data is created by me (the developer), and it's a one-key map. In my opinion, checking for an error here is unnecessary - WDYT?

internal/validator/common_validation.go
arkadiuszos4chain: @kuba-4chain I didn't change the name of the type. I think using a name like `ValidationError` or `ValidatorError` may not be idiomatic, as in client code it would look like `validator.ValidationError`. That said, it's not crucial for me, and if you think (and no one is against it), we could change the names.

@boecklim what I wanted to achieve with this change is to provide a clear understanding of when a client should wrap the error from these functions with `validator.Error`. Please take a look at the previous implementation:
``` go
func (v *DefaultValidator) ValidateTransaction(ctx context.Context, tx *bt.Tx, feeValidation validator.FeeValidation, scriptValidation validator.ScriptValidation) error { 
	if !v.IsExtended(tx) {
	       // Arek: note we wrap error here - why?
		return validator.NewError(fmt.Errorf("transaction is not in extended format"), api.ErrStatusTxFormat)
	}

	err := validator.CommonValidateTransaction(v.policy, tx)
	if err != nil {
	        // Arek: we do not wrap error - why?
		return err
	}

	switch feeValidation {
	case validator.StandardFeeValidation:
		if err := standardCheckFees(tx, api.FeesToBtFeeQuote(v.policy.MinMiningTxFee)); err != nil {
		          // Arek: no wrapping here - why?
			return err
		}
	case validator.NoneFeeValidation:
		// Do not handle the default case on purpose; we shouldn't assume that other types of validation should be omitted
	}

	// 12) The unlocking scripts for each input must validate against the corresponding output locking scripts
	if scriptValidation == validator.StandardScriptValidation {
		if err := checkScripts(tx); err != nil {
		        // Arek: here we are wrapping - why?
			return validator.NewError(err, api.ErrStatusUnlockingScripts)
		}
	}
	(...)
}

func standardCheckFees(tx *bt.Tx, feeQuote *bt.FeeQuote) error {
	feesOK, expFeesPaid, actualFeePaid, err := isFeePaidEnough(feeQuote, tx)
	if err != nil {
		return validator.NewError(err, api.ErrStatusFees)
	}

	if !feesOK {
		err = fmt.Errorf("transaction fee of %d sat is too low - minimum expected fee is %d sat", actualFeePaid, expFeesPaid)
		return validator.NewError(err, api.ErrStatusFees)
	}

	return nil
}

func checkScripts(tx *bt.Tx) error {
	for i, in := range tx.Inputs {
		prevOutput := &bt.Output{
			Satoshis:      in.PreviousTxSatoshis,
			LockingScript: in.PreviousTxScript,
		}

		if err := interpreter.NewEngine().Execute(
			interpreter.WithTx(tx, i, prevOutput),
			interpreter.WithForkID(),
			interpreter.WithAfterGenesis(),
		); err != nil {
			return fmt.Errorf("script execution failed: %w", err)
		}
	}

	return nil
}
```
Note that `standardCheckFees` and `checkScripts` both return error, but one of them needs wrapping (in the `ValidateTransaction` function) while the other does not. In my opinion, this leaks abstraction, forcing the developer to read and understand the body of the function to decide if he should wrap the returned error. With my change, the developer will know from the function definition alone: if it returns error, wrap it! If it returns validator.Error, return it as it is.

Another side effect is that in complex functions like `CommonValidation`, where we return several different errors from various branches, changing the definition to return `*Error` ensures that we return this type every time. With the previous implementation, there could be a situation where new code returns a standard error, which would be inconsistent and could lead to incorrect system behaviors.

internal/validator/default/testdata/fixture.go
arkadiuszos4chain: Changed

internal/woc_client/woc_client.go
arkadiuszos4chain: Fixed

internal/validator/default/default_validator.go
arkadiuszos4chain: fixed

internal/validator/default/default_validator.go
arkadiuszos4chain: New two tests have been added.

boecklim (approved):

kuba-4chain (approved):

PR #506, opened 2024-07-16
feat!(tx_statuses): simplify transactions statuses and their update logic in metamorph database

kuba-4chain:
## What's been done
1. Statuses are simplified. Check README.md to see the new statuses and what/how were they changed.
2. A migration was written for changing the old statuses to new statuses in metamorph database.
3. The logic for statuses update in metamorph database was greatly simplified.
4. The readme and docs (statuses diagram) were udpated.
5. DOUBLE_SPEND_ATTEMPTED status was introduced. More detailed description of the double spend detection will be done in the PR with the feature itself.

## Breaking Changes (kinda). Run metamorph DB migrations.
This PR introduces breaking changes and will require to run a new migration in the metamorph database. After running that migration, everything should run smoothly.

## Why?
1. Previously, the status for `SEEN_IN_ORPHAN_MEMPOOL` had a higher code than `SEEN_ON_NETWORK`. Therefore, in the database a logic for status updates had to be (pseudocode):
```
CHANGE transaction.status
WHEN transactions.status < new_status OR (transaction.status == SEEN_IN_ORPHAN_MEMPOOL AND new_status == SEEN_ON_NETWORK)
```
which now is simplified to:
```
CHANGE transaction.status
WHEN transactions.status < new_status
```

2. There was no room to create any more statuses in between, such as a `DOUBLE_SPEND_ATTEMPTED`. Now, there is a difference (of 10) between the codes, which allows for future introduction of new statuses in between the existing ones without creating more breaking changes.


arkadiuszos4chain:
I know you didn't touch these lines, but could you change `arc/internal/metamorph/store/postgresql/postgres.go:504` to 
```
txHashes[i] = update.Hash[:]
```
please?

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -520,10 +518,9 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
 						UNNEST($1::BYTEA[], $2::INT[], $3::TEXT[])
 				AS t(hash, status, reject_reason)
 			) AS bulk_query
-			WHERE
-			metamorph.transactions.hash=bulk_query.hash
-				AND ( metamorph.transactions.status < bulk_query.status OR ( metamorph.transactions.status = $5 AND bulk_query.status = $4 ))
-			    AND metamorph.transactions.status != $6
+			WHERE metamorph.transactions.hash=bulk_query.hash
+				AND metamorph.transactions.status < bulk_query.status
+			  AND metamorph.transactions.status != $4
```
arkadiuszos4chain, line 523: Why do we need `AND metamorph.transactions.status != MINED` ? If we don't want to update these records maybe it's better to remove them from the arguments?

README.md
```diff
@@ -163,26 +163,26 @@ go run main.go -metamorph=true
...
+| 70  | `ACCEPTED_BY_NETWORK`  | The transaction has been accepted by a connected Bitcoin node on the ZMQ interface. If metamorph is not connected to ZMQ, this status will never by set.                                                 |
+| 80  | `SEEN_IN_ORPHAN_MEMPOOL`             | The transaction has been sent to at least 1 Bitcoin node but parent transaction was not found. |
+| 90  | `DOUBLE_SPEND_ATTEMPTED`             | The transaction is a double spend attempt. Competing transaction(s) will be returned with this status. |
+| 100   | `SEEN_ON_NETWORK`      | The transaction has been seen on the Bitcoin network and propagated to other nodes. This status is set when metamorph receives an INV message for the transaction from another node than it was sent to. |
+| 110  | `MINED`                | The transaction has been mined into a block by a mining node.                                                                                                                                            |
```
arkadiuszos4chain, line 182: Maybe it would be a good idea to switch the `MINED` and `REJECTED` statuses.

If the `MINED` status were the final one (with the highest value), we could remove the `AND metamorph.transactions.status != MINED` condition from the `WHERE` clauses in update queries. In my opinion, these conditions are not intuitive and it is easy to forget about them if someone new writes a query.

kuba-4chain: You have a point. `MINED` should be more important than `REJECTED`. If one node sends us `MINED` but another node rejects our transaction for some reason, we don't want to update that status and keep it on `MINED`.

I applied this switch and simplified the logic as you suggested. Thanks.

@boecklim take a look at this and verify :)

arkadiuszos4chain (approved):

arkadiuszos4chain (approved):

internal/broadcaster/api.go
boecklim, line 1: Maybe we should change this later.
If you leave this broadcaster using `XWaitForStatus` then the e2e test should actually keep working.

boecklim (approved):

PR #509, opened 2024-07-22
feat: move blocktx_api and metamorph_api proto definitions and models from /pkg/ to /internal/

kuba-4chain:


boecklim (approved):
Please create a follow up PR where the e2e tests are updated to not use the now unavailable package `"github.com/bitcoin-sv/arc/pkg/metamorph/metamorph_api"` but comparing strings instead

PR #517, opened 2024-07-25
ARCO-144: cumulative fees validation

arkadiuszos4chain:
# Cumulative Fee Validation

##  Description

The "Cumulative Fee Validation" feature is designed to check if the chain of unmined transactions (submitted transaction and its unmined ancestors) has paid a sufficient amount of fees. This validation is carried out based on a specific HTTP header.

## Usage

To use the "Cumulative Fee Validation" feature, you need to send the `X-CumulativeFeeValidation` header with the value set to `true`. 

Example usage:
```
X-CumulativeFeeValidation: true
```

### Special Cases

If the `X-SkipFeeValidation` header is also sent, the fee validation will be skipped even if `X-CumulativeFeeValidation` is set to `true`.

Example usage:
```
X-CumulativeFeeValidation: true
X-SkipFeeValidation: true
```

In this case, the fee validation will not be performed.

### Example 1: Insufficient Fee Paid by One Ancestor
Transaction t0 has two unmined ancestors t1 and t2.

* t0 has paid a sufficient fee for itself.
* t1 has not paid a sufficient fee.
* t2 has paid a sufficient fee for itself.

#### Validation Result:
The validation will fail because t1 has not paid a sufficient fee and no other transaction cover it.

### Example 2: All Transactions Paid Their Own Fees
Transaction t0 has two unmined ancestors t1 and t2.

* t0 has paid a sufficient fee.
* t1 has paid a sufficient fee.
* t2 has paid a sufficient fee.

#### Validation Result:
The validation will pass because both t1 and t2 have paid sufficient fees.

### Example 3: Ancestors Did Not Pay, But Transaction Covers All Fees
Transaction t0 has two unmined ancestors t1 and t2.

* t1 has not paid a sufficient fee.
* t2 has not paid a sufficient fee.
* t0 covers the fees for itself and both t1 and t2.

#### Validation Result:
The validation will pass because t0 covers the cumulative fees for the entire chain, including t1 and t2.

The system performs the fee validation and returns a result indicating that the chain of transactions has sufficient fees.

## Notes

- The `X-CumulativeFeeValidation` header must be set to `true` for the validation to be performed.
- The `X-SkipFeeValidation` header takes precedence over `X-CumulativeFeeValidation` and causes the fee validation to be skipped.

This feature is crucial to ensure that the chain of unmined transactions has sufficient fees, which is essential for the effective management of the transaction network.


internal/metamorph/metamorph_api/metamorph_api.proto
```diff
@@ -31,6 +31,7 @@ service MetaMorphAPI {
   rpc PutTransaction (TransactionRequest) returns (TransactionStatus) {}
   rpc PutTransactions (TransactionRequests) returns (TransactionStatuses) {}
   rpc GetTransaction (TransactionStatusRequest) returns (Transaction) {}
+  rpc GetManyTransactions (ManyTransactionsStatusRequest) returns (Transactions) {}
```
boecklim, line 34: How about the name `GetTransactions`? IMO that would sound better

internal/metamorph/server.go
```diff
@@ -392,6 +427,21 @@ func (s *Server) getTransactionData(ctx context.Context, req *metamorph_api.Tran
 	return data, announcedAt, minedAt, storedAt, nil
 }
 
+func (s *Server) getTransactionsData(ctx context.Context, req *metamorph_api.ManyTransactionsStatusRequest) ([]*store.StoreData, error) {
```
boecklim, line 430: How about `getTransactions(...)` ?

internal/validator/default/helpers.go
```diff
@@ -58,6 +60,77 @@ func extendTx(ctx context.Context, w validator.TxFinderI, rawTx *bt.Tx) error {
...
+
+		inputs = append(inputs, in)
+		parentInputMap[prevTxID] = inputs
+	}
+
+	// get parents
+	const finderSource = validator.SourceTransactionHandler | validator.SourceWoC
+	parentsTxs, err := w.GetRawTxs(ctx, finderSource, parentsIDs)
+	if err != nil {
+		return nil, fmt.Errorf("cannot extend tx: %w", err)
```
kuba-4chain, line 90: Is this error message correct? Maybe something like "cannot get parent txs for tx: %s"?

internal/validator/default/helpers.go
```diff
@@ -58,6 +60,77 @@ func extendTx(ctx context.Context, w validator.TxFinderI, rawTx *bt.Tx) error {
...
+		if _, found := unmindedAncestorsSet[p.TxID]; found {
+			continue // parent was proccesed already
+		}
+
+		childInputs, found := parentInputMap[p.TxID]
+		if !found {
+			return nil, errParentNotFound
+		}
+
+		// fulfill data about the parent for furhter validation
```
kuba-4chain, line 107: I assume you didn't want to say "furher" 😄 
```suggestion
		// fulfill data about the parent for further validation
```

internal/validator/default/helpers.go
```diff
@@ -58,6 +60,77 @@ func extendTx(ctx context.Context, w validator.TxFinderI, rawTx *bt.Tx) error {
...
+		if err != nil {
+			return nil, err
+		}
+
+		if p.IsMined {
+			continue // we don't need its ancestors
+		}
+
+		bt, _ := bt.NewTxFromBytes(p.Bytes)
+		unmindedAncestorsSet[p.TxID] = bt
```
kuba-4chain, line 118: You could return bt from `extendInputs` to avoid doing it again in line 117
```suggestion
		bt, err := extendInputs(p.Bytes, childInputs)
		if err != nil {
			return nil, err
		}

		if p.IsMined {
			continue // we don't need its ancestors
		}

		unmindedAncestorsSet[p.TxID] = bt
```

internal/validator/helpers.go
```diff
@@ -14,8 +14,20 @@ const (
 	BeefHex
 )
 
+type FindSourceFlag byte
+
+const (
+	SourceTransactionHandler = 1
+	SourceNodes              = 2
+	SourceWoC                = 4
+)
```
kuba-4chain, line 23: I see this is binary. I saw this way of doing it:
```suggestion
const (
	SourceTransactionHandler = 1 << iota // 1 (binary: 001)
	SourceNodes                                          // 2 (binary: 010)
	SourceWoC                                             // 4 (binary: 100)
)
```
Maybe this will be better so someone doesn't change `4` to `3` because that was my initial thought 😄  WDYT?

internal/validator/default/default_validator_test.go
```diff
@@ -333,3 +336,140 @@ func Test_needExtention(t *testing.T) {
...
+						IsMined: false,
+					}
+					return []validation.RawTx{p1, fixture.ParentTx2}, nil
+				}
+
+				if i == 1 {
+					return []validation.RawTx{fixture.AncestorTx1, fixture.AncestorTx2}, nil
+				}
+
+				panic("to many calls")
```
boecklim, line 442: How about `t.Fatal("too many calls")` ?

internal/metamorph/server.go
```diff
@@ -334,6 +334,41 @@ func (s *Server) GetTransaction(ctx context.Context, req *metamorph_api.Transact
 	return txn, nil
 }
 
+func (s *Server) GetManyTransactions(ctx context.Context, req *metamorph_api.ManyTransactionsStatusRequest) (*metamorph_api.Transactions, error) {
```
boecklim, line 337: This function has no test coverage. Could you add a unit test to `server_test.go` for this function?

internal/validator/beef/beef_validator_test.go
```diff
@@ -211,3 +211,100 @@ func getPolicy(satoshisPerKB uint64) *bitcoin.Settings {
 	policy.MinMiningTxFee = float64(satoshisPerKB) / 1e8
 	return policy
 }
+
+func Test_cumulativeCheckFees(t *testing.T) {
```
boecklim, line 215: I think the size of this test could be reduced if it were table based like this

```
func Test_cumulativeCheckFees(t *testing.T) {
	tt := []struct {
		name      string
		minedBeef string
	}{
		{
			name:      "mined parents only - valid fee",
			minedBeef: "0100beef.....",
		},
		{
			...
		},
	}

	for _, tc := range tt {
			hex, _ := hex.DecodeString(tc.minedBeef)
			beefTx, _, _ := beef.DecodeBEEF(hex)

			feeQuote := bt.NewFeeQuote()
			setFees(feeQuote, 1)

			// then
			vErr := cumulativeCheckFees(beefTx, feeQuote)

			// assert
			require.Nil(t, vErr)
		})
	}
```

internal/validator/default/default_validator.go
```diff
@@ -98,6 +102,34 @@ func standardCheckFees(tx *bt.Tx, feeQuote *bt.FeeQuote) *validator.Error {
...
+	cumulativePaidFee := uint64(0)
+
+	for _, tx := range txSet {
+		expFees, err := validator.CalculateMiningFeesRequired(tx.SizeWithTypes(), feeQuote)
+		if err != nil {
+			return validator.NewError(err, api.ErrStatusCumulativeFees)
+		}
+
+		cumulativeExpFee += expFees
+		cumulativePaidFee += (tx.TotalInputSatoshis() - tx.TotalOutputSatoshis())
```
boecklim, line 122: nit: redundant parantheses

internal/validator/default/default_validator_test.go
```diff
@@ -333,3 +336,140 @@ func Test_needExtention(t *testing.T) {
 		})
 	}
 }
+
+func Test_cumulativeCheckFees(t *testing.T) {
```
boecklim, line 340: Also here a table based test would make it smaller and easier to spot the differences between each test case

pkg/api/status.go
```diff
@@ -25,6 +25,7 @@ const (
 	ErrStatusValidatingMerkleRoots  StatusCode = 469
 	ErrStatusFrozenPolicy           StatusCode = 471
 	ErrStatusFrozenConsensus        StatusCode = 472
+	ErrStatusCumulativeFees         StatusCode = 473
```
boecklim, line 28: This needs to be added also here I think: `doc/errors.md`

internal/validator/default/helpers.go
arkadiuszos4chain: You're right we can avoid creating bt.Tx twice! I moved it outside from the `extendInputs`

pkg/api/status.go
```diff
@@ -25,6 +25,7 @@ const (
...
 )
 
 func NewErrorFields(status StatusCode, extraInfo string) *ErrorFields {
```
boecklim, line 31: I think you need to add this error here too e.g.

```
	case ErrStatusCumulativeFees:
		errFields.Detail = "Cumulative fees are too low"
		errFields.Title = "Cumulative fee too low"
		errFields.Type = arcDocServerErrorsUrl + strconv.Itoa(int(ErrStatusCumulativeFees))
```

Otherwise [here](https://github.com/bitcoin-sv/arc/blob/9a353e6079bd7ebd25f4e48e9fd20a24fea48056/pkg/api/handler/default.go#L532) the error will be a generic error

pkg/api/status.go
boecklim: Maybe we need here a comment like: "if a new error is added, don't forget to add to 'doc/errors.md'"

pkg/api/handler/tx_finder.go
```diff
@@ -28,56 +28,71 @@ func (f txFinder) GetRawTxs(ctx context.Context, ids []string) ([]validator.RawT
...
 
 	// try to get remaining txs from the node
-	for _, id := range ids {
-		nTx, err := getTransactionFromNode(f.pc, id)
-		if err != nil {
-			// do we really need this info?
-			f.l.Warn("failed to get transaction from node", slog.String("id", id), slog.String("err", err.Error()))
+	if source.Has(validator.SourceNodes) {
+		for _, id := range ids {
+			nTx, err := getTransactionFromNode(f.pc, id)
```
boecklim, line 65: nit: Move this function `getTransactionFromNode` from `helpers.go` to `tx_finder.go`?

pkg/api/handler/tx_finder.go
```diff
@@ -28,56 +28,71 @@ func (f txFinder) GetRawTxs(ctx context.Context, ids []string) ([]validator.RawT
...
-		} else {
-			remainingIDs = append(remainingIDs, id)
+			if !found {
+				remainingIDs = append(remainingIDs, id)
+			}
 		}
 	}
 
 	ids = remainingIDs[:]
 	remainingIDs = make([]string, 0)
```
boecklim, line 60: Why are we collecting the remaining IDs when we're deleting them here? 🤔

internal/validator/beef/beef_validator_test.go
arkadiuszos4chain: You right, it even looks better now.

internal/validator/default/default_validator_test.go
arkadiuszos4chain: I changed them, but I'm not sure if they are more readable than before. Let me know what you think about the new form

internal/validator/default/helpers.go
arkadiuszos4chain: Fixed

internal/validator/default/helpers.go
arkadiuszos4chain: fixed

internal/validator/helpers.go
arkadiuszos4chain: Good point! Changed

internal/validator/beef/beef_validator_test.go
arkadiuszos4chain: Changed to table based

internal/validator/default/default_validator.go
arkadiuszos4chain: fixed

pkg/api/handler/tx_finder.go
arkadiuszos4chain: Good point

internal/validator/default/default_validator_test.go
arkadiuszos4chain: Changed

pkg/api/handler/tx_finder.go
arkadiuszos4chain: Moved

internal/metamorph/server.go
arkadiuszos4chain: changed

pkg/api/status.go
arkadiuszos4chain: I thought I add this to doc/ in follow up PR

kuba-4chain (approved):

internal/metamorph/metamorph_api/metamorph_api.proto
arkadiuszos4chain: changed

pkg/metamorph/client.go
```diff
@@ -24,6 +24,7 @@ var (
 type TransactionHandler interface {
 	Health(ctx context.Context) error
 	GetTransaction(ctx context.Context, txID string) ([]byte, error)
+	GetManyTransactions(ctx context.Context, txIDs []string) ([]*Transaction, error)
```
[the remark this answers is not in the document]
arkadiuszos4chain: changed

doc/errors.md
```diff
@@ -26,10 +26,19 @@ ErrStatusMalformed: Either the transaction is too small (61 bytes min), there ar
 ErrStatusFees: The fees are too low, sum satoshis out is not less than sum satoshis in.
 
 # 466
-Conflict: Transaction is invalid because the network has already seen a tx which spends the same utxo.
+ErrStatusConflict: Transaction is invalid because the network has already seen a tx which spends the same utxo.
 
-# 481
+# 467
+ErrStatusMinedAncestorsNotFound: invalid BEEF, cannot fined mined ancestor of the transaction.
```
boecklim, line 32: cannot **find** mined ancestor.

boecklim (approved):

internal/metamorph/server.go
arkadiuszos4chain: Added new test

boecklim (approved):

internal/metamorph/server_test.go
```diff
@@ -577,3 +577,85 @@ func TestStartGRPCServer(t *testing.T) {
 		})
 	}
 }
+
+func Test_GetManyTransactions(t *testing.T) {
```
boecklim, line 581: nit: `Test_GetTransactions(...)`

PR #519, opened 2024-07-25
E2E test with JetStream enabled Nats cluster

boecklim:
- NATS JetStream cluster in E2E tests


test/config/config.yaml
```diff
@@ -6,7 +6,7 @@ prometheusEndpoint: /metrics
 prometheusAddr: :2112
 grpcMessageSize: 100000000
 network: regtest
-queueURL: nats://nats:4222
+queueURL: nats://nats-1:4222,nats://nats-2:4223,nats://nats-3:4224
```
kuba-4chain, line 9: The PR is looking great 👍 

I just have one question, why do we need 3 instances of NATS? Is it to make sure no messages are lost in case one NATS instance failing/disconnecting? 

Even if so, it should be extremely rare for that to happen, so maybe it's not worth the resources and time for e2e tests to spin up 3 instances? One should be enough? WDYT?

boecklim: That is right. Just one would be enough for sure.
I just thought it is good to have a cluster here in the e2e test because that also allows to easily lookup how a cluster needs to configured for anybody who wants to run ARC with a JetStream-enabled NATS cluster. So It serves a kind of documentation purposes. What do you think?

arkadiuszos4chain: I think it was a good idea to use a cluster, but @kuba-4chain has a point. Maybe we should limit instances to 2 for easier and faster running tests on local machines.

Maybe we could also apply the same approach to other services?

test/docker-compose.yml
```diff
@@ -101,15 +101,47 @@ services:
...
 
-  nats:
+  nats-1:
     image: nats:2.10.10
-    container_name: 'nats-server'
-    restart: on-failure
-    expose:
-      - "4222"
+    container_name: nats-server-1
+    restart: no
```
arkadiuszos4chain, line 107: Why did you change `restart` option?

test/config/config.yaml
boecklim: Ok, I'll reduce it to 2 instances.

@arkadiuszos4chain Which other services do you mean?

test/docker-compose.yml
boecklim: @arkadiuszos4chain My idea was that the whole e2e test should fail if for any reason of misconfiguration of NATS stops and NATS shouldn't restart in that case. But there is not a strong reason to change it. Could also revert the change.

arkadiuszos4chain: I was just curious. Thanks for the clarification

test/config/config.yaml
arkadiuszos4chain: Nodes and blocktx, I think we are running 3 nodes for tests and scale blocktx to 7 instances

boecklim: @arkadiuszos4chain Sounds good. But let's do it in a separate PR

kuba-4chain (approved):

PR #520, opened 2024-07-29
Feature/jet stream

boecklim:
- Optionally allow to use NATS JetStream client instead of NATS Core client
- Refactor & rename async package and configuration
- Integration test for NATS JetStream client
- Upgrade go version to 1.22.5 => integration test coverage was ignored before


shotasilagadzetaal:
I think we should be merging into staging first from now on? 🤔

internal/message_queue/nats/nats_connection/nats_connection.go
```diff
@@ -0,0 +1,56 @@
...
+			}
+		}),
+		nats.ReconnectHandler(func(_ *nats.Conn) {
+			logger.Info("client reconnected")
+		}),
+		nats.ClosedHandler(func(_ *nats.Conn) {
+			logger.Info("client closed")
+		}),
+		nats.RetryOnFailedConnect(true),
+		nats.PingInterval(2 * time.Minute),
```
shotasilagadzetaal, line 43: Isn't two minutes too long for ping interval?

internal/message_queue/nats/client/nats_jetstream/nats_jetstream_client.go
```diff
@@ -0,0 +1,210 @@
...
+
+	js, err := jetstream.New(nc)
+	if err != nil {
+		return nil, err
+	}
+
+	p.js = js
+
+	for _, topic := range topics {
+		_, err = p.getStream(topic, fmt.Sprintf("%s-stream", topic))
```
shotasilagadzetaal, line 78: What's the point of this call? Just checking if we could encounter error?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/bitcoin-sv/arc
 
-go 1.21.3
+go 1.22.5
```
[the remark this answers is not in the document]
boecklim: Yes, that was a mistake. Will change the description.

internal/message_queue/nats/client/nats_jetstream/nats_jetstream_client.go
boecklim: The `getStream()` function returns the stream, but also ensures it exists i.e. it returns the existing one or creates it if it not exists. For a pure publisher the stream is not needed. It is only needed for a subscriber. Therefore there is no stream returned.

kuba-4chain (approved):

PR #521, opened 2024-07-29
feat: Double Spend Detection - status updates

kuba-4chain:
### The current flow and scope of Double Spend Detection feature:
1. Nodes inform Arc about mempool conflict - there are conflicting transactions - through ZMQ.
2. For each hash from the conflicting transactions a message to the statusChannel is sent with all competing txs excluding self.
3. Status is being updated in batches in `processor.go` in `internal/metamorph`
    - Batches exclude duplicates, but merge unique competing txs ( e.g. [tx1, tx2] + [tx1, tx3] -> [tx1, tx2, tx3] )
4. Statuses are then updated in postgresql store through a separate metamorph store method `UpdateDoubleSpend`
    - the current existing competing txs are firstly fetched from db and merged with incoming competing txs to save only unique values and only update records which have any extra data

### Out of this PR's scope - to be done in the next PR
1. Competing transactions returned together with the status `DOUBLE_SPEND_ATTEMPTED` to the user
2. When one of competing txs is `MINED` the other become `REJECTED`
3. Callbacks for each tx with `DOUBLE_SPEND_ATTEMPTED` status with competing txs


test/double_spend_test.go
```diff
@@ -44,12 +44,14 @@ func TestDoubleSpend(t *testing.T) {
 
 			// send double spending transaction when first tx is in mempool
 			txMempool := createTxToNewAddress(t, privateKey, utxos[0])
-			postTxChecksStatus(t, arcClient, txMempool, Status_REJECTED, tc.extFormat)
+			postTxChecksStatus(t, arcClient, txMempool, Status_DOUBLE_SPEND_ATTEMPTED, tc.extFormat)
```
boecklim, line 47: What would be really nice here is if it could be checked whether the double spend attempting txs can be found in the response

internal/metamorph/processor.go
```diff
@@ -305,34 +289,73 @@ func (p *Processor) StartProcessStatusUpdatesInStorage() {
...
+func (p *Processor) checkAndUpdate(statusUpdatesMap map[chainhash.Hash]store.UpdateStatus) {
+	if len(statusUpdatesMap) == 0 {
+		return
+	}
+
+	statusUpdates := make([]store.UpdateStatus, 0, p.processStatusUpdatesBatchSize)
+	doubleSpendUpdates := make([]store.UpdateStatus, 0)
+
+	for _, status := range statusUpdatesMap {
+		if len(status.CompetingTxs) > 0 {
```
boecklim, line 318: Wouldn't this be a bit more explicit?

```
status.Status == metamorph_api.Status_DOUBLE_SPEND_ATTEMPTED
```

internal/metamorph/processor_helpers.go
```diff
@@ -1,7 +1,78 @@
...
 
+func (p *Processor) GetProcessorMapSize() int {
 	return p.ProcessorResponseMap.Len()
+}
+
+func updateStatusMap(statusUpdatesMap map[chainhash.Hash]store.UpdateStatus, statusUpdate store.UpdateStatus) {
+	foundStatusUpdate, found := statusUpdatesMap[statusUpdate.Hash]
+
+	if !found || (found && shouldUpdateStatus(statusUpdate, foundStatusUpdate)) {
+		if len(statusUpdate.CompetingTxs) > 0 {
```
boecklim, line 16: Wouldn't this be a bit more explicit?

status.Status == metamorph_api.Status_DOUBLE_SPEND_ATTEMPTED

internal/metamorph/processor_helpers.go
```diff
@@ -1,7 +1,78 @@
...
+	if !unorderedEqual(statusUpdate.CompetingTxs, foundStatus.CompetingTxs) {
+		return true
+	}
+
+	return false
+}
+
+// unorderedEqual checks if two string slices contain
+// the same elements, regardless of order
+func unorderedEqual(sliceOne, sliceTwo []string) bool {
```
boecklim, line 40: This looks correct. Alternatively maybe also https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff could be used

See https://stackoverflow.com/a/67624073

Or

```
deep.Equal([]string{"hello", "world"}, []string{"world", "hello"}, deep.FLAG_IGNORE_SLICE_ORDER) 
```

https://github.com/go-test/deep/issues/28#issuecomment-754269711

internal/metamorph/processor_helpers.go
```diff
@@ -1,7 +1,78 @@
...
+)
 
+func (p *Processor) GetProcessorMapSize() int {
 	return p.ProcessorResponseMap.Len()
+}
+
+func updateStatusMap(statusUpdatesMap map[chainhash.Hash]store.UpdateStatus, statusUpdate store.UpdateStatus) {
+	foundStatusUpdate, found := statusUpdatesMap[statusUpdate.Hash]
+
+	if !found || (found && shouldUpdateStatus(statusUpdate, foundStatusUpdate)) {
```
boecklim, line 15: The second `found` is not needed as it is always true
```
if !found || (shouldUpdateStatus(statusUpdate, foundStatusUpdate))
```

boecklim (approved):

internal/metamorph/processor.go
kuba-4chain: That's correct, because for now only `DOUBLE_SPEND_ATTEMPTED` transactions have CompetingTxs. 

But, I can imagine, in the future, transactions having a different status (e.g. `REJECTED`) with CompetingTxs. Then, they should also be updated in the same way that double spend transactions are updated. That's why I decided on that check instead of `Status == metamorph_api.Status_DOUBLE_SPEND_ATTEMPTED`

Let me know if my reasoning makes sense. If you this use case I described will never occur, I'll change it to your suggestion.

internal/metamorph/processor_helpers.go
kuba-4chain: Here I followed the same reasoning as I explained above. Let me know what you think

internal/metamorph/processor_helpers.go
kuba-4chain: The second option would require installing a new dependency. I'm not sure if this simple function is worth it?

As for the first option, I run some small benchmarks, which suggest that this custom implementation that I did is much faster than using a `go-cmp` with custom comparator. Here are the results, I can also include them in the code if you'd like:

```
goos: darwin
goarch: arm64
pkg: github.com/bitcoin-sv/arc/internal/metamorph
BenchmarkUnorderedEqual/txIDs-1-10              71081755                16.63 ns/op
BenchmarkUnorderedEqual/txIDs-10-10              3006540               400.3 ns/op
BenchmarkUnorderedEqual/txIDs-100-10              171099              6963 ns/op
BenchmarkUnorderedEqual/txIDs-1000-10              14553             82411 ns/op
BenchmarkUnorderedEqual/txIDs-10000-10              1507            789109 ns/op
BenchmarkCmpDiff/txIDs-1-10                       120276              9846 ns/op
BenchmarkCmpDiff/txIDs-10-10                       26148             45883 ns/op
BenchmarkCmpDiff/txIDs-100-10                       1446            820165 ns/op
BenchmarkCmpDiff/txIDs-1000-10                       177           6744339 ns/op
BenchmarkCmpDiff/txIDs-10000-10                       18          62019458 ns/op
```

Here's the function I used ad CmpDiff:
```
func cmpDiff(sliceOne, sliceTwo []string) bool {
	comparator := func(a, b string) bool { return a < b }
	return cmp.Diff(sliceOne, sliceTwo, cmpopts.SortSlices(comparator)) == ""
}
```

internal/metamorph/processor_helpers.go
kuba-4chain: You're right, fixed

internal/metamorph/processor_helpers.go
boecklim: Hey, nice. Then we for sure stick with the faster implementation of yours 👍

internal/metamorph/processor.go
boecklim: Yes, I think you're right. Because the only reason you want to keep updating them is not the status but the competing transactions 👍

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -573,6 +591,135 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+		return nil, err
+	}
+	defer rows.Close()
+
+	res, err := p.getStoreDataFromRows(rows)
+	if err != nil {
+		return res, err
+	}
+
+	err = tx.Commit()
```
arkadiuszos4chain, line 715: Shouldn't we commit the transaction before we map the rows?

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -573,6 +591,135 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+		err := rows.Scan(
+			&hash,
+			&competingTxs,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(hash) > 0 {
+			data.Hash, err = chainhash.NewHash(hash)
```
arkadiuszos4chain, line 664: Maybe it would be overengineering, premature microoptimization, or even an incorrect solution, but...

BUT maybe we don't need to use `chainhash`es here and then compare them later. We already have pure bytes of hashes (from line 630),  so maybe it would be simpler to compare them directly?

internal/metamorph/zmq.go
```diff
@@ -181,6 +155,53 @@ func (z *ZMQ) Start(zmqi ZMQI) error {
...
+		return
+	}
+
+	if txInfo.IsMempoolConflictDetected || txInfo.IsDoubleSpendDetected || len(txInfo.CollidedWith) > 0 {
+		// In case of Double Spend Attempted, we don't want to add an error to the response, as it will immediately
+		// return this response to the user. We want to inform him about the status update by callback and wait,
+		// because the node can accept one of the transactions from double spend and reject the other(s). Then,
+		// we will update the other transactions with the REJECTED status and an error.
+		status = metamorph_api.Status_DOUBLE_SPEND_ATTEMPTED
+		for _, ctx := range txInfo.CollidedWith {
```
arkadiuszos4chain, line 184: Please rename `ctx` to `cTx`/`tx` or something since ctx usually means "context"

arkadiuszos4chain (changes requested):

internal/metamorph/processor_test.go
```diff
@@ -500,9 +518,10 @@ func TestStartSendStatusForTransaction(t *testing.T) {
 					t.Fatal("expected callbacks never sent")
 				}
 			}
-			time.Sleep(time.Millisecond * 100)
+			time.Sleep(time.Millisecond * 300)
```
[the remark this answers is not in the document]
kuba-4chain: Interestingly enough, yes. I have no idea why, but with 100 milliseconds all the functions below return 0. With 300 milliseconds they return correct values. No idea 🤷‍♂️

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -503,7 +516,11 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
 	for i, update := range updates {
 		txHashes[i] = update.Hash.CloneBytes()
 		statuses[i] = update.Status
-		rejectReasons[i] = update.RejectReason
+
+		rejectReasons[i] = ""
```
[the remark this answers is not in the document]
kuba-4chain: Removed

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -573,6 +591,135 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+		,metamorph.transactions.raw_tx
+		,metamorph.transactions.locked_by
+		,metamorph.transactions.merkle_path
+		,metamorph.transactions.retries
+		;
+    `
+
+	txHashes := make([][]byte, len(updates))
+	for i, update := range updates {
+		txHashes[i] = update.Hash.CloneBytes()
```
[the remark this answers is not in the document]
kuba-4chain: Right, done

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -573,6 +591,135 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+
+		dbData = append(dbData, data)
+	}
+
+	statuses := make([]metamorph_api.Status, len(updates))
+	competingTxs := make([]string, len(updates))
+	rejectReasons := make([]string, len(updates))
+
+	for i, update := range updates {
+		txHashes[i] = update.Hash.CloneBytes()
```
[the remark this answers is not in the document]
kuba-4chain: I wanted to make it explicit, but you're right. Removed

internal/metamorph/store/postgresql/postgres.go
kuba-4chain: That's a good question and I was thinking about it a lot. 

I decided to rollback the transaction in case of error when mapping rows for 2 reasons:
1. It's consistent with other methods in this file.
2. In case something malformed is added to db and it throws an error when trying to map it from rows, we probably don't want that malformed record in the db anyways, because it will always return an error when retrieving it.

I implemented the rollback, but if you think we should do it another way, let me know

internal/metamorph/zmq.go
kuba-4chain: Right, done

internal/metamorph/store/postgresql/postgres.go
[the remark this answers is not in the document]
kuba-4chain: This doesn't work. Fails unit tests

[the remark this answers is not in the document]
arkadiuszos4chain: Hmm I'll take a look why

internal/metamorph/store/postgresql/postgres.go
kuba-4chain: I think it's a brilliant idea, I implemented it in the next PR: https://github.com/bitcoin-sv/arc/pull/527

It was easier for me that way, because I already had a separate function for this. Please check it there :)

arkadiuszos4chain (approved):

PR #527, opened 2024-07-31
feat: Reject competing txs when one is mined

kuba-4chain:
### What has been done
1. When metamorph receives information about mined transactions from blocktx and if any `MINED` transaction has competing txs, each competing tx (if exists in metamorph db) is given `REJECTED` status with the `RejectReason = "double spend attempted"`.
2. Optimisation of comparing hashes, we do it with `[]byte` arrays now instead of `*chainhash.Hash`.
3. Small refactor of methods in metamorph postgresql store.
4. Integration test of all Double Spend Detection feature


boecklim (approved):

internal/metamorph/store/postgresql/postgres_helpers.go
```diff
@@ -19,3 +35,194 @@ func mergeUnique(arr1, arr2 []string) []string {
...
+			data.CallbackUrl = callbackUrl.String
+		}
+
+		if callbackToken.Valid {
+			data.CallbackToken = callbackToken.String
+		}
+
+		if rejectReason.Valid {
+			data.RejectReason = rejectReason.String
+		}
```
arkadiuszos4chain, line 122: I think there is no reason to check if the field is `valid` here

kuba-4chain: I'll remove that in the next PR

PR #528, opened 2024-07-31
Fix/rollback err

boecklim:
- Rollback if getStoreDataFromRows() fails
- Do not overwrite rollback error but join errors

### DIFF
--- a/internal/metamorph/store/postgresql/postgres.go
+++ b/internal/metamorph/store/postgresql/postgres.go
@@ -434,8 +434,8 @@ func (p *PostgreSQL) GetSeenOnNetwork(ctx context.Context, since time.Time, unti
 	LIMIT $4 OFFSET $5
 	FOR UPDATE`, metamorph_api.Status_SEEN_ON_NETWORK, since, untilTime, limit, offset, p.hostname)
 	if err != nil {
-		if err := tx.Rollback(); err != nil {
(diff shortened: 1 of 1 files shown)


internal/metamorph/store/postgresql/postgres.go
```diff
@@ -545,24 +545,24 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
 		}
 		return nil, err
 	}
 	defer rows.Close()
 
 	res, err := p.getStoreDataFromRows(rows)
-	if err != nil {
-		return res, err
+	if rollBackErr := tx.Rollback(); rollBackErr != nil {
+		return nil, errors.Join(err, fmt.Errorf("failed to rollback: %v", rollBackErr))
```
kuba-4chain, line 565: You need to check the `err` first and then rollback

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -631,24 +631,24 @@ func (p *PostgreSQL) UpdateMined(ctx context.Context, txsBlocks []*blocktx_api.T
...
```
kuba-4chain, line 651: Also here, check if there's any error

kuba-4chain (approved):

PR #529, opened 2024-08-01
disable HandleTransactionAnnouncement

shotasilagadzetaal:
We have memory issue caused by processing too many messages in HandleTransactionAnnouncement. On every message we create new go routine which pushes message into the channel. When checking the leak cause I see too many go rougines blocked on the channel push. Having many go rougines created caused memory issue.

We don't actually need to process every transaction with this message. We will get SEEN_ON_NETWORK status for our transactions from HandleTransaction handler instead.


boecklim (changes requested):
We need the sensitivity of the INV messages coming in from the network to see which txs were seen on network. 
Also I don't see how this could be the cause of the memory leak where we have these slow increase of memory usage.

arkadiuszos4chain (commented):
What is the motivation for the change?
What would be the impact of it?
What are the benefits?

internal/metamorph/peer_handler.go
```diff
@@ -45,12 +45,6 @@ func (m *PeerHandler) HandleTransactionSent(msg *wire.MsgTx, peer p2p.PeerI) err
 
 // HandleTransactionAnnouncement is a message sent to the PeerHandler when a transaction INV message is received from a peer.
 func (m *PeerHandler) HandleTransactionAnnouncement(msg *wire.InvVect, peer p2p.PeerI) error {
-	m.messageCh <- &PeerTxMessage{
```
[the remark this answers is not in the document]
arkadiuszos4chain: Please DO NOT resolve the conversation I started without providing an answer. It's unprofessional.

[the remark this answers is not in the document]
shotasilagadzetaal: I reverted this change and that's why it was resolved. 

Then we had discussion again and decided to remove it again. The load of messages received from this handler was too much to handle in database. That caused creation of too many go routines waiting on this channel to push new message. The number of new go routines caused memory issue eventually. 

Now we will get seen_on_network status from HandleTransaction function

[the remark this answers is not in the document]
boecklim: @shotasilagadzetaal 
I think this PR is not ready for review yet.

We're going to experiment with this branch.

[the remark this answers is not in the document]
shotasilagadzetaal: I will simply fix the test cases and then we can decide.

test/endpoint_test.go
```diff
@@ -242,7 +242,7 @@ func TestPostCallbackToken(t *testing.T) {
 
 			require.Equal(t, http.StatusOK, response.StatusCode())
 			require.NotNil(t, response.JSON200)
-			require.Equal(t, Status_SEEN_ON_NETWORK, response.JSON200.TxStatus)
+			require.Equal(t, Status_ACCEPTED_BY_NETWORK, response.JSON200.TxStatus)
```
boecklim, line 245: Hey @shotasilagadzetaal 
I don't think the solution should be to expect now another status everywhere. IMO it should work the same.

Here is an alternative approach how this can be achieved by sending the request 1 or 2 seconds later: https://github.com/bitcoin-sv/arc/commit/ae2d53de08991301d2e061201c881b3138b11d26

shotasilagadzetaal: Reverted

internal/metamorph/peer_handler.go
```diff
@@ -45,15 +45,6 @@ func (m *PeerHandler) HandleTransactionSent(msg *wire.MsgTx, peer p2p.PeerI) err
 
 // HandleTransactionAnnouncement is a message sent to the PeerHandler when a transaction INV message is received from a peer.
 func (m *PeerHandler) HandleTransactionAnnouncement(msg *wire.InvVect, peer p2p.PeerI) error {
-	select {
```
arkadiuszos4chain, line 48: Please add comment why we do not handle INV messages

internal/metamorph/processor.go
```diff
@@ -704,6 +745,15 @@ func (p *Processor) ProcessTransaction(ctx context.Context, req *ProcessorReques
 	p.responseProcessor.Add(statusResponse)
 }
 
+func (p *Processor) RequestTransaction(txHash *chainhash.Hash) {
```
arkadiuszos4chain, line 748:
```suggestion
func (p *Processor) DeferRequestTransaction(txHash *chainhash.Hash) {
```

internal/metamorph/processor.go
```diff
@@ -704,6 +745,15 @@ func (p *Processor) ProcessTransaction(ctx context.Context, req *ProcessorReques
...
+	p.announcedTransactionsLock.Lock()
+	p.announcedTransactions = append(p.announcedTransactions, AnnouncedTransaction{
+		second: uint64(time.Now().Unix()),
```
arkadiuszos4chain, line 751:
```suggestion
		second: uint64(p.now().Unix()),
```

internal/metamorph/processor.go
```diff
@@ -86,6 +86,14 @@ type Processor struct {
 
 	processMinedInterval  time.Duration
 	processMinedBatchSize int
+
+	announcedTransactions     []AnnouncedTransaction
+	announcedTransactionsLock sync.Mutex
+}
+
+type AnnouncedTransaction struct {
+	second uint64
```
arkadiuszos4chain, line 95:
```suggestion
	timestamp uint64
```

internal/metamorph/processor.go
```diff
@@ -86,6 +86,14 @@ type Processor struct {
 
 	processMinedInterval  time.Duration
 	processMinedBatchSize int
+
+	announcedTransactions     []AnnouncedTransaction
```
arkadiuszos4chain, line 90: Please add a comment that the collection is ordered, as it will make it much easier to understand how the code below works.
Or change name to smt like "orderedAnnouncedTxsCollection"

internal/metamorph/processor.go
```diff
@@ -384,6 +394,37 @@ func (p *Processor) StartSendStatusUpdate() {
...
+		defer p.waitGroup.Done()
+		for {
+			select {
+			case <-p.ctx.Done():
+				return
+
+			case <-ticker.C:
+				p.announcedTransactionsLock.Lock()
+				for k := 0; k < len(p.announcedTransactions); k++ {
+					if p.announcedTransactions[k].second < uint64(time.Now().Unix())-1 {
```
arkadiuszos4chain, line 410: 1. use `p.now()`
2. move calculating timestamp (`uint64(time.Now().Unix())-1`) above the loop
3. remove redundant casting from `p.pm.RequestTransaction((*chainhash.Hash)(p.announcedTransactions[k].hash))`

internal/metamorph/processor.go
```diff
@@ -384,6 +394,37 @@ func (p *Processor) StartSendStatusUpdate() {
...
+	go func() {
+		defer p.waitGroup.Done()
+		for {
+			select {
+			case <-p.ctx.Done():
+				return
+
+			case <-ticker.C:
+				p.announcedTransactionsLock.Lock()
+				for k := 0; k < len(p.announcedTransactions); k++ {
```
arkadiuszos4chain, line 409: if you move initialization of index counter above the loop you could simply the code like:
```
var k int
for k=0;k <len(p.announcedTransactions); k++{
    if p.announcedTransactions[k].second < uint64(time.Now().Unix())-1 {
     // request tx (...)
    }else{
       break
    }
}

p.announcedTransactions = p.announcedTransactions[k:]
```

internal/metamorph/processor.go
```diff
@@ -384,6 +394,37 @@ func (p *Processor) StartSendStatusUpdate() {
...
+						if k == len(p.announcedTransactions)-1 {
+							p.announcedTransactions = []AnnouncedTransaction{}
+							break
+						}
+					} else {
+						p.announcedTransactions = p.announcedTransactions[k:]
+						break
+					}
+				}
+				p.announcedTransactionsLock.Unlock()
```
arkadiuszos4chain, line 422: Reset the ticker at the end

internal/metamorph/processor.go
```diff
@@ -689,6 +727,9 @@ func (p *Processor) ProcessTransaction(ctx context.Context, req *ProcessorReques
 		return
 	}
 
+	// will be requesting transaction after ~2 seconds to get SEEN_ON_NETWORK status
```
arkadiuszos4chain, line 730: This operation may be blocking. Can we move it to the end of the processing? Does it make sense?

internal/metamorph/processor.go
```diff
@@ -384,6 +394,37 @@ func (p *Processor) StartSendStatusUpdate() {
 	}()
 }
 
+func (p *Processor) StartCheckingTransactionsInNetwork() {
+	p.waitGroup.Add(1)
+	ticker := time.NewTicker(100 * time.Millisecond)
```
arkadiuszos4chain, line 399: Open question: Is it necessary to lock the `announcedTransactions` collection every 100 milliseconds? A better approach might be to increase the interval to 0.5-1 second and decrease the delta between the current time and the transaction announcement time. WDYT?

arkadiuszos4chain (changes requested):

internal/metamorph/processor.go
shotasilagadzetaal: It's not actually timestamp, it's simply number of seconds.

internal/metamorph/processor.go
shotasilagadzetaal: Don't understand second point here.

internal/metamorph/processor.go
shotasilagadzetaal: What do you mean blocking? It's simple mutex lock nothing complicated here.

internal/metamorph/processor.go
shotasilagadzetaal: Honestly not sure what we need here exactly. Or in other words what's the best time interval to wait before requesting transaction.

internal/metamorph/processor.go
arkadiuszos4chain: Yes, but we lock the same collection in `StartCheckingTransactionsInNetwork`

shotasilagadzetaal: Sorry still confused. Yes we lock it in `StartCheckingTransactionsInNetwork` too but why is it a problem?

internal/metamorph/processor.go
arkadiuszos4chain:
```
for k := 0; k < len(p.announcedTransactions); k++ {
	if p.announcedTransactions[k].second < uint64(time.Now().Unix())-1{
	      (...)
	}
	(...)
}
					
```
To:
```
until :=  uint64(p.Now().Unix())-1
for k := 0; k < len(p.announcedTransactions); k++ {
	if p.announcedTransactions[k].second < until{
	    (...)
	}
	(...)
}

internal/metamorph/processor.go
arkadiuszos4chain: It's the number of seconds since the Unix epoch. It is used to determine if an object is older than a specific moment in time. Seems like a timestamp to me

boecklim:
Closed because the fix for issue of too much DB load will be resolved later

PR #534, opened 2024-08-06
ARCO-163: update go version and dependencies

pawellewandowski98:
# Description:
- Updated GO version to 1.22.5
- Updated all dependencies 
- Updated tools in `Makefile`
- Replaced `oapi-codegen` version from v1.11.0 to v2.3.0
- Regenerated `arc.go` file using `Makefile`


kuba-4chain (approved):

internal/metamorph/store/postgresql/postgres_test.go
```diff
@@ -52,7 +52,7 @@ func TestMain(m *testing.M) {
 		log.Fatalf("failed to create pool: %v", err)
 	}
 
-	port := "5433"
+	port := "5434"
```
arkadiuszos4chain, line 55: Why?

boecklim: @arkadiuszos4chain There is another test which uses the same port. There can be an error because that port is already used

boecklim (approved):

arkadiuszos4chain (approved):

internal/metamorph/store/postgresql/postgres_test.go
pawellewandowski98: Because we had two tests using this same port, I fixed it in the same time as Kuba. His PR is already merged into main so I will revert my changes

PR #537, opened 2024-08-07
feat(ARCO-103): added competingTxs to callback and API responses

kuba-4chain:
These feature have been added in both Callback and API responses:
1. CompetingTxs added as a field
2. If a transaction has status `MINED` and has `CompetingTxs`, we don't return the `CompetingTxs` but instead add a message to the `ExtraInfo` field that this tx was "previously double spend attempted"


internal/metamorph/server_test.go
```diff
@@ -48,85 +48,96 @@ func TestHealth(t *testing.T) {
...
-		server := metamorph.NewServer(s, processor, metamorph.WithMaxTimeoutDefault(100*time.Millisecond))
-
-		var txStatus *metamorph_api.TransactionStatus
-		txRequest := &metamorph_api.TransactionRequest{
-			RawTx: testdata.TX1Raw.Bytes(),
-		}
-
-		processor.ProcessTransactionFunc = func(req *metamorph.ProcessorRequest) {
-			time.Sleep(10 * time.Millisecond)
+	testCases := []struct {
```
boecklim, line 51: Great. I prefer table based too 😸

boecklim (approved):

pkg/metamorph/client.go
```diff
@@ -132,13 +133,14 @@ func (m *Metamorph) GetTransactionStatus(ctx context.Context, txID string) (stat
...
-		ExtraInfo:   tx.GetRejectReason(),
-		Timestamp:   m.now().Unix(),
+		TxID:         txID,
+		MerklePath:   tx.GetMerklePath(),
+		Status:       tx.GetStatus().String(),
+		BlockHash:    tx.GetBlockHash(),
+		BlockHeight:  tx.GetBlockHeight(),
+		ExtraInfo:    tx.GetRejectReason(),
+		CompetingTxs: tx.GetCompetingTxs(),
+		Timestamp:    m.now().Unix(),
```
arkadiuszos4chain, line 143: Why do we use `now().Unix()` here?

arkadiuszos4chain (approved):

PR #539, opened 2024-08-08
ARCO-149: change callback column to allow saving multiple parameters

pawellewandowski98:
# Description
- changed transaction db, two columns (callback_url and callback_token) were combined into callbacks (type jsonb) which allows for saving multiple callbacks for one tx
- Adjusted queries to new structure


internal/metamorph/callbacker.go
```diff
@@ -124,9 +124,14 @@ func (p *Callbacker) SendCallback(logger *slog.Logger, tx *store.StoreData) {
 		status.ExtraInfo = &tx.RejectReason
 	}
 
-	for i := 0; i < CallbackTries; i++ {
+	for _, callback := range tx.Callbacks {
+		p.sendCallback(logger, tx, callback, status, sleepDuration)
```
kuba-4chain, line 128: Maybe this could be run in a goroutine, so we send all callbacks at the same time instead of one-by-one

internal/metamorph/server.go
```diff
@@ -174,11 +174,14 @@ func (s *Server) PutTransaction(ctx context.Context, req *metamorph_api.Transact
...
+	callback := store.StoreCallback{
+		CallbackURL:   req.GetCallbackUrl(),
+		CallbackToken: req.GetCallbackToken(),
+	}
 	sReq := &store.StoreData{
 		Hash:              hash,
 		Status:            statusReceived,
-		CallbackUrl:       req.GetCallbackUrl(),
-		CallbackToken:     req.GetCallbackToken(),
+		Callbacks:         []store.StoreCallback{callback},
```
kuba-4chain, line 184: It's a nit-pick, just an idea, but maybe it would be better to create it here directly, without creating a separate struct above. WDYT?
```suggestion
		Callbacks: []store.StoreCallback{{
			CallbackURL:   req.GetCallbackUrl(),
			CallbackToken: req.GetCallbackToken(),
		}},
```

internal/metamorph/server.go
```diff
@@ -208,11 +211,14 @@ func (s *Server) PutTransactions(ctx context.Context, req *metamorph_api.Transac
...
+		callback := store.StoreCallback{
+			CallbackURL:   txReq.GetCallbackUrl(),
+			CallbackToken: txReq.GetCallbackToken(),
+		}
 		sReq := &store.StoreData{
 			Hash:              hash,
 			Status:            statusReceived,
-			CallbackUrl:       txReq.GetCallbackUrl(),
-			CallbackToken:     txReq.GetCallbackToken(),
+			Callbacks:         []store.StoreCallback{callback},
```
kuba-4chain, line 221: Same thing here, if you think it's a better way

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -375,6 +385,23 @@ func (p *PostgreSQL) SetBulk(ctx context.Context, data []*store.StoreData) error
...
+
+func readCallbacksFromDB(callbacks string) ([]store.StoreCallback, error) {
+	var callbacksData []store.StoreCallback
+	err := json.Unmarshal([]byte(callbacks), &callbacksData)
+	if err != nil {
+		return nil, err
+	}
+	return callbacksData, nil
+}
+
```
kuba-4chain, line 404: Move these to `postgres_helpers.go`

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -98,8 +98,7 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
 	var status sql.NullInt32
 	var blockHeight sql.NullInt64
 	var blockHash []byte
-	var callbackUrl sql.NullString
-	var callbackToken sql.NullString
+	var callbacksData sql.NullString
```
kuba-4chain, line 101: I think it should be `[]byte`

internal/metamorph/store/postgresql/postgres_helpers.go
```diff
@@ -50,8 +50,7 @@ func getStoreDataFromRows(rows *sql.Rows) ([]*store.StoreData, error) {
 		var blockHeight sql.NullInt64
 		var blockHash []byte
 
-		var callbackUrl sql.NullString
-		var callbackToken sql.NullString
+		var callbacksData sql.NullString
```
kuba-4chain, line 53: Probably should be `[]byte`

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -375,6 +385,23 @@ func (p *PostgreSQL) SetBulk(ctx context.Context, data []*store.StoreData) error
...
+	callbacksBytes, err := json.Marshal(callbacks)
+	if err != nil {
+		return nil, err
+	}
+	return callbacksBytes, nil
+}
+
+func readCallbacksFromDB(callbacks string) ([]store.StoreCallback, error) {
+	var callbacksData []store.StoreCallback
+	err := json.Unmarshal([]byte(callbacks), &callbacksData)
```
kuba-4chain, line 398: Then here, you wouldn't have to retype the string to []byte for unmarshalling, you could just do this:
```suggestion
func readCallbacksFromDB(callbacks []byte) ([]store.StoreCallback, error) {
	var callbacksData []store.StoreCallback
	err := json.Unmarshal(callbacks, &callbacksData)
```
But I may be wrong so please verify :)

internal/metamorph/callbacker.go
boecklim: I think it is better to send them one by one.

That seems to be a problem with the way we send callbacks currently. As we send them in parallel we can overwhelm the callback endpoints. That can lead to failing callbacks even though the callback endpoints exist.

internal/metamorph/store/postgresql/migrations/000016_multiple_callbacks.up.sql
```diff
@@ -0,0 +1,17 @@
...
+    ADD COLUMN callbacks JSONB;
+
+-- Step 2: Populate the 'callback' column with data from 'callback_url' and 'callback_token'
+UPDATE metamorph.transactions
+SET callbacks = json_build_array(
+    json_build_object(
+        'callback_url', callback_url,
+        'callback_token', callback_token
+    )
+);
```
boecklim, line 12: nit: In order not to have this security warning: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&pullRequest=539&types=BUG&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&inNewCodePeriod=true&sinceLeakPeriod=true&id=bitcoin-sv_arc

add
```
WHERE 'callback_url' != '' OR 'callback_token' != ''
```

internal/metamorph/store/postgresql/migrations/000016_multiple_callbacks.down.sql
```diff
@@ -0,0 +1,13 @@
+-- Step 1: Add the old 'callback_url' and 'callback_token' columns back
+ALTER TABLE metamorph.transactions
+    ADD COLUMN callback_url TEXT,
+    ADD COLUMN callback_token TEXT;
+
+-- Step 2: Populate 'callback_url' and 'callback_token' with the first object in the 'callbacks' JSON array
+UPDATE metamorph.transactions
+SET callback_url = (callbacks->0->>'callback_url'),
+    callback_token = (callbacks->0->>'callback_token');
```
boecklim, line 9: nit: In order not to have this security warning: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&pullRequest=539&types=BUG&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&inNewCodePeriod=true&sinceLeakPeriod=true&id=bitcoin-sv_arc

add something like
```
WHERE jsonb_array_length(callbacks) > 0
```

internal/metamorph/server.go
pawellewandowski98: Done

internal/metamorph/server.go
pawellewandowski98: Done

internal/metamorph/store/postgresql/postgres.go
pawellewandowski98: Done

internal/metamorph/store/postgresql/postgres_helpers.go
pawellewandowski98: Done

internal/metamorph/store/postgresql/postgres.go
pawellewandowski98: Done

internal/metamorph/store/postgresql/migrations/000016_multiple_callbacks.down.sql
pawellewandowski98: Done

internal/metamorph/store/postgresql/migrations/000016_multiple_callbacks.up.sql
pawellewandowski98: I used `IS NOT NULL` because `!= ''` cause another error

kuba-4chain (approved):

internal/metamorph/callbacker.go
```diff
@@ -135,23 +140,26 @@ func (p *Callbacker) SendCallback(logger *slog.Logger, tx *store.StoreData) {
...
 		var response *http.Response
 		response, err = p.httpClient.Do(request)
 		if err != nil {
-			logger.Debug("Couldn't send transaction callback", slog.String("url", tx.CallbackUrl), slog.String("token", tx.CallbackToken), slog.String("hash", tx.Hash.String()), slog.String("err", err.Error()))
+			logger.Debug("Couldn't send transaction callback", slog.String("url", callback.CallbackURL), slog.String("token", callback.CallbackToken), slog.String("hash", tx.Hash.String()), slog.String("err", err.Error()))
 			continue
 		}
-		defer response.Body.Close()
+
+		defer func() {
```
arkadiuszos4chain, line 160: Why did you changed one-line `defer response.Body.Close()` into this anonymous function?

internal/metamorph/store/postgresql/migrations/000016_multiple_callbacks.up.sql
```diff
@@ -0,0 +1,17 @@
...
+    ADD COLUMN callbacks JSONB;
+
+-- Step 2: Populate the 'callback' column with data from 'callback_url' and 'callback_token'
+UPDATE metamorph.transactions
+SET callbacks = json_build_array(
+    json_build_object(
+        'callback_url', callback_url,
+        'callback_token', callback_token
+    )
+)WHERE callback_url IS NOT NULL OR callback_token IS NOT NULL;
```
arkadiuszos4chain, line 12: I think we can't do anything with callback without url so probably we can ignore them

arkadiuszos4chain: I'm also almost 100% sure that we have empty strings instead of null in all records, so it would be better to check the length

internal/metamorph/callbacker.go
pawellewandowski98: IDE/linter will not return information about unhandled error

internal/metamorph/store/postgresql/migrations/000016_multiple_callbacks.up.sql
pawellewandowski98: Done

internal/metamorph/callbacker.go
pawellewandowski98: Reverted change

arkadiuszos4chain (approved):

boecklim (approved):

PR #542, opened 2024-08-09
feat(ARCO-169): refactor ResponseProcessor and competingTxs string to NullString

kuba-4chain:
### What's been done
1. Switch from `map` to `sync.Map` in `ResponseProcessor` to avoid accidental race conditions.
2. Better handling of null and empty string values when retrieving `competing_txs` from database.

### DIFF
--- a/internal/metamorph/store/postgresql/postgres.go
+++ b/internal/metamorph/store/postgresql/postgres.go
@@ -100,7 +100,7 @@ func (p *PostgreSQL) Get(ctx context.Context, hash []byte) (*store.StoreData, er
 	var callbacksData []byte
 	var fullStatusUpdates bool
 	var rejectReason sql.NullString
-	var competingTxs string
+	var competingTxs sql.NullString
 	var rawTx []byte
 	var lockedBy string
 	var merklePath sql.NullString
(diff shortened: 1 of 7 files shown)


internal/metamorph/store/postgresql/postgres.go
```diff
@@ -642,7 +635,8 @@ func (p *PostgreSQL) UpdateDoubleSpend(ctx context.Context, updates []store.Upda
 			) AS bulk_query
 			WHERE metamorph.transactions.hash=bulk_query.hash
 				AND metamorph.transactions.status <= bulk_query.status
-				AND LENGTH(metamorph.transactions.competing_txs) < LENGTH(bulk_query.competing_txs)
+				AND (metamorph.transactions.competing_txs IS NULL
+						OR LENGTH(metamorph.transactions.competing_txs) < LENGTH(bulk_query.competing_txs))
```
boecklim, line 639: That means that the competing txs will only be updated if the updates contain more competing txs than the existing one, right?

What if there is just one more competing tx that should get added which is not among the existing ones, shouldn't that also be updated? 🤔

boecklim (approved):

internal/metamorph/store/postgresql/postgres.go
kuba-4chain: If you look at line `678` and then at line `693`, we first get already existing competing transactions from the database and then we merge them with incoming transactions.

Therefore:
* if at least one new transaction is coming that was not previously there, it will be merged with already existing transactions and updated in the record
* if no new transactions are coming, or if the incoming transactions are already existing in the db, we will avoid the update to save resources

arkadiuszos4chain (approved):

PR #546, opened 2024-08-12
Refactor/e2e tests

boecklim:
- Remove dependency to ARC from e2e tests
- Use generic functions for POSTing and GETting requests to ARC
- Rename tests and test files
- Use test table if there are multiple test cases
- Remove some duplicate tests


test/submit_single_test.go
```diff
@@ -0,0 +1,337 @@
...
+			t.Fatalf("callback error: %v", err)
+		case <-callbackTimeout:
+			t.Fatal("callback exceeded timeout")
+		}
+	})
+}
+
+func TestSubmitQueued(t *testing.T) {
+	t.Run("queued", func(t *testing.T) {
+		address, privateKey := getNewWalletAddress(t)
```
arkadiuszos4chain, line 156: Lines: 156-160 can be replaced with: `address, privateKey := fundNewWallet(t)`

test/submit_single_test.go
```diff
@@ -0,0 +1,337 @@
...
+		skipFeeValidation bool
+		skipTxValidation  bool
+	}{
+		{
+			name:              "post transaction with skip fee",
+			skipFeeValidation: true,
+			skipTxValidation:  false,
+		},
+		{
+			name:              "post transaction with skip script validation",
```
arkadiuszos4chain, line 302: This test case skips whole tx validation not only script validation

test/submit_single_test.go
```diff
@@ -0,0 +1,337 @@
...
+			skipFeeValidation: false,
+			skipTxValidation:  true,
+		},
+		{
```
arkadiuszos4chain, line 306: `skipFeeValidation` is redundant here.
We could add test case where we setup `X-SkipFeeValidation = false` and `X-SkipScriptValidation= false` but `X-SkipTxValidation = true` and check if invalid transaction will be accepted by ARC

kuba-4chain (approved):

arkadiuszos4chain (approved):

pawellewandowski98 (approved):

PR #547, opened 2024-08-13
Transaction processing micro optimisation

kuba-4chain:
This PR introduces context to transaction processing in metamorph.

This helps to:
- avoid a situation in which a status is still being sent to the channel, even tho a response was sent to the user and nothing is consuming from this channel anymroe
- avoid blocking when the channel is unconsumed and its capacity full

### DIFF
--- a/internal/metamorph/response_processor.go
+++ b/internal/metamorph/response_processor.go
@@ -1,14 +1,15 @@
 package metamorph
 
 import (
+	"context"
 	"sync"
-	"time"
 
 	"github.com/bitcoin-sv/arc/internal/metamorph/metamorph_api"
 	"github.com/libsv/go-p2p/chaincfg/chainhash"
 )
 
 type StatusResponse struct {
+	ctx      context.Context
 	statusCh chan StatusAndError
 	mu       sync.RWMutex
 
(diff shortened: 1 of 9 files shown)


internal/metamorph/response_processor.go
```diff
@@ -1,46 +1,39 @@
...
-
-	r.mu.Unlock()
-
-	if r.statusCh != nil {
-		r.statusCh <- StatusAndError{
-			Hash:         r.Hash,
-			Status:       statusAndError.Status,
-			Err:          statusAndError.Err,
-			CompetingTxs: statusAndError.CompetingTxs,
+	if r.statusCh != nil && r.ctx != nil {
```
boecklim, line 26: nit: I find it usually nice if the happy path is on the left which IMO improves readability.
That principle is nicely explained here: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88
What do you think?

```
func (r *StatusResponse) UpdateStatus(statusAndError StatusAndError) {
	if r.statusCh == nil || r.ctx == nil {
		return
	}

	select {
	case <-r.ctx.Done():
		return
	default:
		r.statusCh <- StatusAndError{
			Hash:         r.Hash,
			Status:       statusAndError.Status,
			Err:          statusAndError.Err,
			CompetingTxs: statusAndError.CompetingTxs,
		}
	}
}
```

boecklim (approved):

internal/metamorph/response_processor.go
kuba-4chain: Totally agree. Especially in Go this is a good principle. Changed

arkadiuszos4chain (approved):
Really like it!

PR #549, opened 2024-08-13
Fix/message channel

boecklim:
- Ensure writing to storage status update channel is non-blocking
- Eliminate small function
- Increase buffer size


internal/metamorph/processor.go
```diff
@@ -345,8 +345,28 @@ func (p *Processor) StartSendStatusUpdate() {
...
+				select {
+				case p.storageStatusUpdateCh <- store.UpdateStatus{
+					Hash:         *msg.Hash,
+					Status:       msg.Status,
+					Error:        msg.Err,
+					CompetingTxs: msg.CompetingTxs,
+				}:
+				default: // Ensure that writing to channel is non-blocking
+				}
+
```
kuba-4chain, line 360: Aren't we going to lose data this way?

If the `p.storageStatusUpdateCh` will be full, then we will just simply NOT send the status there and lose that information.

Is this intentional?

kuba-4chain (approved):

PR #557, opened 2024-08-19
ARCO-175: move the callback logic to a separate microservice

arkadiuszos4chain:
* add callback microservice
* call callback service via gRPC in metamorph
* update GH workflow (use linter v 1.60.1)


internal/callbacker/server.go
```diff
@@ -0,0 +1,145 @@
...
+	}()
+
+	return nil
+}
+
+func (s *Server) Shutdown() {
+	s.logger.Info("Shutting down")
+
+	s.grpcServer.GracefulStop()
+	s.grpcServer.Stop()
```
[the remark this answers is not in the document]
arkadiuszos4chain: Good question. Probably not, I'll remove it

internal/callbacker/callbacker_api/callbacker_api.proto
```diff
@@ -0,0 +1,55 @@
...
+  ANNOUNCED_TO_NETWORK = 40;
+  REQUESTED_BY_NETWORK = 50;
+  SENT_TO_NETWORK = 60;
+  ACCEPTED_BY_NETWORK = 70;
+  SEEN_IN_ORPHAN_MEMPOOL = 80;
+  SEEN_ON_NETWORK = 90;
+  DOUBLE_SPEND_ATTEMPTED = 100;
+  REJECTED = 110;
+  MINED = 120;
+}
```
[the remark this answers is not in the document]
arkadiuszos4chain: That was my first thought, but:

1. based on the statuses, we also take some actions in the `callbackerc service (e.g., prometheus statistics)
2. message containing a number is smaller (premature optimization, I admit it)

Even if the second point isn't worth implementing the enum on the `callbacker` side, the first point essentially requires it.

Or maybe I could move its definition to a higher level and import it into both `metamorph.proto` and `callbacker.proto`? WDYT?

internal/callbacker/callbacker_api/callbacker_api.proto
```diff
@@ -0,0 +1,55 @@
...
+
+service CallbackerAPI{
+  rpc Health (google.protobuf.Empty) returns (HealthResponse) {}
+  rpc SendCallback (SendCallbackRequest) returns (google.protobuf.Empty) {}
+}
+
+// Note: Values of the statuses have a difference between them in case
+// it's necessary to add another values in between. This will allow to
+// create new statuses without changing the values in the existing db.
+enum Status {
```
boecklim, line 18: Could we use here just an integer instead?

I think `callbacker` probably doesn't need to know any logic about statuses anyway

internal/callbacker/callbacker_api/callbacker_api.proto
```diff
@@ -0,0 +1,55 @@
...
+// swagger:model HealthResponse
+message HealthResponse {
+  google.protobuf.Timestamp timestamp = 1;
+}
+
+// swagger:model SendCallbackRequest
+message SendCallbackRequest {
+  repeated CallbackEndpoint callback_endpoints = 1;
+  string txid = 2;
+  Status status = 3;
```
boecklim, line 43: Could we use here just an integer instead?

I think `callbacker` probably doesn't need to know any logic about statuses anyway

internal/callbacker/callbacker_api/callbacker_api.proto
arkadiuszos4chain: We have prometeus stats based on the status
image

boecklim: I see now, it's used here: https://github.com/bitcoin-sv/arc/blob/93d51c633c3983b586c392391355bc18cf28dd5e/internal/callbacker/callbacker.go#L179

Nevermined

internal/callbacker/callbacker_api/callbacker_api.proto
boecklim: I see now, it's used here: https://github.com/bitcoin-sv/arc/blob/93d51c633c3983b586c392391355bc18cf28dd5e/internal/callbacker/callbacker.go#L179

Nevermined

internal/metamorph/grpc_callbacker.go
boecklim, line 1: If we want to follow the existing pattern, then this file should be in callbacker `pkg/callbacker`. And filename would be `client.go`.
Every service which needs to call `callbacker` can then import this client.

For example this is how it is done for metamorph: `pkg/metamorph/client.go`

internal/callbacker/callbacker_api/callbacker_api.proto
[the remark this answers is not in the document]
kuba-4chain: > move its definition to a higher level and import it into both metamorph.proto and callbacker.proto?

I think it's a great idea, I'm in favour. If you want to do this in the next PR, you can resolve this comment :)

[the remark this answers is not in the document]
arkadiuszos4chain: Great, I'll make this change in separate PR

internal/metamorph/grpc_callbacker.go
arkadiuszos4chain: If I understand correctly, we want to completely isolate the `internal/metamorph` and `internal/callbacker packages` by introducing an additional layer `pkg/callbacker`?

arkadiuszos4chain: Still inside the metamorph we would need implementation of mappings between dtos

boecklim: Actually we can at some point remove the clients from `pkg/` folder (separate refactoring) as they don't need to be public really.

Ok, we could maybe refactor that later if needed. Resolving the conversation.

boecklim (approved):

cmd/arc/services/metamorph.go
```diff
@@ -286,3 +291,16 @@ func initPeerManager(logger *slog.Logger, s store.MetamorphStore, arcConfig *con
 
 	return pm, peerHandler, messageCh, nil
 }
+
+func initGrpCallbackerConn(address, prometheusEndpoint string, grpcMsgSize int) (callbacker_api.CallbackerAPIClient, error) {
```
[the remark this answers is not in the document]
arkadiuszos4chain: Fixed

kuba-4chain (approved):

PR #563, opened 2024-08-21
ARCO-182: send callbacks to the same url one by one

arkadiuszos4chain:
## Description of Changes

Callbacks are sent one by one to the same URL. In the previous implementation, each callback request created a new goroutine to send the callback, which could result in a potential DDoS of the callback receiver. The new approach sends callbacks to the same receiver in a serial manner. 

Note that URLs are not locked by the `callbacker` instance, so serial sends occur only within a single instance. In other words, the level of parallelism is determined by the number of `callbacker` instances.

## Testing Procedure

- [x] All tests pass locally

## Checklist:

- [x] I have made corresponding changes to the documentation
- [x] I have updated `CHANGELOG.md` with my changes


internal/callbacker/dispatcher.go
```diff
@@ -0,0 +1,108 @@
+package callbacker
+
+import (
+	"sync"
+	"time"
+)
+
+type CallbackDispatcher struct {
+	c        CallbackerI
+	managers map[string]*sendManager
```
boecklim, line 10: As an alternative strategy, here you could use a `sync.Map`
Here is an example where it is used: https://github.com/bitcoin-sv/arc/blob/6669b2339f9b019dec239acc264bf0214d705e33/internal/metamorph/response_processor.go#L44

Then you don't need the mutex and can remove all the lock/unlock statements
```
	d.mu.Lock()
	defer d.mu.Unlock()
```

kuba-4chain (approved):

internal/callbacker/sender.go
```diff
@@ -91,24 +89,19 @@ func (p *CallbackSender) Health() error {
...
-				slog.String("hash", dto.Txid),
-				slog.Int("retries", retries))
-
-			p.stats.callbackFailedCount.Inc()
-		}
-	}()
+	ok := p.sendCallbackWithRetries(url, token, dto)
+
+	if ok {
+		p.updateSuccessStats(dto.TxStatus)
```
boecklim, line 95: nit: if you return here you could remove `else { ... }`

internal/callbacker/dispatcher.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (m *sendManager) GracefulStop() {
+	m.wg.Wait() // wait for all accepted callbacks to be consumed
+	close(m.ch) // signal the `run` goroutine to exit
+}
+
+func (m *sendManager) run() {
+	go func() {
+		for {
```
boecklim, line 96: nit: Equivalent alternative with slightly less code

```
		for callback := range m.ch {

			m.c.Send(m.url, callback.token, callback.data)
			m.wg.Done()

			time.Sleep(m.sleep)
		}
```

arkadiuszos4chain: I cannot change it since I need to check if the channel is still open. Error from the compiler:
`range over m.ch (variable of type chan *callbackEntry) permits only one iteration variable`

internal/callbacker/dispatcher.go
arkadiuszos4chain: Unfortunately, with sync.Map, I am unable to check if a manager exists and create one only if it doesn't, in an atomic way.

The code I was able to write is:
```go
func (d *CallbackDispatcher) GracefulStop() {
	d.managers.Range(func(key, value any) bool {
		m := value.(*sendManager)
		m.GracefulStop()

		return true
	})
}

func (d *CallbackDispatcher) dispatch(url, token string, dto *Callback) {
	m, loaded := d.managers.LoadOrStore(url, &sendManager{
		url:   url,
		c:     d.c,
		sleep: d.sleep,

		ch: make(chan *callbackEntry),
	})

	manager := m.(*sendManager)
	if !loaded {
		manager.run()
	}

	manager.Add(token, dto)
}
```

I don't think this code is simpler than what already exists, and  TBH I'm not a fan of the idea of creating new manager structure for every call.

internal/callbacker/sender.go
arkadiuszos4chain: Done

internal/callbacker/dispatcher.go
boecklim: Ok 👌

internal/callbacker/dispatcher.go
boecklim: Ok. Understood 👍

PR #569, opened 2024-08-27
ARCO-176: (Reorg Support) Mark block as STALE or LONGEST

kuba-4chain:
## Description of Changes

1. Blocks table in BlockTx store has now 2 new fields: `status` and `chainwork`.
2. Unique index on block height was removed - allowing for storing multiple blocks at the same height in BlockTx store.
3. Method for calculating `chainwork` from block bits was added.
4. Incoming blocks are given status `LONGEST`.
5. Incoming blocks that are competing on a given block height are given status `STALE` or `LONGEST` based on their chainwork.
6. Integration test for functionalities described above was added.

#### Out of scope
1. Perform Reorg - update statuses for competing and previous blocks - will be done in #ARCO-177
2. Update transactions statuses in case of forks and reorgs - will be done in #ARCO-178 and #ARCO-179
3. Improvements to the integration test - each next PR with come with a small improvement to the test

## Testing Procedure

- [x] All tests pass locally

## Checklist:


internal/blocktx/integration_test/reorg_integration_test.go
```diff
@@ -0,0 +1,210 @@
...
+// Flow of this test:
+// 		1. A list of blocks from height 822014 to 822017 is added to db from fixtures
+// 		2. A hardcoded msg with competing block at height 822015 is being send through the mocked PeerHandler
+// 		3. This block has a chainwork lower than the current tip of chain - becomes STALE
+// 		4. Next competing block, at height 822016 is being send through the mocked PeerHandler
+// 		5. This block has a greater chainwork than the current tip of longest chain - it becomes LONGEST despite not being the highest
+//
+// Future implementation in next tasks:
+// 		- Verify if reorg was performed correctly, if previous blocks have updated statuses
+// 		- Include metamorph in this test and verify that transactions statuses are properly updated
```
boecklim, line 17: nit:
How about
```
// Todo: Next tasks:
// 		- Verify if reorg was performed correctly, if previous blocks have updated statuses
// 		- Include metamorph in this test and verify that transactions statuses are properly updated
```

internal/blocktx/integration_test/reorg_integration_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	err = pool.Purge(resource)
+	if err != nil {
+		log.Fatalf("failed to purge pool: %v", err)
+	}
+
+	os.Exit(code)
+}
+
+func TestReorg(t *testing.T) {
```
boecklim, line 135: This this test really Reorg?

I think it is testing just a new block being received at the same hight, but smaller chain work, not?

internal/blocktx/processor.go
```diff
@@ -481,38 +401,69 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
...
+	competing, err := p.competingChainsExist(ctx, incomingBlock)
+	if err != nil {
+		p.logger.Error("unable to check for competing chains", slog.String("hash", blockHash.String()), slog.Uint64("height", msg.Height), slog.String("err", err.Error()))
+		return
+	}
+
+	if competing {
+		p.logger.Info("Competing blocks found", slog.String("incoming block hash", blockHash.String()), slog.Uint64("height", incomingBlock.Height))
+
+		greatestChainwork, err := p.hasGreatestChainwork(ctx, incomingBlock)
```
boecklim, line 424: nit: It would be more visible that `greatestChainwork` is boolean if it were called eg.e `hasGreatestChainwork`

internal/blocktx/processor.go
```diff
@@ -481,38 +401,69 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
...
+			p.logger.Error("unable to get the chain tip to verify chainwork", slog.String("hash", blockHash.String()), slog.Uint64("height", incomingBlock.Height), slog.String("err", err.Error()))
+			return
+		}
+
+		if greatestChainwork {
+			// TODO: perform reorg - next ticket
+			incomingBlock.Status = blocktx_api.Status_LONGEST
+		} else {
+			incomingBlock.Status = blocktx_api.Status_STALE
+		}
```
boecklim, line 435: nit:

`else` could be elimintated by 

```
incomingBlock.Status = blocktx_api.Status_STALE
if greatestChainwork {
  // TODO: perform reorg - next ticket
  incomingBlock.Status = blocktx_api.Status_LONGEST
} 
```

internal/blocktx/processor.go
```diff
@@ -525,27 +476,97 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
...
+
+func (p *Processor) competingChainsExist(ctx context.Context, block *blocktx_api.Block) (bool, error) {
+	if block.Status == blocktx_api.Status_ORPHANED {
+		return false, nil
+	}
+
+	if block.Status == blocktx_api.Status_LONGEST {
+		competingBlock, err := p.store.GetBlockByHeight(ctx, block.Height, blocktx_api.Status_LONGEST)
+		if err != nil && !errors.Is(err, store.ErrBlockNotFound) {
+			blockHash := (*chainhash.Hash)(block.Hash)
```
boecklim, line 514: Thinking if it wouldn't be better to use the chainhash function here as the conversion could fail and it is not checked here🤔 

```
blockHash, err := chainhash.NewHash(block.Hash)
```

internal/blocktx/processor.go
```diff
@@ -525,27 +476,97 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
...
+	}
+
+	// If STALE status
+	return true, nil
+}
+
+func (p *Processor) hasGreatestChainwork(ctx context.Context, incomingBlock *blocktx_api.Block) (bool, error) {
+	tip, err := p.store.GetChainTip(ctx)
+	if err != nil && !errors.Is(err, store.ErrBlockNotFound) {
+		blockHash := (*chainhash.Hash)(incomingBlock.Hash)
```
boecklim, line 536: Thinking if it wouldn't be better to use the chainhash function here as the conversion could fail and it is not checked here🤔 

```
blockHash, err := chainhash.NewHash(block.Hash)
```

internal/blocktx/integration_test/reorg_integration_test.go
```diff
@@ -0,0 +1,210 @@
+package integrationtest
+
+// Components of this test:
+// 		Postgresql Store - running on docker
+// 		Blocktx Processor
+// 		PeerHandler - mocked
+//
+// Flow of this test:
+// 		1. A list of blocks from height 822014 to 822017 is added to db from fixtures
+// 		2. A hardcoded msg with competing block at height 822015 is being send through the mocked PeerHandler
```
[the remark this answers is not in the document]
kuba-4chain: Thanks, fixed

internal/blocktx/integration_test/reorg_integration_test.go
kuba-4chain: Well, the idea is for it to become the Reorg test in the near future (as per the next tasks described in the top). It's just that in this PR we don't perform reorg yet. So this test just checks if the first incoming competing block with smaller chainwork will receive status `STALE` and then the next incoming competing block with a higher chainwork will receive status `LONGEST`.

But I changed the name to `TestBlockStatus` to be more clear, and will rename it back to `TestReorg` when we have reorg implemented. I hope that's okay :)

internal/blocktx/processor.go
kuba-4chain: Right, changed

internal/blocktx/processor.go
kuba-4chain: I don't think it can be eliminated, let me explain.

In this implementation, the incoming block first gets the status of the previous block. So if the previous block was `LONGEST` the incoming block will initially also be `LONGEST`. This is done this way, because this will be the most likely and most common scenario for all incoming blocks (assuming that forks are rather rare).

Only then we check for competing blocks at the height of this incoming block and mark the block accordingly (STALE or LONGEST) based on chainwork.

So, the first competing block that will come at any height, will have the `LONGEST` block as it's previous block and will initially be also marked as `LONGEST`. Then we will see that this block is actually a competing block, so if it has a lower chainwork, we have to mark is as `STALE`.

This all happens before any write to the database, and with all possible edge cases I came up with, this is the easiest and most efficient way to do this. Hope I explained it well :)

internal/blocktx/processor.go
kuba-4chain: This should never ever happen in current implementation, but you're right, we should check for error here in case some other logic will ever be added to this code. I changed it.

internal/blocktx/processor.go
kuba-4chain: Changed

internal/blocktx/integration_test/reorg_integration_test.go
boecklim: Alright, nice 👍

boecklim (approved):

shotasilagadzetaal:
One question, why was there need for chainwork field to be added?

kuba-4chain:
@shotasilagadzetaal

The longest chain, in case of forks, is figured out based on this field `chainwork`, which is a standard practice in blockchain.

Usually, blocks on the same height will have equal chainwork, but it’s not a rule.

Similarly, a chain with greater block height will most likely have a greater chainwork, but it’s
also not a rule.

Therefore, the chain, which **tip** (last, highest block) has the greatest chainwork, is
considered to be the longest chain. Nodes are incentivised to build upon that chain.

Here's a great article that explains it in more detail: https://learnmeabitcoin.com/technical/blockchain/longest-chain/

internal/blocktx/processor_helpers.go
```diff
@@ -0,0 +1,110 @@
...
+func createBlock(msg *p2p.BlockMessage, prevBlock *blocktx_api.Block, longestTipExists bool) *blocktx_api.Block {
+	hash := msg.Header.BlockHash()
+	prevHash := msg.Header.PrevBlock
+	merkleRoot := msg.Header.MerkleRoot
+	chainwork := calculateChainwork(msg.Header.Bits)
+
+	var status blocktx_api.Status
+	if prevBlock == nil {
+		if longestTipExists {
+			status = blocktx_api.Status_ORPHANED
```
boecklim, line 42: That is the extreme edge case that the first block already has a competing block, right?
Wouldn't it be `Status_STALE` then? 🤔

boecklim (approved):

internal/blocktx/store/postgresql/migrations/000013_add_block_status.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE blocks
```
boecklim, line 1: blocktx.blocks

internal/blocktx/store/postgresql/migrations/000014_remove_index_on_block_height.down.sql
```diff
@@ -0,0 +1,3 @@
+CREATE UNIQUE INDEX pux_blocks_height ON blocks(height)
```
boecklim, line 1: blocktx.blocks

internal/blocktx/store/postgresql/migrations/000013_add_block_status.up.sql
kuba-4chain: done

internal/blocktx/store/postgresql/migrations/000014_remove_index_on_block_height.down.sql
kuba-4chain: done

internal/blocktx/processor_helpers.go
kuba-4chain: No no, this is for the edge case that there is no block in the database, it's empty.

Then, the first block that comes will be marked as `ORPHANED` because it doesn't have any previous blocks and then every next block will also be marked as `ORPHANED` because it's extending the orphaned chain.

To avoid that situation, I'm checking if there is any block in the db (by using the getTip method) and if there isn't, then the incoming block is `LONGEST`. If there is, then the incoming block (without the prevBlock in our db) will be marked as `ORPHANED`, correctly.

Please let me know if I explained it well

kuba-4chain: Also, I left the further handling of ORPHANED blocks out of scope of this PR. I should probably create a ticket for that.

kuba-4chain: By "further handling" I mean:
1. Removing the `orhpanedyn` field, and using the newly created status.
2. Putting the blocks out of `ORPHANED` state into either `LONGEST` or `STALE` in case the previous block comes to our database

shotasilagadzetaal:
> @shotasilagadzetaal
> 
> The longest chain, in case of forks, is figured out based on this field `chainwork`, which is a standard practice in blockchain.
> 
> Usually, blocks on the same height will have equal chainwork, but it’s not a rule.
> 
> Similarly, a chain with greater block height will most likely have a greater chainwork, but it’s also not a rule.
> 
> Therefore, the chain, which **tip** (last, highest block) has the greatest chainwork, is considered to be the longest chain. Nodes are incentivised to build upon that chain.
> 
> Here's a great article that explains it in more detail: https://learnmeabitcoin.com/technical/blockchain/longest-chain/

yeah, I get that but my question is why do WE need it? Couldn't we simply get the latest block from the chain ? 

It looks like we are doing consensus stuff on our side by that field which may not be necessary I think. All we need basically is to check if the latest block is still what we think it is and if not go back (cancelling transactions), reach the common ancestor of our wrong block and new chain tip block and processing them to the right direction (adding transactions back again).

internal/blocktx/store/postgresql/migrations/000014_remove_index_on_block_height.up.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX pux_blocks_height;
```
boecklim, line 1: blocktx.pux_blocks_height

internal/blocktx/store/postgresql/migrations/000014_remove_index_on_block_height.down.sql
```diff
@@ -0,0 +1,3 @@
+CREATE UNIQUE INDEX pux_blocks_height ON blocktx.blocks(height)
```
boecklim, line 1: blocktx.pux_blocks_height

kuba-4chain:
@shotasilagadzetaal 

We're not doing any consensus stuff on our side, we're just updating the status of blocks (and transactions - in next PRs) based on what information we receive from nodes. Chainwork, calculated from bits, is one piece of that information, that is the best indicator of which chain might become the longest.

> Couldn't we simply get the latest block from the chain ?

There is no simple way to just "get the latest block from the chain", especially since we're processing blocks as they come, we cannot wait for couple of more blocks to be mined to say which one of the competing blocks is the correct one. But it's possible that I don't understand what you mean well, so if you could explain a bit more, I'd appreciate that.

Imagine a situation in which we get a message about new block at height 822012 from one peer (node). Then, 5 seconds later, we receive information about another new block also at height 822012 from another peer. How would you handle that?

> All we need basically is to check if the latest (canonical) block is still what we think it is and if not go back (cancelling transactions on the way), reach the common ancestor of our wrong block and new chain tip block (from new canonical chain) and processing them to the right direction to the new canonical chain tip block (adding transactions back again).

I'm not really sure what you mean, please explain in more detail how would you do it in the current architecture of blocktx and peer-to-peer communication with the nodes 🙏

internal/blocktx/processor.go
```diff
@@ -525,27 +486,118 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
...
+
+func (p *Processor) hasGreatestChainwork(ctx context.Context, incomingBlock *blocktx_api.Block) (bool, error) {
+	tip, err := p.store.GetChainTip(ctx)
+	if err != nil && !errors.Is(err, store.ErrBlockNotFound) {
+		blockHash, err := chainhash.NewHash(incomingBlock.Hash)
+		if err != nil {
+			p.logger.Error("failed to create blockHash", slog.Uint64("height", incomingBlock.Height), slog.String("err", err.Error()))
+			return false, err
+		}
+		_, errDel := p.store.DelBlockProcessing(ctx, blockHash, p.hostname)
```
arkadiuszos4chain, line 568: Please refactor the `processBlock` function to avoid repetition of `DelBlockProcessing` and side effects in functions like `hasGreatestChainwork`

arkadiuszos4chain (changes requested):

internal/blocktx/processor.go
```diff
@@ -525,27 +486,118 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
...
+	tipChainWork.SetString(tip.Chainwork, 10)
+
+	incomingBlockChainwork := new(big.Int)
+	incomingBlockChainwork.SetString(incomingBlock.Chainwork, 10)
+
+	if tipChainWork.Cmp(incomingBlockChainwork) < 0 {
+		return true, nil
+	}
+
+	return false, nil
```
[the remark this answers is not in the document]
kuba-4chain: Done

internal/blocktx/processor.go
```diff
@@ -481,38 +400,80 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
...
+	}
+
+	p.logger.Info("Inserting block", slog.String("hash", blockHash.String()), slog.Uint64("height", incomingBlock.Height), slog.String("status", incomingBlock.Status.String()))
+
+	blockId, err := p.store.InsertBlock(ctx, incomingBlock)
+	if err != nil {
+		_, errDel := p.store.DelBlockProcessing(ctx, &blockHash, p.hostname)
 		if errDel != nil {
-			ph.logger.Error("failed to delete block processing - after inserting block failed", slog.String("hash", blockHash.String()), slog.String("err", errDel.Error()))
+			p.logger.Error("failed to delete block processing - after inserting block failed", slog.String("hash", blockHash.String()), slog.String("err", errDel.Error()))
```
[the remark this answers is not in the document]
kuba-4chain: Done by refactor, logs from DelBlocksProcessing are last now.

internal/blocktx/processor.go
```diff
@@ -525,27 +486,118 @@ func (ph *Processor) processBlock(msg *p2p.BlockMessage) {
 		TxCount:      uint64(len(msg.TransactionHashes)),
 	}
 
-	if err = ph.markBlockAsProcessed(ctx, block); err != nil {
-		_, errDel := ph.store.DelBlockProcessing(ctx, &blockHash, ph.hostname)
+	if err = p.markBlockAsProcessed(ctx, block); err != nil {
+		_, errDel := p.store.DelBlockProcessing(ctx, &blockHash, p.hostname)
 		if errDel != nil {
-			ph.logger.Error("failed to delete block processing - after marking block as processed failed", slog.String("hash", blockHash.String()), slog.String("err", errDel.Error()))
+			p.logger.Error("failed to delete block processing - after marking block as processed failed", slog.String("hash", blockHash.String()), slog.String("err", errDel.Error()))
```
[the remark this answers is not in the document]
kuba-4chain: Done by refactor

internal/blocktx/processor.go
kuba-4chain: Very good point, done

internal/blocktx/store/postgresql/get_block.go
```diff
@@ -13,12 +13,14 @@ import (
...
 	q := `
 		SELECT
-		 b.hash
-		,b.prevhash
-		,b.merkleroot
-		,b.height
-		,b.processed_at
-		,b.orphanedyn
+			b.hash,
+			b.prevhash,
```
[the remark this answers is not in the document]
kuba-4chain: I didn't know that, refactored to the previous version :)

internal/blocktx/store/postgresql/get_block_by_height.go
```diff
@@ -0,0 +1,50 @@
...
+		FROM blocktx.blocks b
+		WHERE b.height = $1 AND b.status = $2
+	`
+
+	var block blocktx_api.Block
+
+	var processed_at sql.NullString
+
+	if err := p.db.QueryRowContext(ctx, q, height, status).Scan(
+		&block.Hash,
```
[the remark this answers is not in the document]
kuba-4chain: That's a great idea, allowed to greatly refactor all the GET methods from the store, thanks

arkadiuszos4chain (approved):

PR #575, opened 2024-09-02
ARCO-184: persist unprocessed callbacks

arkadiuszos4chain:
## Description of Changes

On termination, the `callbacker` will save any unprocessed callbacks to the database instead of blocking termination to process them. When the instance is initialized, these callbacks will be retrieved from the database and processed.

## Linked Issues / Tickets

ARCO-184

## Testing Procedure

- [x] All tests pass locally

## Checklist:


internal/callbacker/dispatcher.go
```diff
@@ -38,12 +44,28 @@ func (d *CallbackDispatcher) GracefulStop() {
 	}
 }
 
+func (d *CallbackDispatcher) Init() error {
+	const bundleSize = 100
```
boecklim, line 48: I think for consistency to other services we could call this `batchSize`

internal/callbacker/dispatcher.go
```diff
@@ -87,22 +114,75 @@ func (m *sendManager) Add(token string, dto *Callback) {
...
 
 func (m *sendManager) GracefulStop() {
-	m.wg.Wait() // wait for all accepted callbacks to be consumed
+	m.stop <- struct{}{} // signal the `run` goroutine to stop processing
+	m.wg.Wait()          // wait for all accepted callbacks to be consumed
+
 	close(m.ch) // signal the `run` goroutine to exit
+
+	<-m.stop // wait for the `run` goroutine to exit
+	close(m.stop)
```
boecklim, line 123: Why not use a cancel context here? That would be more consistent to how we stop go routines in other services e.g. here
* https://github.com/bitcoin-sv/arc/blob/d357d5eb57cf627037110bc3a1ce26126b22432b/internal/metamorph/processor.go#L222
* https://github.com/bitcoin-sv/arc/blob/d357d5eb57cf627037110bc3a1ce26126b22432b/internal/metamorph/processor.go#L393

internal/callbacker/dispatcher.go
```diff
@@ -87,22 +114,75 @@ func (m *sendManager) Add(token string, dto *Callback) {
...
 }
 
 func (m *sendManager) run() {
 	go func() {
```
boecklim, line 127: Don't we also need to add 1 to the wait group here? 🤔

internal/callbacker/store/postgresql/postgres.go
```diff
@@ -0,0 +1,229 @@
...
+			blockHeights[i] = sql.NullInt64{Int64: int64(*d.BlockHeight), Valid: true}
+		} else {
+			blockHeights[i] = sql.NullInt64{}
+		}
+
+		if len(d.CompetingTxs) > 0 {
+			competingTxs[i] = sql.NullString{String: strings.Join(d.CompetingTxs, ","), Valid: true}
+		} else {
+			competingTxs[i] = sql.NullString{}
+		}
```
kuba-4chain, line 86: You can avoid the `else` statements and save some lines of code by a small refactor
```suggestion
		urls[i] = d.Url
		tokens[i] = d.Token
		timestamps[i] = d.Timestamp
		txids[i] = d.TxID
		txStatuses[i] = d.TxStatus
		
		extraInfos[i] = sql.NullString{}
		merklePaths[i] = sql.NullString{}
		blockHashes[i] = sql.NullString{}
		blockHeights[i] = sql.NullInt64{}
		competingTxs[i] = sql.NullString{}

		if d.ExtraInfo != nil {
			extraInfos[i] = sql.NullString{String: *d.ExtraInfo, Valid: true}
		}

		if d.MerklePath != nil {
			merklePaths[i] = sql.NullString{String: *d.MerklePath, Valid: true}
		}

		if d.BlockHash != nil {
			blockHashes[i] = sql.NullString{String: *d.BlockHash, Valid: true}
		}

		if d.BlockHeight != nil {
			blockHeights[i] = sql.NullInt64{Int64: int64(*d.BlockHeight), Valid: true}
		}

		if len(d.CompetingTxs) > 0 {
			competingTxs[i] = sql.NullString{String: strings.Join(d.CompetingTxs, ","), Valid: true}
		}
```

internal/callbacker/store/postgresql/postgres.go
```diff
@@ -0,0 +1,229 @@
...
+					,UNNEST($2::TEXT[])
+					,UNNEST($3::TEXT[])
+					,UNNEST($4::TEXT[])
+					,UNNEST($5::TEXT[])
+					,UNNEST($6::TEXT[])
+					,UNNEST($7::TEXT[])
+					,UNNEST($8::BIGINT[])
+					,UNNEST($9::TIMESTAMPTZ[])		
+					,UNNEST($10::TEXT[])					
+				ON CONFLICT DO NOTHING`
```
kuba-4chain, line 112: Maybe we should update on conflict? Because there might come a callback with the same tx, the same status, but for example one additional competing tx. 

callback 1: TxID: "123", `DOUBLE_SPEND_ATTEMPTED`, competingTx: [tx1]
callback 2: TxID: "123", `DOUBLE_SPEND_ATTEMPTED`, competingTx: [tx1, tx2]

cmd/arc/services/callbacker.go
```diff
@@ -20,23 +23,16 @@ func StartCallbacker(logger *slog.Logger, appConfig *config.ArcConfig) (func(),
...
-	if err != nil {
-		return nil, fmt.Errorf("GRPCServer failed: %v", err)
-	}
-
-	healthServer, err := StartHealthServerCallbacker(server, config.Health, logger)
-	if err != nil {
-		return nil, fmt.Errorf("failed to start health server: %v", err)
-	}
+		err error
+	)
```
kuba-4chain, line 35: If you'd move the creation of `stopFn` to the end of this function, you wouldn't have to declare these variables here and then check if they are != null in the `stopFn`

kuba-4chain: I missed that, thanks for pointing it out. Nice!

internal/callbacker/store/postgresql/postgres.go
arkadiuszos4chain: Oh, if there may be such a situation, I think I should persist both callbacks

internal/callbacker/store/postgresql/postgres.go
arkadiuszos4chain: You're right! I don't need those `elses`

internal/callbacker/dispatcher.go
arkadiuszos4chain: I use the `stop` channel in both directions:
1. to send information to stop processing
2. to receive information that the goroutine has finished

How can I achieve this with context cancellation?

internal/callbacker/dispatcher.go
arkadiuszos4chain: this wait group is used for accepted entries (callbacks), so it is modified in the `Add()` method.

internal/callbacker/dispatcher.go
arkadiuszos4chain: changed

internal/callbacker/store/postgresql/postgres.go
arkadiuszos4chain: changed

internal/callbacker/store/postgresql/postgres.go
arkadiuszos4chain: changed

internal/callbacker/dispatcher.go
boecklim: I think it could be done like this with a second wait group. But if not no worries. We could do it later if needed.

```

func (m *sendManager) GracefulStop() {
	m.cancelAll()

	m.wg.Wait() // wait for all accepted callbacks to be consumed

	m.wgRun.Wait()

	close(m.ch) // signal the `run` goroutine to exit

}

func (m *sendManager) run() {

	ctx, cancelAll := context.WithCancel(context.Background())

	m.ctx = ctx
	m.cancelAll = cancelAll

	go func() {
		var danglingCallbacks []*store.CallbackData

		defer func() {
			m.wgRun.Done()
		}()

	handleCallbacks:
		for {
			select {
			case callback, ok := <-m.ch:
				if !ok {
					break handleCallbacks
				}

				if m.stopping {
					// add callback to save
					danglingCallbacks = append(danglingCallbacks, toStoreDto(m.url, callback))
				} else {
					m.c.Send(m.url, callback.token, callback.data)
					time.Sleep(m.sleep)
				}

				m.wg.Done()

			case <-m.ctx.Done():
				m.stopping = true
			}
		}

		_ = m.s.SetMany(context.Background(), danglingCallbacks)
		m.stop <- struct{}{}
	}()
}
```

boecklim (approved):

internal/callbacker/dispatcher.go
arkadiuszos4chain: I'll try to change it in next PR

PR #586, opened 2024-09-11
feat(ARCO-194): update statuses in bulk query

pawellewandowski98:
## Description of Changes

Fix updating statuses in `UpdateStatusBulk`
- If there was more than one status change before last update, use `StatusHistory` field and current status when updating `status_history`
- If there is update with lower status than current one but it is not in status history add his history plus status to current status history 

## Linked Issues / Tickets

Fixes changes from ARCO-194

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


internal/metamorph/processor_helpers.go
```diff
@@ -9,14 +9,19 @@ func (p *Processor) GetProcessorMapSize() int {
...
 	if !found || shouldUpdateStatus(statusUpdate, foundStatusUpdate) {
 		if len(statusUpdate.CompetingTxs) > 0 {
 			statusUpdate.CompetingTxs = mergeUnique(statusUpdate.CompetingTxs, foundStatusUpdate.CompetingTxs)
 		}
 
+		statusUpdate.StatusHistory = append(foundStatusUpdate.StatusHistory, store.StatusWithTimestamp{
+			Status:    foundStatusUpdate.Status,
+			Timestamp: p.now(),
+		})
+
```
kuba-4chain, line 24: You can have a nil pointer exception here, because the if statement has also `!found` check. So you need to check if found and only then append

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -569,6 +569,7 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
 	txHashes := make([][]byte, len(updates))
 	statuses := make([]metamorph_api.Status, len(updates))
 	rejectReasons := make([]string, len(updates))
+	statusHistories := make([]*string, len(updates))
```
kuba-4chain, line 572: Does this have to be `*string` instead of just `string`?

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -577,26 +578,43 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+					)
 			FROM
 			(
-				SELECT *
-						FROM
-						UNNEST($2::BYTEA[], $3::INT[], $4::TEXT[])
-				AS t(hash, status, reject_reason)
+				SELECT t.hash, t.status, t.reject_reason, h.history_update
+				FROM UNNEST($2::BYTEA[], $3::INT[], $4::TEXT[]) AS t(hash, status, reject_reason)
+				LEFT JOIN UNNEST($5::JSONB[]) AS h(history_update) ON TRUE
```
kuba-4chain, line 615: Why LEFT JOIN here? Will this not work?
```suggestion
				SELECT t.hash, t.status, t.reject_reason, t.history_update
				FROM UNNEST($2::BYTEA[], $3::INT[], $4::TEXT[], $5::JSONB[]) AS t(hash, status, reject_reason, history_update)
```

pawellewandowski98: Yes it will work, I forgot to change it after trying many solutions. Now it should be ok

kuba-4chain (approved):

internal/metamorph/processor_helpers.go
```diff
@@ -22,78 +24,112 @@ func (p *Processor) GetProcessorMapSize() int {
...
 	}
-}
 
-func (p *Processor) updateStatusMap(statusUpdate store.UpdateStatus) (map[chainhash.Hash]store.UpdateStatus, error) { //nolint:unused
-	statusUpdatesMap := p.getStatusUpdateMap()
+	if shouldUpdateStatus(statusUpdate, *currentStatusUpdate) {
+		p.logger.Info("updateStatusMap - updating status", slog.String("hash", statusUpdate.Hash.String()), slog.String("old_status", currentStatusUpdate.Status.String()), slog.String("new_status", statusUpdate.Status.String()))
+		currentStatusUpdate.StatusHistory = append(currentStatusUpdate.StatusHistory, store.StatusWithTimestamp{
+			Status:    currentStatusUpdate.Status,
+			Timestamp: p.now(),
```
boecklim, line 47: I think this is not the correct place to set the timestamp. The `currentStatusUpdate` should already contain the timestamp from that moment when the original event happened.

The `PeerTxMessage` type has a field `Start` which we can use to capture the timestamp when the event happened and then update the history accordingly

internal/metamorph/processor.go
```diff
@@ -492,8 +526,19 @@ func (p *Processor) statusUpdateWithCallback(ctx context.Context, statusUpdates,
 		updatedData = append(updatedData, updatedDoubleSpendData...)
 	}
 
+	statusHistoryUpdates := filterUpdates(statusUpdates, updatedData)
+	shUpdatedData, err := p.store.UpdateStatusHistoryBulk(ctx, statusHistoryUpdates)
+	if err != nil {
+		return err
```
boecklim, line 532: I think if the update of status history fails, we shouldn't return an error, but just log it.

This feature just provides additional information for each transaction, but is not critical

pawellewandowski98: Done

internal/metamorph/processor_helpers.go
pawellewandowski98: Done

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -673,7 +695,129 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = tx.Exec(`SELECT * FROM metamorph.transactions WHERE hash in (SELECT UNNEST($1::BYTEA[])) ORDER BY hash FOR UPDATE`, pq.Array(txHashes))
+	if err != nil {
+		if rollBackErr := tx.Rollback(); rollBackErr != nil {
+			return nil, errors.Join(err, fmt.Errorf("failed to rollback: %v", rollBackErr))
+		}
+		return nil, err
```
kuba-4chain, line 817: Same as above ☝️

internal/metamorph/processor_helpers.go
```diff
@@ -163,3 +166,20 @@ func mergeUnique(arr1, arr2 []string) []string {
...
+				isProcessed = true
+				break
+			}
+		}
+		if !isProcessed {
+			result = append(result, unproc)
+		}
+	}
+	return result
+}
```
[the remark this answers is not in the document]
pawellewandowski98: Done

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -673,7 +695,129 @@ func (p *PostgreSQL) UpdateStatusBulk(ctx context.Context, updates []store.Updat
 		return nil, err
 	}
 
-	rows, err := tx.QueryContext(ctx, qBulk, p.now(), pq.Array(txHashes), pq.Array(statuses), pq.Array(rejectReasons))
+	rows, err := tx.QueryContext(ctx, qBulk, p.now(), pq.Array(txHashes), pq.Array(statuses), pq.Array(rejectReasons), pq.Array(statusHistories), pq.Array(timestamps))
+	if err != nil {
+		if rollBackErr := tx.Rollback(); rollBackErr != nil {
```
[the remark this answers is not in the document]
pawellewandowski98: Done

internal/metamorph/store/postgresql/postgres.go
pawellewandowski98: Done

PR #592, opened 2024-09-18
ARCO-173: add op return to broadcaster-cli transactions

pawellewandowski98:
## Description of Changes

Added new config variable and flag that defines text which will be added to OP_RETURN in broadcaster-cli transactions.

## Linked Issues / Tickets

ARCO-173

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


internal/broadcaster/rate_broadcaster.go
```diff
@@ -182,6 +188,8 @@ utxoLoop:
 
 			b.satoshiMap.Store(tx.TxID(), tx.Outputs[0].Satoshis)
 
+			b.logger.Info("tx submitted", slog.String("tx id", tx.TxID()))
```
boecklim, line 191: This function does not submit txs, it just creates them. But I think we should remove this log as it will create super much logs if you do performance testing using this tool

pawellewandowski98: Removed

internal/broadcaster/rate_broadcaster.go
```diff
@@ -173,7 +173,13 @@ utxoLoop:
 				return nil, fmt.Errorf("failed to pay transaction %d: %v", amount, err)
 			}
 
-			// Todo: Add OP_RETURN with text "ARC testing" so that WoC can tag it
+			if b.opReturn != "" {
+				err = tx.AddOpReturnOutput([]byte(b.opReturn))
+				if err != nil {
+					return nil, fmt.Errorf("failed to add OP_RETURN output: %v", err)
+				}
+				b.logger.Info("added OP_RETURN output")
```
boecklim, line 181: This creates too many logs. Please remove, but add on line 75
```go
b.logger.Info("Start broadcasting", slog.String("wait for status", b.waitForStatus.String()), slog.String("op return", b.opReturn))
```

cmd/broadcaster-cli/app/utxos/broadcast/broadcast.go
```diff
@@ -200,4 +206,10 @@ func init() {
 		log.Fatal(err)
 	}
 
+	Cmd.Flags().Int("opReturn", 0, "Text which will be added to an OP_RETURN output. If empty, no OP_RETURN output will be added")
```
boecklim, line 209: This flag needs to have type string
```suggestion
	Cmd.Flags().String("opReturn", "", "Text which will be added to an OP_RETURN output. If empty, no OP_RETURN output will be added")
```

internal/broadcaster/rate_broadcaster.go
pawellewandowski98: Done

cmd/broadcaster-cli/app/utxos/broadcast/broadcast.go
pawellewandowski98: Done

boecklim (approved):

PR #597, opened 2024-09-23
Test/utxo create

kameshraj23:
Test to cover utxo create

Screenshot 2024-09-23 at 20 36 35

### DIFF
--- a/internal/broadcaster/utxo_creator_test.go
+++ b/internal/broadcaster/utxo_creator_test.go
@@ -8,7 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestUTXOCreator_Start_WithMock(t *testing.T) {
+func TestUTXOCreator(t *testing.T) {
 	tt := []struct {
 		name               string
 		outputs            int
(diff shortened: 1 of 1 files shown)


internal/broadcaster/utxo_creator_test.go
```diff
@@ -0,0 +1,85 @@
...
+			// Verify if the error is as expected
+			if tc.expectedError != nil {
+				require.Error(t, err)
+				require.EqualError(t, err, tc.expectedError.Error())
+			} else {
+				require.NoError(t, err)
+			}
+
+			// Verify the number of times Start was called
+			require.Equal(t, tc.expectedStartCalls, len(mockedCreator.StartCalls()))
```
pawellewandowski98, line 81: use the GWT patterning that is now used in all tests

```suggestion
                        // given
			// Create the mock
			sut := &mocks.CreatorMock{
				StartFunc: tc.startFunc,
			}

                        // when
			// Call the Start method with mock
			err := sut(tc.outputs, tc.satoshisPerOutput)

                        // then
			// Verify if the error is as expected
			if tc.expectedError != nil {
				require.Error(t, err)
				require.EqualError(t, err, tc.expectedError.Error())
			} else {
				require.NoError(t, err)
			}

			// Verify the number of times Start was called
			require.Equal(t, tc.expectedStartCalls, len(mockedCreator.StartCalls()))
```

kuba-4chain: I would even remove the comments
```
// Create the mock
// Call the Start method with mock
// Verify if the error is as expected
```

It's all quite self-explanatory :)

kameshraj23: this is now addressed

PR #603, opened 2024-09-28
BPAAS-1649: add retry type of policy

shotasilagadzetaal:
I'm introducing retry policy. I didn't quite like [the idea of submitting the transaction with message queue and implementing that specific logic in case of error ](https://taaltech.atlassian.net/browse/BPAAS-1649)for the following reasons:

1. Submitting TX to message queue doesn't actually do anything different except it retries processing the request after some time (maybe immediately)
2. implementing the the same logic (checking if we have targeted status or request expired) sounds redundant while we have that logic already in `processTransaction`

In this current implementation I do exactly what message queue would do and even more. If we have successful response we return it immediately. If not, we will retry processing request until the timeout expires or we get non-nil response. On top of that, we are actually also checking status not just processing the tx.

With this approach we don't write the same logic here and we will retry a few times.

### DIFF
--- a/pkg/metamorph/client.go
+++ b/pkg/metamorph/client.go
@@ -196,20 +196,22 @@ func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction
 		}, nil
 	}
 
-	response, err := m.client.PutTransaction(ctx, request)
-	if err != nil {
-		m.logger.Warn("Failed to submit transaction", slog.String("hash", tx.TxID()), slog.String("key", err.Error()))
-		if m.mqClient != nil {
-			err := m.mqClient.PublishMarshal(SubmitTxTopic, request)
-			if err != nil {
-				return nil, err
-			}
-
-			return &TransactionStatus{
-				TxID:      tx.TxID(),
-				Status:    metamorph_api.Status_QUEUED.String(),
-				Timestamp: m.now().Unix(),
-			}, nil
+	var response *metamorph_api.TransactionStatus
+	var err error
+	// in case of error try PutTransaction until timeout expires
+	start := time.Now()
+	const interval = 300 * time.Millisecond
+	maxTimeout := time.Duration(5 * time.Second)
+	maxTimeout = max(time.Duration(request.MaxTimeout)*time.Second, maxTimeout)
+	for {
+		response, err = m.client.PutTransaction(ctx, request)
+		if err == nil {
+			break
+		}
+		m.logger.Error("Failed to put transactions", slog.String("err", err.Error()))
+		time.Sleep(interval)
+		if maxTimeout >= time.Since(start) {
+			continue
 		}
 
 		return nil, err
@@ -258,28 +260,23 @@ func (m *Metamorph) SubmitTransactions(ctx context.Context, txs sdkTx.Transactio
 	}
 
 	// put all transactions together
-	responses, err := m.client.PutTransactions(ctx, in)
-	if err != nil {
-		m.logger.Warn("Failed to submit transactions", slog.String("key", err.Error()))
-		if m.mqClient != nil {
-			for _, tx := range in.Transactions {
-				err := m.mqClient.PublishMarshal(SubmitTxTopic, tx)
-				if err != nil {
-					return nil, err
-				}
-			}
-
-			// parse response and return to user
-			ret := make([]*TransactionStatus, 0)
-			for _, tx := range txs {
-				ret = append(ret, &TransactionStatus{
-					TxID:      tx.TxID(),
-					Status:    metamorph_api.Status_QUEUED.String(),
-					Timestamp: m.now().Unix(),
-				})
-			}
-
-			return ret, nil
+	start := time.Now()
+	const interval = 300 * time.Millisecond
+	maxTimeout := time.Duration(5 * time.Second)
+	if len(in.Transactions) != 0 {
+		maxTimeout = time.Duration(in.Transactions[0].MaxTimeout) * time.Second
+	}
+	var responses *metamorph_api.TransactionStatuses
+	var err error
+	for {
+		responses, err = m.client.PutTransactions(ctx, in)
+		if err == nil {
+			break
+		}
+		m.logger.Error("Failed to put transactions", slog.String("err", err.Error()))
+		time.Sleep(interval)
+		if maxTimeout >= time.Since(start) {
+			continue
 		}
 
 		return nil, err
--- a/pkg/metamorph/client_test.go
+++ b/pkg/metamorph/client_test.go
@@ -121,40 +121,6 @@ func TestClient_SubmitTransaction(t *testing.T) {
 			expectedErrorStr: "failed to put tx",
 		},
 		{
-			name: "wait for received, put tx err, with mq client",
-			options: &metamorph.TransactionOptions{
-				WaitForStatus: metamorph_api.Status_RECEIVED,
-			},
-			putTxStatus: &metamorph_api.TransactionStatus{
-				Txid:   testdata.TX1Hash.String(),
-				Status: metamorph_api.Status_RECEIVED,
-			},
-			putTxErr:     errors.New("failed to put tx"),
-			withMqClient: true,
-
-			expectedStatus: &metamorph.TransactionStatus{
-				TxID:      testdata.TX1Hash.String(),
-				Status:    metamorph_api.Status_QUEUED.String(),
-				Timestamp: now.Unix(),
-			},
-		},
-		{
-			name: "wait for received, put tx err, with mq client, publish submit tx err",
-			options: &metamorph.TransactionOptions{
-				WaitForStatus: metamorph_api.Status_RECEIVED,
-			},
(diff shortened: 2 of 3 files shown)


pkg/metamorph/client.go
```diff
@@ -196,23 +196,38 @@ func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction
...
+					TxID:      tx.TxID(),
+					Status:    metamorph_api.Status_QUEUED.String(),
+					Timestamp: m.now().Unix(),
+				}, nil
+			}
+
+			return nil, err
+		} else {
+			break
+		}
```
boecklim, line 230: This `else` is not needed, I think

pkg/metamorph/client.go
```diff
@@ -196,23 +196,38 @@ func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction
...
-		m.logger.Warn("Failed to submit transaction", slog.String("hash", tx.TxID()), slog.String("key", err.Error()))
-		if m.mqClient != nil {
-			err := m.mqClient.PublishMarshal(SubmitTxTopic, request)
-			if err != nil {
-				return nil, err
+	var response *metamorph_api.TransactionStatus
+	var err error
+	// in case of error try PutTransaction until timeout expires
+	for {
+		start := time.Now()
```
boecklim, line 203: Doesn't this need to be defined outside of the for loop?

Also it should use the now function of the object `start := m.now()`

pkg/metamorph/client.go
```diff
@@ -196,23 +196,38 @@ func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction
...
+		response, err = m.client.PutTransaction(ctx, request)
+		duration := time.Since(start).Seconds() + 1
+		time.Sleep(300 * time.Millisecond)
+		if err != nil {
```
boecklim, line 207: The function could look a bit simpler if execution breaks in case of `err == nil`

I think this reflects essentially the same logic
```go
	start := m.now()
	const interval = 300 * time.Millisecond
	maxTimeout := time.Duration(request.MaxTimeout) * time.Second
	for {
		response, err = m.client.PutTransaction(ctx, request)
		if err == nil {
			break
		}
		m.logger.Warn("Failed to submit transaction", slog.String("hash", tx.TxID()), slog.String("key", err.Error()))

		time.Sleep(interval)

		if maxTimeout >= m.now().Sub(start) {
			continue
		}

		if m.mqClient != nil {
			err := m.mqClient.PublishMarshal(SubmitTxTopic, request)
			if err != nil {
				return nil, err
			}

			return &TransactionStatus{
				TxID:      tx.TxID(),
				Status:    metamorph_api.Status_QUEUED.String(),
				Timestamp: m.now().Unix(),
			}, nil
		}

		return nil, err
	}
```

pkg/metamorph/client_test.go
```diff
@@ -58,6 +58,33 @@ func TestClient_SetUnlockedByName(t *testing.T) {
 	}
 }
 
+func TestStatusWaiting(t *testing.T) {
```
boecklim, line 61: Would be nice if this test could be just another test case in `TestClient_SubmitTransaction`

pkg/metamorph/client.go
shotasilagadzetaal: No, we are measuring `PutTransaction` call duration on every iteration

pkg/metamorph/client.go
```diff
@@ -196,9 +196,24 @@ func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction
...
+	// in case of error try PutTransaction until timeout expires
+	start := time.Now()
+	const interval = 300 * time.Millisecond
+	maxTimeout := time.Duration(request.MaxTimeout) * time.Second
+	for {
+		response, err = m.client.PutTransaction(ctx, request)
+		if err == nil {
+			break
+		}
 		m.logger.Warn("Failed to submit transaction", slog.String("hash", tx.TxID()), slog.String("key", err.Error()))
```
boecklim, line 210: Unrelated but can you changes this?
```go
slog.String("err", err.Error())
```

pkg/metamorph/client.go
```diff
@@ -258,9 +273,25 @@ func (m *Metamorph) SubmitTransactions(ctx context.Context, txs sdkTx.Transactio
...
+		maxTimeout = time.Duration(in.Transactions[0].MaxTimeout) * time.Second
+	}
+	var responses *metamorph_api.TransactionStatuses
+	var err error
+	for {
+		responses, err = m.client.PutTransactions(ctx, in)
+		if err == nil {
+			break
+		}
 		m.logger.Warn("Failed to submit transactions", slog.String("key", err.Error()))
```
boecklim, line 289: Unrelated but can you changes this?
```go
slog.String("err", err.Error())
```

pkg/metamorph/client.go
```diff
@@ -196,9 +196,24 @@ func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction
...
+			break
+		}
 		m.logger.Warn("Failed to submit transaction", slog.String("hash", tx.TxID()), slog.String("key", err.Error()))
+
+		time.Sleep(interval)
+		if maxTimeout >= time.Since(start) {
+			continue
+		}
+
 		if m.mqClient != nil {
```
boecklim, line 217: Please remove this block

```go
		if m.mqClient != nil {
			err := m.mqClient.PublishMarshal(SubmitTxTopic, request)
			if err != nil {
				return nil, err
			}
			return &TransactionStatus{
				TxID:      tx.TxID(),
				Status:    metamorph_api.Status_QUEUED.String(),
				Timestamp: m.now().Unix(),
			}, nil
		}

```

pkg/metamorph/client.go
```diff
@@ -258,9 +273,25 @@ func (m *Metamorph) SubmitTransactions(ctx context.Context, txs sdkTx.Transactio
...
+		if err == nil {
+			break
+		}
 		m.logger.Warn("Failed to submit transactions", slog.String("key", err.Error()))
+		time.Sleep(interval)
+		if maxTimeout >= time.Since(start) {
+			continue
+		}
+
 		if m.mqClient != nil {
```
boecklim, line 295: Please remove this block

```go
if m.mqClient != nil {
			for _, tx := range in.Transactions {
				err := m.mqClient.PublishMarshal(SubmitTxTopic, tx)
				if err != nil {
					return nil, err
				}
			}
			// parse response and return to user
			ret := make([]*TransactionStatus, 0)
			for _, tx := range txs {
				ret = append(ret, &TransactionStatus{
					TxID:      tx.TxID(),
					Status:    metamorph_api.Status_QUEUED.String(),
					Timestamp: m.now().Unix(),
				})
			}
			return ret, nil
		}
```

pkg/metamorph/client.go
```diff
@@ -196,20 +196,21 @@ func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction
...
+	// in case of error try PutTransaction until timeout expires
+	start := time.Now()
+	const interval = 300 * time.Millisecond
+	maxTimeout := time.Duration(request.MaxTimeout) * time.Second
+	for {
+		response, err = m.client.PutTransaction(ctx, request)
+		if err == nil {
+			break
+		}
+		slog.String("err", err.Error())
```
boecklim, line 210: That is not going to log anything

This would be needed

```go
m.logger.Error("Failed to put transaction", slog.String("err", err.Error()))
```

pkg/metamorph/client.go
```diff
@@ -258,28 +259,23 @@ func (m *Metamorph) SubmitTransactions(ctx context.Context, txs sdkTx.Transactio
...
+		maxTimeout = time.Duration(in.Transactions[0].MaxTimeout) * time.Second
+	}
+	var responses *metamorph_api.TransactionStatuses
+	var err error
+	for {
+		responses, err = m.client.PutTransactions(ctx, in)
+		if err == nil {
+			break
+		}
+		slog.String("err", err.Error())
```
boecklim, line 275: That is not going to log anything

This would be needed

```go
m.logger.Error("Failed to put transactions", slog.String("err", err.Error()))
```

pkg/metamorph/client.go
```diff
@@ -258,28 +259,23 @@ func (m *Metamorph) SubmitTransactions(ctx context.Context, txs sdkTx.Transactio
...
-					TxID:      tx.TxID(),
-					Status:    metamorph_api.Status_QUEUED.String(),
-					Timestamp: m.now().Unix(),
-				})
-			}
-
-			return ret, nil
+	start := time.Now()
+	const interval = 300 * time.Millisecond
+	maxTimeout := time.Duration(0)
```
boecklim, line 264: I think it may be better to set this to the default of 5s

pkg/metamorph/client.go
boecklim: Or maybe better `Warn` than error 🤔

boecklim (approved):

PR #611, opened 2024-10-16
feat(ARCO-115): Reorg Support

kuba-4chain:
## Description of Changes

1. Processing of blocks received from peers was improved to allow for:
    - proper handling of ORPHANED blocks - updating them backwards with each new incoming block that extends the orphaned chain and publishing their transactions if they become part of the longest chain
    - secure handling of concurrent blocks being processed - by updating the `UpsertBlock` query with additional checks
2. Database method `GetBlockGaps` was updated and simplified.
4. The sum of chainworks of competing forks is now compared to find out the longest chain, instead of comparing the chainwork of only the tips of forks.
5. A great number of fixtures and unit tests were added to ensure maximum test coverage - as this is a rare scenario, difficult to spot and debug in production.
6. E2E tests were improved, by decreasing the fillGapsInterval to 1 second - to make sure there are no gaps when the tests begin.
7. A new transaction status `MINED_IN_STALE_BLOCK` was introduced in metamorph api.
5. The way of publishing mined transactions to metamorph was updated
    - transactions from incoming block, that's extending the longest chain, are published to metamorph as `MINED` - no change here
    - transactions from incoming block, that's extending a stale chain, that are not found in the longest chain are in `SEEN_ON_NETWORK` state (ignored) until they are found in the longest chain (which eventually should happen, because nodes put those transactions back in their mempool)
    - in case of **reorg**, transactions from the entire stale chain, that is now becoming the longest chain, are published
        - transactions that were previously found in the longest chain that is now becoming a stale chain are published with updated block data (block hash, block height and merkle path)
        - transactions that were not found in the longest chain are published as newly mined
    - in case of **reorg**, transactions from the previously longest chain that were **NOT** found in the new longest chain are published as `MINED_IN_STALE_BLOCK` - this is to inform users that their transaction is no longer mined in the longest chain, so they have to wait until it's `MINED` again
8. An integration test for the entire reorg process, including updating transactions, was written, covering all of the above-mentioned scenarios and some more edge cases - for single-instance processing.
9. Small refactor of structures
    - `PeerTxMessage` -> `TxStatusMessage`
    - `GetMinedTransactionsResult` -> `TransactionBlock`
    - New structure `chain` -> slice of blocks

## Flow diagram of the updated `processBlock` function in blocktx processor

ReorgSupport drawio

## Linked Issues / Tickets

[ARCO-178](https://bitcoinassociation.atlassian.net/browse/ARCO-178)
[ARCO-179](https://bitcoinassociation.atlassian.net/browse/ARCO-179)
[ARCO-199](https://bitcoinassociation.atlassian.net/browse/ARCO-199)

## Testing Procedure

- [x] All tests pass locally

## Checklist:


internal/blocktx/store/postgresql/get_orphaned_chain.go
```diff
@@ -0,0 +1,64 @@
+package postgresql
+
+import (
+	"context"
+
+	"github.com/bitcoin-sv/arc/internal/blocktx/blocktx_api"
+)
+
+// GetOrphanedChainUpFromHash is a function that recursively searches for blocks marked
```
boecklim, line 9: nit: ~~is a function that~~

internal/blocktx/store/postgresql/migrations/000019_add_is_longest.up.sql
```diff
@@ -0,0 +1,13 @@
+-- field `is_longest` is an implementation detail that will help to
+-- make sure that there is only one longest chain at any given height
+-- and is also used as a helper when querying for longest chain
+ALTER TABLE blocktx.blocks
+ADD COLUMN is_longest BOOLEAN NOT NULL DEFAULT TRUE;
```
boecklim, line 5: Why is there the need for a field `is_longest` and `status`?
I think `status=10` means longest, so isn't `is_longest` redundant?

internal/blocktx/processor.go
```diff
@@ -386,6 +407,7 @@ func (p *Processor) publishMinedTxs(txHashes []*chainhash.Hash) error {
 			BlockHash:       minedTx.BlockHash,
 			BlockHeight:     minedTx.BlockHeight,
 			MerklePath:      minedTx.MerklePath,
+			BlockStatus:     minedTx.BlockStatus,
 		}
 		err = p.mqClient.PublishMarshal(MinedTxsTopic, txBlock)
```
boecklim, line 412: That is unrelated to the PR, but here we only return the err if it happens in the last iteration.
I think it should be at each iteration

```go
	for _, minedTx := range minedTxs {
		txBlock := &blocktx_api.TransactionBlock{
			TransactionHash: minedTx.TxHash,
			BlockHash:       minedTx.BlockHash,
			BlockHeight:     minedTx.BlockHeight,
			MerklePath:      minedTx.MerklePath,
			BlockStatus:     minedTx.BlockStatus,
		}
		err = p.mqClient.PublishMarshal(MinedTxsTopic, txBlock)
	        if err != nil {
		    return fmt.Errorf("failed to publish mined transactions: %v", err)
	        }
       }
```

internal/blocktx/store/postgresql/postgres.go
```diff
@@ -66,14 +82,65 @@ func New(dbInfo string, idleConns int, maxOpenConns int, opts ...func(postgreSQL
...
+		maxPostgresBulkInsertRows: maxPostgresBulkInsertRows,
+		tracingEnabled:            p.tracingEnabled,
+		tracingAttributes:         p.tracingAttributes,
+	}
+
+	return cloneStore, nil
+}
+
+// UnitOfWork methods below
+func (p *PostgreSQL) Commit() error {
```
boecklim, line 118: Add one new line between function and comment as the comment does not describe the function

internal/blocktx/store/postgresql/postgres_test.go
```diff
@@ -309,36 +373,149 @@ func TestPostgresDB(t *testing.T) {
 		stale4, err := postgresDB.GetBlock(ctx, hash4Stale)
 		require.NoError(t, err)
 		require.Equal(t, blocktx_api.Status_LONGEST, stale4.Status)
+
+		// when
+		err = postgresDB.UpdateBlocksStatuses(ctx, blockStatusUpdatesViolating)
+		require.True(t, errors.Is(err, store.ErrFailedToUpdateBlockStatuses))
```
boecklim, line 379: nit: You could use here `require.ErrorIs(t, actualErr, tc.expectedError)`

internal/blocktx/store/postgresql/postgres_test.go
```diff
@@ -309,36 +373,149 @@ func TestPostgresDB(t *testing.T) {
...
+			},
+		}
+
+		// when
+		actualTxs, err := postgresDB.GetRegisteredTxsByBlockHashes(ctx, blockHashes)
+
+		// then
+		require.NoError(t, err)
+		require.Equal(t, expectedTxs, actualTxs)
 	})
```
boecklim, line 515: This seems to be the exact same test case as the previous one "get registered txs".
I think one of them could be removed

internal/blocktx/store/postgresql/postgres_test.go
```diff
@@ -474,6 +651,24 @@ func TestPostgresDB(t *testing.T) {
...
+
+		err = uow.WriteLockBlocksTable(context.Background())
+		require.NoError(t, err)
+
+		err = uow.Rollback()
+		require.NoError(t, err)
+
+		err = uow.Commit()
+		require.Equal(t, ErrNoTransaction, err)
+	})
```
boecklim, line 671: Here we could have some additional checks to ensure the correct behaviour of commits/rollbacks etc

```go
t.Run("lock blocks table", func(t *testing.T) {
		prepareDb(t, postgresDB, "")

		err := postgresDB.WriteLockBlocksTable(context.Background())
		require.Error(t, err)
		require.Equal(t, ErrNoTransaction, err)

		uow, err := postgresDB.StartUnitOfWork(context.Background())
		require.NoError(t, err)

		err = uow.WriteLockBlocksTable(context.Background())
		require.NoError(t, err)

		err = uow.Rollback()
		require.NoError(t, err)

		err = uow.Commit()
		require.Equal(t, ErrNoTransaction, err)

		// ensure writing is possible after commit
		blockHash := testutils.RevChainhash(t, "000000000000000001b8adefc1eb98896c80e30e517b9e2655f1f929d9958a48")
		expectedBlock := &blocktx_api.Block{
			Hash:   blockHash[:],
			Height: 100,
			Status: blocktx_api.Status_LONGEST,
		}

		_, err = postgresDB.UpsertBlock(ctx, expectedBlock)
		require.NoError(t, err)

		// ensure commit persists a change
		blockHash2 := testutils.RevChainhash(t, "0000000000000000064cbaac5cedf71a5447771573ba585501952c023873817b")
		expectedBlock2 := &blocktx_api.Block{
			Hash:   blockHash2[:],
			Height: 101,
			Status: blocktx_api.Status_LONGEST,
		}
		uow, err = postgresDB.StartUnitOfWork(context.Background())
		require.NoError(t, err)

		err = uow.WriteLockBlocksTable(context.Background())
		require.NoError(t, err)

		_, err = uow.UpsertBlock(ctx, expectedBlock2)
		require.NoError(t, err)

		err = uow.Commit()
		require.NoError(t, err)

		block, err := postgresDB.GetBlock(ctx, blockHash)
		require.NoError(t, err)
		require.Equal(t, expectedBlock.Hash, block.Hash)

		// ensure rollback does roll back a transaction
		uow, err = postgresDB.StartUnitOfWork(context.Background())
		require.NoError(t, err)

		err = uow.WriteLockBlocksTable(context.Background())
		require.NoError(t, err)

		blockHash3 := testutils.RevChainhash(t, "0000000000000000025855b62f4c2e3732dad363a6f2ead94e4657ef96877067")
		expectedBlock3 := &blocktx_api.Block{
			Hash:   blockHash3[:],
			Height: 102,
			Status: blocktx_api.Status_LONGEST,
		}

		_, err = uow.UpsertBlock(ctx, expectedBlock3)
		require.NoError(t, err)

		err = uow.Rollback()
		require.NoError(t, err)

		_, err = postgresDB.GetBlock(ctx, blockHash3)
		require.ErrorIs(t, err, store.ErrBlockNotFound)
	})
```

boecklim: I also tried to ensure writing to the locked table fails, but that is blocking infinitely.

Tried to check if it is possible to ensure `UpsertBlock` is not blocking in that case but returning an error, but haven't found out how.

internal/blocktx/processor.go
```diff
@@ -563,65 +635,73 @@ func (p *Processor) competingChainsExist(ctx context.Context, block *blocktx_api
...
 	}
 
-	err = p.store.UpdateBlocksStatuses(ctx, blockStatusUpdates)
-	return err
+	return nil
 }
 
-func (p *Processor) markTransactionsAsMined(ctx context.Context, blockID uint64, merkleTree []*chainhash.Hash, blockHeight uint64, blockhash *chainhash.Hash) error {
+func (p *Processor) storeTransactions(ctx context.Context, blockID uint64, block *blocktx_api.Block, merkleTree []*chainhash.Hash) error {
 	ctx, span := tracing.StartTracing(ctx, "markTransactionsAsMined", p.tracingEnabled, p.tracingAttributes...)
```
boecklim, line 695: Rename the trace to `storeTransactions`

internal/blocktx/processor.go
```diff
@@ -658,56 +738,205 @@ func (p *Processor) markTransactionsAsMined(ctx context.Context, blockID uint64,
...
-		txBlock := &blocktx_api.TransactionBlock{
-			TransactionHash: updResp.Hash[:],
-			BlockHash:       blockhash[:],
-			BlockHeight:     blockHeight,
-			MerklePath:      updResp.MerklePath,
+	return nil
+}
+
+func (p *Processor) updateOrphans(ctx context.Context, incomingBlock *blocktx_api.Block, competing bool) (chain, bool, error) {
+	chain := []*blocktx_api.Block{incomingBlock}
```
boecklim, line 768: I think a trace `updateOrphans` could be useful

internal/blocktx/processor.go
```diff
@@ -658,56 +738,205 @@ func (p *Processor) markTransactionsAsMined(ctx context.Context, blockID uint64,
...
+	chain = append(chain, orphanedBlocks...)
+
+	// if we found any orphans and marked them as STALE
+	// we need to find out if they are part of the longest
+	// or stale chain, so competing is returned as true
+	return chain, true, nil
+}
+
+func (p *Processor) performReorg(ctx context.Context, staleChainTip *blocktx_api.Block) ([]store.TransactionBlock, error) {
+	uow, err := p.store.StartUnitOfWork(ctx)
```
boecklim, line 831: I think a trace `performReorg` could be useful

internal/blocktx/integration_test/reorg_integration_test.go
```diff
@@ -1,20 +1,32 @@
...
 // 		2. A hardcoded msg with competing block at height 822015 is being sent through the mocked PeerHandler
 // 		3. This block has a chainwork lower than the current tip of chain - becomes STALE
-// 		4. Next competing block, at height 822016 is being send through the mocked PeerHandler
-// 		5. This block has a greater chainwork than the current tip of longest chain - it becomes LONGEST despite not being the highest
-//
-// Todo: Next tasks:
-// 		- Verify if reorg was performed correctly, if previous blocks have updated statuses
-// 		- Include metamorph in this test and verify that transactions statuses are properly updated
+// 		4. Registered transactions from this block that are not in the longest chain are published to metamorph message queue with blockstatus = STALE
+// 		5. Next competing block, at height 822016 is being send through the mocked PeerHandler
```
boecklim, line 14: nid: ... is being **sent** through...

internal/blocktx/integration_test/reorg_integration_test.go
```diff
@@ -1,20 +1,32 @@
...
+// 		6. This block has a greater chainwork than the current tip of longest chain - it becomes LONGEST despite not being the highest
+// 		7. Verification of reorg - checking if statuses are correctly switched
+// 		8. Verification of transactions
+// 			- transactions from the stale chain becoming the longest are published
+// 			- transactions that were previously in the longest chain are published with udpated block data
```
boecklim, line 19: nit: ... with **updated** block ...

internal/blocktx/store/postgresql/postgres.go
kuba-4chain: Unit of Work was removed

internal/blocktx/store/postgresql/postgres_test.go
kuba-4chain: Right, I fixed it in other places as well

internal/blocktx/store/postgresql/postgres_test.go
kuba-4chain: You're right, the first one should be removed, thanks for spotting that!

internal/blocktx/store/postgresql/postgres_test.go
kuba-4chain: > if it is possible to ensure UpsertBlock is not blocking in that case but returning an error

Yes, it's absolutely possible, but I made this blocking on purpose, so that if 2 reorgs would happen at the same  time (trying to acquire the same lock), one would wait for the other to finish and then execute. I didn't want it to throw an error.

If you want it to throw an error when 2 instances try to access the lock:
`LOCK TABLE blocktx IN SHARE MODE`

And with blocking / no error:
`LOCK TABLE blocktx IN EXCLUSIVE MODE`


Anyways, the locking mechanism is now removed, as after going through all the edge cases, it turned out it's not necessary. So I removed it completely.

internal/blocktx/processor.go
kuba-4chain: Good catch, fixed

internal/blocktx/processor.go
kuba-4chain: Right, added tracing in multiple places

internal/blocktx/processor.go
kuba-4chain: I think we don't return here on purpose:
1. If there's a problem sending one transaction, we still probably want to send the rest of them
2. If we return an error from this method, we would considered the block unprocessed, when in reality it's processed, but there's only a problem with transaction.

I think it's okay the way it is, but let me know if I'm not missing something.

internal/blocktx/processor.go
```diff
@@ -209,16 +211,31 @@ func (p *Processor) StartBlockProcessing() {
...
+				storeErr := p.store.MarkBlockAsDone(p.ctx, &blockHash, blockMsg.Size, uint64(len(blockMsg.TransactionHashes)))
+				if storeErr != nil {
+					p.logger.Error("unable to mark block as processed", slog.String("hash", blockHash.String()), slog.String("err", storeErr.Error()))
+					p.unlockBlock(p.ctx, &blockHash)
+					p.stopBlockProcessGuard(&blockHash) // release guardian
+					continue
 				}
+
+				// add the total block processing time to the stats
+				p.logger.Info("Processed block", slog.String("hash", blockHash.String()), slog.Int("txs", len(blockMsg.TransactionHashes)), slog.String("duration", time.Since(timeStart).String()))
```
boecklim, line 237: Nice! 👍

internal/blocktx/processor.go
```diff
@@ -426,206 +449,241 @@ func (p *Processor) processBlock(msg *p2p.BlockMessage) (err error) {
...
 
-	return true, nil
-}
+	if prevBlock == nil {
+		// This check is only in case there's a fresh, empty database
+		// with no blocks, to mark the first block as the LONGEST chain
+		var longestTipExists bool
+		longestTipExists, err = p.longestTipExists(ctx)
+		if err != nil {
+			p.logger.Error("unable to verify the longest tip existance in db", slog.String("hash", getHashStringNoErr(block.Hash)), slog.Uint64("height", block.Height), slog.String("err", err.Error()))
```
boecklim, line 557: nit: exist**e**nce

boecklim (approved):

internal/blocktx/processor.go
kuba-4chain: fixed

arkadiuszos4chain (approved):

PR #613, opened 2024-10-17
ARCO-220: fix block handling 

arkadiuszos4chain:
## Description of Changes

Fixing two critical issues related to transaction and block processing:
* never mined transactions – issue where transactions were stuck in the SEEN_ON_NETWORK status but were never mined
* never processed missing block – issue where logs showing "Requesting missing block" would continue indefinitely, but the block was never processed

Changes made:
* increased the limit for queued blocks to be processed by the blocktx instance
* fixed error handling for block processing failures
* moved the periodic check for gaps in the blockchain from the processor to a background worker

## Linked Issues / Tickets

ARCO-220

## Testing Procedure

- [x] All tests pass locally

## Checklist:


internal/blocktx/background_workers_test.go
```diff
@@ -0,0 +1,109 @@
...
+	hostname, err := os.Hostname()
+	require.NoError(t, err)
+
+	tt := []struct {
+		name            string
+		hostname        string
+		getBlockGapsErr error
+		blockGaps       []*store.BlockGap
+
+		minExpectedGetBlockCapsCalls int
```
kuba-4chain, line 30:
```suggestion
		minExpectedGetBlockGapsCalls int
```

internal/blocktx/peer_handler.go
```diff
@@ -107,18 +109,37 @@ func (ph *PeerHandler) HandleBlockAnnouncement(msg *wire.InvVect, peer p2p.PeerI
...
-		ph.logger.Debug(ErrUnableToCastWireMessage.Error())
 		return ErrUnableToCastWireMessage
 	}
 
 	ph.blockProcessCh <- msg
-
 	return nil
 }
+
+func extractHeightFromCoinbaseTx(tx *sdkTx.Transaction) uint64 {
```
kuba-4chain, line 125: This function has 2 small tests in `processor_helpers_test.go`. We can leave them there or create a separate file `peer_handler_test.go`, but just letting you know

internal/blocktx/peer_handler.go
```diff
@@ -107,18 +109,37 @@ func (ph *PeerHandler) HandleBlockAnnouncement(msg *wire.InvVect, peer p2p.PeerI
...
+	// Coinbase tx has a special format, the height is encoded in the first 4 bytes of the scriptSig (BIP-34)
+
+	cscript := *(tx.Inputs[0].UnlockingScript)
+
+	if cscript[0] >= script.Op1 && cscript[0] <= script.Op16 { // first 16 blocks are treated differently (we have to handle it due to our tests)
+		return uint64(cscript[0] - 0x50)
+	}
```
kuba-4chain, line 132: Would be nice to have a small single test for this case, e.g. returning height of block 9

https://whatsonchain.com/block-height/9

internal/blocktx/processor_helpers.go
```diff
@@ -140,3 +121,29 @@ func progressIndices(total, steps int) map[int]int {
 	progress[total] = 100
 	return progress
 }
+
+func withRetry(ctx context.Context, fn func() error, n int) error {
+	n = int(math.Max(float64(n), 1)) // ensure n is at least 1
```
kuba-4chain, line 126:
```suggestion
	n = max(n, 1) // ensure n is at least 1
```

internal/blocktx/processor.go
```diff
@@ -188,45 +187,56 @@ func (p *Processor) StartBlockProcessing() {
...
-func (p *Processor) StartFillGaps(peers []p2p.PeerI) {
+// unlock block for future processing
+func (p *Processor) unlockBlock(ctx context.Context, hash *chainhash.Hash) {
+	// use closures for retries
+	unlockFn := func() error {
+		_, err := p.store.DelBlockProcessing(ctx, hash, p.hostname)
+		return err
+	}
+
+	if unlockErr := withRetry(ctx, unlockFn, 5); unlockErr != nil {
```
kuba-4chain, line 212: Let's put that `5` to the consts on top, maybe `unlockBlockRetries` or something

internal/blocktx/processor.go
```diff
@@ -188,45 +187,56 @@ func (p *Processor) StartBlockProcessing() {
...
+		p.logger.ErrorContext(ctx, "failed to delete block processing", slog.String("hash", hash.String()), slog.String("err", unlockErr.Error()))
+	}
+}
+
+func (p *Processor) startBlockProcessedGuard(ctx context.Context, hash *chainhash.Hash) {
```
kuba-4chain, line 217: This is a really nice feature, but:
1. In case of an error in block processing, before that block is upserted into db, the block will be unlocked in the `processBlock` function, while this function will do 6 queries to database which will try to unlock an already unlocked block.
2. When we process a block, get an error when processing, unlock the block and then `fillGaps` will get us that block again, this goroutine may unlock that block even tho it's being processed (you found it 😄 )

For case 1. I recommend removing throwing an error when `rowsAffected != 1` in `DelBlockProcessing`. It will then stop trying to unlock an already unlocked block after the first try.

kuba-4chain (changes requested):

internal/blocktx/peer_handler.go
arkadiuszos4chain: 1. It was added because tests only
2. We can see that it works (or not) looking at test logs
3. Actual block with height 9 is in version 1 so the code won't work for it :) 

I see no reason to add more code to code base

internal/blocktx/peer_handler.go
arkadiuszos4chain: Moved to the new file

internal/blocktx/background_workers.go
```diff
@@ -0,0 +1,104 @@
...
+	go func() {
+		defer w.workersWg.Done()
+
+		t := time.NewTicker(interval)
+		peerIndex := 0
+
+		for {
+			select {
+			case <-t.C:
+				if peerIndex >= len(peers) {
```
shotasilagadzetaal, line 51: you could simply do `peerIndex%len(peers)` in fillGaps call

internal/blocktx/processor_helpers.go
```diff
@@ -140,3 +121,29 @@ func progressIndices(total, steps int) map[int]int {
...
+
+			// return if the function completes successfully
+			if err == nil {
+				return nil
+			}
+
+			rerr = errors.Join(rerr, fmt.Errorf("failure on %d try: %w", i+1, err))
+			time.Sleep(100 * time.Millisecond)
+		}
+	}
```
boecklim, line 146: This could be done similarly as the other retry function

https://github.com/bitcoin-sv/arc/blob/2b723ddcdab35fb753041ae700177376609f0f3e/internal/test_utils/utils.go#L57:L70

But this is also fine

internal/blocktx/peer_handler.go
boecklim, line 132: @arkadiuszos4chain 
Function `extractHeightFromCoinbaseTx()` is the same as before except this change, right?
What is the context of this change?

boecklim: Ah nice. I always wondered where this issue came from. Nice  👍 👍

internal/blocktx/processor.go
```diff
@@ -188,45 +187,56 @@ func (p *Processor) StartBlockProcessing() {
...
-				}
 
-				err := p.fillGaps(peers[peerIndex])
-				if err != nil {
-					p.logger.Error("failed to fill gaps", slog.String("error", err.Error()))
-				}
+		// respect processor shutdown while waiting
+		select {
+		case <-p.ctx.Done():
+			return // we can do nothing here
```
boecklim, line 226: Shouldn't we also run `p.unlockBlock(ctx, hash)` before exiting the function in case that the go routine stops due to a cancelled context? 🤔

arkadiuszos4chain: I made some changes to the logic and added more descriptive comments.

In summary, this goroutine will stop (via its context) when block processing is complete or when the processor shuts down, so we don't want to unlock the blocks here. I moved the unlocking of all blocks in progress to the Shutdown function.

internal/blocktx/processor.go
arkadiuszos4chain: Changed! 
Now this guardian goroutine completes when block processing finishes. Also, `unlock` no longer treats `ErrBlockNotFound` as a reason to retry

arkadiuszos4chain:
https://github.com/bitcoin-sv/arc/pull/620 <- proposition I mentioned on daily

internal/blocktx/background_workers.go
```diff
@@ -0,0 +1,100 @@
...
+	go func() {
+		defer w.workersWg.Done()
+
+		t := time.NewTicker(interval)
+		i := 0
+
+		for {
+			select {
+			case <-t.C:
+				err := w.fillGaps(peers[i%len(peers)], retentionDays, blockRequestCh)
```
kuba-4chain, line 55: Maybe this would be better so we don't overflow the `i`? 
```suggestion
				i = i%len(peers)
				err := w.fillGaps(peers[i], retentionDays, blockRequestCh)
```

internal/blocktx/processor.go
```diff
@@ -55,7 +56,8 @@ type Processor struct {
 	registerRequestTxsInterval  time.Duration
 	registerTxsBatchSize        int
 	registerRequestTxsBatchSize int
-	fillGapsInterval            time.Duration
+
+	processGruadsMap sync.Map
```
kuba-4chain, line 60:
```suggestion
	processGuardsMap sync.Map
```

internal/blocktx/processor.go
```diff
@@ -188,47 +190,81 @@ func (p *Processor) StartBlockProcessing() {
...
+			block, _ := p.store.GetBlock(execCtx, hash)
 
-				peerIndex++
-				ticker.Reset(p.fillGapsInterval)
+			if block != nil && block.Processed {
+				return // success
 			}
+
+			p.logger.Error(fmt.Sprintf("block was not processed after %v. Unlock the block to be processed later", waitForBlockProcessing), slog.String("hash", hash.String()))
+			p.unlockBlock(execCtx, hash)
```
kuba-4chain, line 233: We should also add here
```go
p.processGuardsMap.Delete(*hash)
```

kuba-4chain (changes requested):

internal/blocktx/background_workers_test.go
kuba-4chain: I still see `Caps` 😄

kuba-4chain (approved):

internal/callbacker/background_workers.go
```diff
@@ -43,8 +43,12 @@ func (w *BackgroundWorkers) StartQuarantineCallbacksDispatch(interval time.Durat
 }
 
 func (w *BackgroundWorkers) GracefulStop() {
+	w.l.Info("Shutting down")
+
 	w.cancelAll()
 	w.workersWg.Wait()
+
+	w.l.Info("Shutted down")
```
boecklim, line 51: Shutted down sounds a bit strange.

"Shutdown complete" ? 🤔

boecklim (approved):

PR #615, opened 2024-10-22
ARCO-223: add tracing to api and metamorph

pawellewandowski98:
## Description of Changes

Added tracing to API and metamorph:
- added tracer to arc handler and metamorph client
- added global tracer to metamorph package and used it in server and processor methods, created one tracer object which is used in grpc call and whole metamorph

## Linked Issues / Tickets

#ARCO-223

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


test/docker-compose.yaml
```diff
@@ -238,12 +244,24 @@ services:
       - "9090"
       - "2112"
     command: [ "./arc", "-api=true", "-config=." ]
+    environment:
+      - ARC_TRACING_DIALADDR
```
boecklim, line 248: How does this env var work? There is no value
Don't you have to add the setting in `/test/config/config.yaml`?

pkg/metamorph/client.go
```diff
@@ -89,7 +98,8 @@ func NewClient(client metamorph_api.MetaMorphAPIClient, opts ...func(client *Met
 }
 
 func DialGRPC(address string, prometheusEndpoint string, grpcMessageSize int) (*grpc.ClientConn, error) {
-	dialOpts, err := grpc_opts.GetGRPCClientOpts(prometheusEndpoint, grpcMessageSize)
+	tracer := otel.GetTracerProvider().Tracer("Metamorph-Client")
+	dialOpts, err := grpc_opts.GetGRPCClientOpts(prometheusEndpoint, grpcMessageSize, tracer)
```
kuba-4chain, line 102: So here we will always pass this tracer, even if it's disabled?

cmd/arc/services/api.go
```diff
@@ -49,15 +49,24 @@ func StartAPIServer(logger *slog.Logger, arcConfig *config.ArcConfig) (func(), e
 		return nil, fmt.Errorf("failed to connect to metamorph server: %v", err)
 	}
 
+	mtmOpts := []func(*metamorph.Metamorph){
+		metamorph.WithMqClient(mqClient),
+		metamorph.WithLogger(logger),
+	}
+
+	tracingEnabled := arcConfig.Tracing != nil
```
kuba-4chain, line 57:
```suggestion
	tracingEnabled := arcConfig.Tracing != nil && arcConfig.Tracing.DialAddr != ""
```

test/docker-compose.yaml
kuba-4chain: This env var is to able the usage of env when running docker compose in Makefile:
`ARC_TRACING_DIALADDR=http://arc-jaeger:4317 docker compose -f test/docker-compose.yaml up --build ...
`

If we add it to `/test/config/config.yaml` then we will always have tracing when running the tests locally. With this solution we have 2 separate commands in the Makefile
`run_e2e_tests`
and
`run_e2e_tests_with_tracing`

boecklim: Ah, understand. If it is in config then tracing will be always enabled instead of optional

pkg/metamorph/client.go
```diff
@@ -181,6 +211,11 @@ func (m *Metamorph) Health(ctx context.Context) error {
 
 // SubmitTransaction submits a transaction to the bitcoin network and returns the transaction in raw format.
 func (m *Metamorph) SubmitTransaction(ctx context.Context, tx *sdkTx.Transaction, options *TransactionOptions) (*TransactionStatus, error) {
+	if m.tracer != nil {
+		_, span := m.tracer.Start(ctx, "SubmitTransaction")
```
boecklim, line 215: I think here the context form tracer has to be returned so that it is used in line 242, not?

```go
response, err = m.client.PutTransaction(ctx, request)
```

boecklim:
Is replaced by https://github.com/bitcoin-sv/arc/pull/626

PR #624, opened 2024-10-28
new config for ipv6 related changes

shotasilagadzetaal:
## Description of Changes

New config parameters for ipv6 related changes

## Linked Issues / Tickets

Closes #ARCO-229

## Testing Procedure

I have  added test case to check if all the values are correctly read from config

## Checklist:


config/defaults.go
```diff
@@ -47,25 +47,31 @@ func getDefaultPeerRpcConfig() *PeerRpcConfig {
...
 }
 
-func getPeersConfig() []*PeerConfig {
-	return []*PeerConfig{
-		{
-			Host: "localhost",
-			Port: &PeerPortConfig{
-				P2P: 18333,
-				ZMQ: 28332,
+func getPeersConfig() *BroadcastingConfig {
```
kuba-4chain, line 50: Rename to `getBroadcastingConfig`
```suggestion
func getBroadcastingConfig() *BroadcastingConfig {
```

test/config/config.yaml
```diff
@@ -31,6 +31,27 @@ peers:
       p2p: 18333
       zmq: 28332
 
+broadcasting:
+  mode: unicast
```
kuba-4chain, line 35: Maybe we should remove the `peers` config above the `broadcasting` config?

config/config.go
```diff
@@ -31,6 +31,19 @@ type ArcConfig struct {
 	Cache              *CacheConfig        `mapstructure:"cache"`
 }
 
+type BroadcastingConfig struct {
+	Mode           string        `mapstructure:"mode"`
+	Ipv6Enabled    bool          `mapstructure:"ipv6Enabled"`
+	MulticastGroup string        `mapstructure:"multicastGroup"`
```
kuba-4chain, line 37: I'm not an expert on IPv6, so it's just a question out of curiosity, what is this multicastGroup for? Is it the group that Arc is a part of, or is it the group that Arc will be broadcasting to?

And is it enough to always have only one group?

kuba-4chain:
Nice work. I just have a question, because maybe I'm missing something, this config is a preparation for further changes to implement the IPv6 communications, but for now it's not yet being used anywhere, right?

arkadiuszos4chain:
I see two potential approaches:

### 1. Use P2P for Multicasting
_Not my preferred option, but workable._

- Update the `PeerConfig` structure as follows:
  ```yaml
  Peer:
    mode: unicast | multicast
    address | host: url
    port: number

- Explanation:
  - if `mode` is `unicast`, the peer operates as it currently does, with no additional changes
  - if `mode` is `multicast`, the url `must` be an IPv6 multicast group address
  
### 2.  Implement a Multicast Handler (that works alongside P2P)
- This handler would have similar logic to the peer but be simpler (no reconnections, no ping-pong messages, no handshakes).

- Add a new configuration key:
  ```yaml
  broadcast | network-communication:
    Peers: []           # (optional) same as current setup
    Multicast:
      group: url
      port: number      # (optional)
      
- Currently, it is unclear how groups would function in the network. Possible options include:
    - using a single multicast group with different ports to handle distinct topics
    - using a separate multicast group per topic
  therefore, the final schema of `Multicast` could change.
    
I believe that `go-p2p` is diagnostic regarding IP version, so there is no need to introduce the IPv6enabled flag.

test/config/config.yaml
boecklim: Yes, `peers` can be removed

test/config/config.yaml
```diff
@@ -31,6 +31,27 @@ peers:
...
+  ipv6Enabled: true
+  multicastGroup: 172.28.56.77
```
boecklim, line 37: Thinking now maybe it would be better to have a distinct multicast object or unicast object like this 🤔 

Also comments should be added for each new setting

```yaml
broadcasting: # settings for connection to nodes
  mode: unicast # one of unicast | multicast
  multicast: # must be specified if mode = multicast
    ipv6Enabled: true # indicates whether ipv6 is enabled for multicasting
    multicastGroup: 172.28.56.77 # address of multicast group, needs to be ipv6 address if ipv6 is enabled
    interfaces: # network interfaces
      - "eth0"
      - "eth1"
  unicast: # must be specified if mode = unicast
    peers: # list of bitcoin node peers to connect to
    - host: node1
      port:
        p2p: 18333
        zmq: 28332
    - host: node2
      port:
        p2p: 18333
        zmq: 28332
    - host: node3
      port:
        p2p: 18333
        zmq: 28332
```

config/config.go
boecklim: @shotasilagadzetaal 
After a meeting with Jake & Yurii we found that `MulticastGroup` needs to be a list as there can be multiple multicast groups

shotasilagadzetaal:
> Nice work. I just have a question, because maybe I'm missing something, this config is a preparation for further changes to implement the IPv6 communications, but for now it's not yet being used anywhere, right?

yeap, exactly

test/config/config.yaml
shotasilagadzetaal: Config is according to the scheme above

kuba-4chain (approved):

test/config/config.yaml
```diff
@@ -17,19 +17,29 @@ peerRpc:
...
-      zmq: 28332
+broadcasting: # settings for connection to nodes
+  mode: unicast # one of unicast | multicast
+  multicast:
+    ipv6Enabled: true # indicates whether ipv6 is enabled for multicasting
+    multicastGroups: # must be specified if mode = multicast
+    - "172.28.56.77" # address of multicast group, needs to be ipv6 address if ipv6 is enabled
+    interfaces:
+    - "eth0"
+    - "eth1"
```
boecklim, line 28: nit: This can be removed I think

boecklim (approved):

PR #628, opened 2024-10-31
ARCO-186: add release action

pawellewandowski98:
## Description of Changes

Adding an action that creates a release when a new tag is pushed to main  branch. 

From now It is required to use conventional commits in format: `feat(ARCO-XXX): example message` thanks to which all of the commits will be properly sorted when creating the description of the new version. If you use Squash and merge option you need to use conventional commits format only when merging the PR.

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,32 @@
+# .github/workflows/release.yml
+name: release
+
+on:
+  push:
+    tags:
+      - "*"
```
boecklim, line 7: I think we should only allow releases on the `main` branch. Could this be restricted here?

kuba-4chain (approved):

.github/workflows/release.yaml
pawellewandowski98: I added it in line 15, if I would add branch name here this action will be triggered also by pushing changes into main branch. This options work separately

boecklim (approved):

PR #629, opened 2024-10-31
Optionally add attributes to traces

boecklim:
## Description of Changes

- Optionally add attributes to traces
- Remove traces from health checks

## Testing Procedure

Run `make run_e2e_tests_with_tracing`

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/cmd/arc/services/api.go
+++ b/cmd/arc/services/api.go
@@ -58,8 +58,7 @@ func StartAPIServer(logger *slog.Logger, arcConfig *config.ArcConfig) (func(), e
 
 	shutdownFns := make([]func(), 0)
 
-	tracingEnabled := false
-	if arcConfig.Tracing != nil && arcConfig.Tracing.DialAddr != "" {
+	if arcConfig.Tracing != nil && arcConfig.Tracing.DialAddr != "" && arcConfig.Tracing.Enabled {
 		cleanup, err := tracing.Enable(logger, "api", arcConfig.Tracing.DialAddr)
 		if err != nil {
 			logger.Error("failed to enable tracing", slog.String("err", err.Error()))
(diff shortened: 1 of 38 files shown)


internal/blocktx/processor.go
```diff
@@ -735,6 +737,16 @@ func (p *Processor) GetBlockRequestCh() chan BlockRequest {
...
+		if tracer == nil {
+			return ctx, nil
+		}
+
+		if len(p.attributes) > 0 {
+			ctx, span = tracer.Start(ctx, spanName, trace.WithAttributes(p.attributes...))
+			return ctx, span
+		}
+
 		ctx, span = otel.Tracer("").Start(ctx, spanName)
```
arkadiuszos4chain, line 750: Please use `tracer` from line 740

internal/blocktx/store/postgresql/postgres.go
```diff
@@ -77,7 +80,17 @@ func (p *PostgreSQL) Ping(ctx context.Context) error {
 func (p *PostgreSQL) startTracing(ctx context.Context, spanName string) (context.Context, trace.Span) {
 	if p.tracingEnabled {
 		var span trace.Span
-		ctx, span = otel.Tracer("").Start(ctx, spanName)
+		tracer := otel.Tracer("")
```
arkadiuszos4chain, line 83: I would move this to helper method to avoid repetition

cmd/arc/services/api.go
```diff
@@ -58,22 +58,19 @@ func StartAPIServer(logger *slog.Logger, arcConfig *config.ArcConfig) (func(), e
 
 	shutdownFns := make([]func(), 0)
 
-	tracingEnabled := false
-	if arcConfig.Tracing != nil && arcConfig.Tracing.DialAddr != "" {
+	if arcConfig.Tracing != nil && arcConfig.Tracing.DialAddr != "" && arcConfig.Tracing.Enabled {
```
kuba-4chain, line 61: In some places there's a check
```
arcConfig.Tracing != nil && arcConfig.Tracing.DialAddr != "" && arcConfig.Tracing.Enabled
```
and in others there's only
```
arcConfig.Tracing != nil && arcConfig.Tracing.Enabled
```

I'm thinking whether it would make sense to create a helper function, e.g. in `config/utils.go`
```
func isTracingEnabled(tracing *config.TracingConfig) bool {
    return tracing != nil && tracing.DialAddr != "" && tracing.Enabled
}
```
or maybe even this way:
```
func isTracingEnabled(tracing *config.TracingConfig) (bool, error) {
    if tracing == nil {
        return false, nil
    }

    if tracing.Enabled && tracing.DialAddr == "" {
        return nil, errors.New("tracing enabled but dial address was not given")
    }

    return tracing.Enabled
}
```
because if someone wanted to enable tracing but forgot to put a DialAddr we could just stop Arc before running and throw this error.

WDYT?

internal/blocktx/processor.go
```diff
@@ -57,6 +58,7 @@ type Processor struct {
 	registerTxsBatchSize        int
 	registerRequestTxsBatchSize int
 	tracingEnabled              bool
+	attributes                  []attribute.KeyValue
```
kuba-4chain, line 61: Nit:
Will this field be used only for tracing?

If so, maybe a better name would be something like `tracingAttributes`?

arkadiuszos4chain (approved):

cmd/arc/services/api.go
boecklim: Hey @kuba-4chain 
Yes, good input 👍 

I'll create such a helper function

kuba-4chain (approved):

PR #631, opened 2024-11-05
Feature/trace events

boecklim:
## Description of Changes

- Add tracing events
- Add more traces
- Remove event ID in context as tracing serves the same purpose
- Fix env var in docker compose file
- Rm tracing attributes from e2e test config
- Modify e2e tests so they pass locally

Trace events show for each tx the event when their status changed with timestamp and the final status it had
Screenshot 2024-11-05 at 12 52 06

## Testing Procedure

To see new traces and events run `make run_e2e_tests_with_tracing`

- [X] All tests pass locally

## Checklist:


internal/metamorph/processor.go
```diff
@@ -298,7 +298,7 @@ func (p *Processor) StartProcessMinedCallbacks() {
 }
 
 func (p *Processor) updateMined(ctx context.Context, txsBlocks []*blocktx_api.TransactionBlock) {
-	_, span := tracing.StartTracing(ctx, "updateMined", p.tracingEnabled, p.tracingAttributes...)
+	ctx, span := tracing.StartTracing(ctx, "updateMined", p.tracingEnabled, p.tracingAttributes...)
 	defer tracing.EndTracing(span)
 
 	updatedData, err := p.store.UpdateMined(p.ctx, txsBlocks)
```
arkadiuszos4chain, line 304: Shouldn't we include `UpdateMined()` in the trace as well?

kuba-4chain:
> Remove event ID in context as tracing serves the same purpose

I also think there is a place for both. EventID is greatly enhancing the logs, making it much easier and more precise to track possible issues, while tracing is great for catching time-related issues, such as slowness of responses or function executions.

Arek said it well
> tracing can help us identify bottlenecks, but it may not provide insights into the scenarios that lead to those slowdowns

I want to support that thought 😄

internal/metamorph/processor.go
kuba-4chain, line 304: We need to pass here `ctx` generated 3 lines above instead of `p.ctx` to have a continuous trace or it doesn't matter?

kuba-4chain (approved):

internal/metamorph/processor.go
kuba-4chain: To the `store.UpdateMined` function, just to clarify ☝️

boecklim: Oh, good catch 👍 

It should be `ctx` not `p.ctx`

arkadiuszos4chain (approved):

PR #636, opened 2024-11-07
refactor: sample impl with new p2p

arkadiuszos4chain:
## Description of Changes

Sample implementation using the new experimental P2P functionality.

Several files have been modified due to a package name change in go-p2p, but the most significant changes are in:

- `internal/blocktx/p2p` - includes all P2P logic, a new message handler (equivalent to the previous peer_handler), initialization, etc.
- `internal/metamorph/p2p` - new message handler, also equivalent to the previous peer_handler
- `internal/metamorph/processor.go` - updates in `StartProcessExpiredTransactions()`, now sending/requesting transactions in batches
- `internal/metamorph/store/postgresql/postgres` - introduces `IncrementRetriesForMany()`

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


go.mod
```diff
@@ -172,3 +167,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/libsv/go-p2p => github.com/arkadiuszos4chain/go-p2p v0.0.0-20241108223608-ded470665895
```
kuba-4chain, line 171: Just a reminder - to be removed and replaced with go-p2p v2

internal/metamorph/processor.go
```diff
@@ -646,9 +669,9 @@ func (p *Processor) StartProcessExpiredTransactions() {
 	}()
 }
 
-// GetPeers returns a list of connected and a list of disconnected peers
-func (p *Processor) GetPeers() []p2p.PeerI {
-	return p.pm.GetPeers()
+// GetPeersNmber returns a list of connected and a list of disconnected peers
+func (p *Processor) GetPeersNmber() uint {
+	return p.pm.CountConnectedPeers()
 }
```
kuba-4chain, line 675: I don't know if it was on purpose, but just in case 😄 
```suggestion
// GetPeersNumber returns a list of connected and a list of disconnected peers
func (p *Processor) GetPeersNumber() uint {
	return p.pm.CountConnectedPeers()
}
```

internal/metamorph/server.go
```diff
@@ -127,23 +126,25 @@ func (s *Server) Health(ctx context.Context, _ *emptypb.Empty) (*metamorph_api.H
...
+
+	// peersConnected := make([]string, 0, len(peers))
+	// peersDisconnected := make([]string, 0, len(peers))
+	// for _, peer := range peers {
+	// 	if peer.Connected() {
+	// 		peersConnected = append(peersConnected, peer.String())
+	// 	} else {
+	// 		peersDisconnected = append(peersDisconnected, peer.String())
+	// 	}
+	// }
```
kuba-4chain, line 141: It's not my call, but IMO we can remove it. But maybe, it would be useful to mark each peer in p2p as `healthy`/`unhealthy` anyways in go-p2p? I think that might be useful. WDYT?

@arkadiuszos4chain @boecklim

internal/metamorph/processor.go
```diff
@@ -607,32 +603,59 @@ func (p *Processor) StartProcessExpiredTransactions() {
...
 						}
 
-						// mark that we retried processing this transaction once more
-						if err = p.store.IncrementRetries(ctx, tx.Hash); err != nil {
-							p.logger.Error("Failed to increment retries in database", slog.String("err", err.Error()))
-						}
+						// // mark that we retried processing this transaction once more
+						// if err = p.store.IncrementRetries(ctx, tx.Hash); err != nil {
+						// 	p.logger.Error("Failed to increment retries in database", slog.String("err", err.Error()))
+						// }
```
kuba-4chain, line 618: If you're using `IncrementRetriesForMany`, this comment can be removed, right?

internal/metamorph/server.go
arkadiuszos4chain: In the new implementation, there is no need to distinguish between a connected peer and a healthy one. If the peer is connected, it is healthy. If it becomes unhealthy (error, ping-pong failed), it will disconnect and wait for the client(owner) to initiate a reconnect

arkadiuszos4chain:
not relevant anymore

PR #643, opened 2024-11-13
feat(ARCO-283): remove freecache, combine methods with cache implementation

pawellewandowski98:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


internal/metamorph/integration_test/double_spend_integration_test.go
```diff
@@ -124,11 +122,9 @@ func TestDoubleSpendDetection(t *testing.T) {
 	require.NoError(t, err)
 	defer metamorphStore.Close(context.Background())
 
-	cStore := cache.NewFreecacheStore(freecache.NewCache(baseCacheSize))
-
 	pm := &mocks.PeerManagerMock{ShutdownFunc: func() {}}
 
-	processor, err := metamorph.NewProcessor(metamorphStore, cStore, pm, statusMessageChannel,
+	processor, err := metamorph.NewProcessor(metamorphStore, nil, pm, statusMessageChannel,
```
arkadiuszos4chain, line 127: use MemoryStore

internal/metamorph/processor_helpers.go
```diff
@@ -79,7 +66,15 @@ func (p *Processor) getStatusUpdateMap() map[chainhash.Hash]store.UpdateStatus {
 	}
 
 	// If the key doesn't exist or there was an error unmarshalling the value return new map
-	return make(map[chainhash.Hash]store.UpdateStatus)
+	return make(StatusUpdateMap)
+}
+
+func (p *Processor) clearStatusUpdateMap() error {
+	if p.cacheStore == nil {
```
arkadiuszos4chain, line 73: You can remove this if

internal/metamorph/processor_test.go
```diff
@@ -73,7 +69,7 @@ func TestNewProcessor(t *testing.T) {
 	for _, tc := range tt {
 		t.Run(tc.name, func(t *testing.T) {
 			// when
-			sut, actualErr := metamorph.NewProcessor(tc.store, cStore, tc.pm, nil,
+			sut, actualErr := metamorph.NewProcessor(tc.store, nil, tc.pm, nil,
```
arkadiuszos4chain, line 72: Use MemoryStore

internal/cache/in-memory.go
kuba-4chain, line 1: File should be named `in_memory` to be consistent with other file names

internal/metamorph/integration_test/double_spend_integration_test.go
pawellewandowski98: Done

internal/cache/in-memory.go
pawellewandowski98: Done

arkadiuszos4chain (approved):

kuba-4chain (approved):

internal/cache/in_memory.go
```diff
@@ -0,0 +1,50 @@
+package cache
+
+import (
+	"sync"
+	"time"
+)
+
+type MemoryStore struct {
+	data map[string][]byte
+	mu   sync.RWMutex
```
boecklim, line 10: As an alternative a `sync.Map` could be used. We already use that e.g. here: https://github.com/bitcoin-sv/arc/blob/6669b2339f9b019dec239acc264bf0214d705e33/internal/metamorph/response_processor.go#L44
or here: https://github.com/bitcoin-sv/arc/blob/4e031edcef23927fa0b79b82fbe50642dff0efbe/internal/broadcaster/rate_broadcaster.go#L42

The advantage of the sync map is that you don't need to use a separate mutex as it is concurrency safe. Here is more information on it: https://cs.opensource.google/go/go/+/go1.23.3:src/sync/map.go;l=38

internal/metamorph/processor_helpers.go
```diff
@@ -136,7 +123,7 @@ func mergeUnique(arr1, arr2 []string) []string {
 	return uniqueSlice
 }
 
-func serializeStatusMap(updateStatusMap map[chainhash.Hash]store.UpdateStatus) ([]byte, error) { //nolint:unused
+func serializeStatusMap(updateStatusMap StatusUpdateMap) ([]byte, error) { //nolint:unused
```
boecklim, line 126: Remove nolint commet

internal/metamorph/processor_helpers.go
```diff
@@ -149,9 +136,9 @@ func serializeStatusMap(updateStatusMap map[chainhash.Hash]store.UpdateStatus) (
 	return bytes, nil
 }
 
-func deserializeStatusMap(data []byte) (map[chainhash.Hash]store.UpdateStatus, error) { //nolint:unused
+func deserializeStatusMap(data []byte) (StatusUpdateMap, error) { //nolint:unused
```
boecklim, line 139: Remove nolint commet

internal/cache/in_memory.go
pawellewandowski98: Done

internal/metamorph/processor_helpers.go
pawellewandowski98: Removed

internal/metamorph/processor_helpers.go
pawellewandowski98: Removed

internal/metamorph/processor_helpers.go
```diff
@@ -22,64 +23,95 @@ func (p *Processor) GetProcessorMapSize() int {
...
+	bytes, err := json.Marshal(status)
+	if err != nil {
+		return errors.Join(ErrFailedToSerialize, err)
+	}
 
-	if !found || shouldUpdateStatus(statusUpdate, foundStatusUpdate) {
-		if len(statusUpdate.CompetingTxs) > 0 {
-			statusUpdate.CompetingTxs = mergeUnique(statusUpdate.CompetingTxs, foundStatusUpdate.CompetingTxs)
-		}
+	err = p.cacheStore.Set(&CacheStatusUpdateHash, status.Hash.String(), bytes, processStatusUpdatesIntervalDefault)
```
kuba-4chain, line 52: I don't think it's a good idea to create a TTL that's equal to UpdateInterval, because there may be some hiccups with metamorph service, in which case we can loose data.

Also, this TTL is doing nothing when we're giving a hash to the method (see my next comment)

I would choose either:
1. A very long const TTL in order not to clear records that for some reason were never queried.
2. No TTL at all and periodical clearing cache (what I think we have currently already implemented with `clearCache` on every update.

internal/cache/redis.go
```diff
@@ -22,29 +22,54 @@ func NewRedisStore(ctx context.Context, c redis.UniversalClient) *RedisStore {
...
-	err := r.client.Set(r.ctx, key, value, ttl).Err()
+// Set stores a value with a TTL for a specific hash (if given).
+func (r *RedisStore) Set(hash *string, key string, value []byte, ttl time.Duration) error {
+	var err error
+
+	if hash == nil {
+		err = r.client.Set(r.ctx, key, value, ttl).Err()
+	} else {
+		err = r.client.HSet(r.ctx, *hash, key, value).Err()
+	}
```
kuba-4chain, line 53: We're only passing in the `ttl` when we're using the regular `Set`, otherwise, with `HSet` there's not TTL

internal/cache/in_memory.go
```diff
@@ -0,0 +1,121 @@
...
+	data sync.Map
+}
+
+func NewMemoryStore() *MemoryStore {
+	return &MemoryStore{}
+}
+
+// Get retrieves a value by key and hash (if given)
+func (s *MemoryStore) Get(hash *string, key string) ([]byte, error) {
+	if hash != nil {
```
kuba-4chain, line 19: This `hash` is optional, as I understand, but because of that, in every method we have to check `if hash != nil`. Also, currently, we're always using a hash and never access the general store.

What do you think about defining a list of `allowedHashSets` (e.g. in `cache.go`) with hashes of sets that we can use and only allow to use those HSets, instead of allowing using the general store? Or even better - just define those allowable sets in an Enum, which we can use everywhere in code.

Then, you could also check if given hash exists in `allowedHashSets` (or enum) and if not - return an ErrNotAllowedSet or something like that. You wouldn't have to retrieve values in a different way depending on whether the hash exists or not and it would not have to be a pointer.

internal/cache/redis.go
```diff
@@ -22,29 +22,54 @@ func NewRedisStore(ctx context.Context, c redis.UniversalClient) *RedisStore {
...
+// Get retrieves a value by key and hash (if given).
+func (r *RedisStore) Get(hash *string, key string) ([]byte, error) {
+	var result string
+	var err error
+
+	if hash == nil {
+		result, err = r.client.Get(r.ctx, key).Result()
+	} else {
+		result, err = r.client.HGet(r.ctx, *hash, key).Result()
+	}
```
kuba-4chain, line 34: Currently, we're only saving and returning a `[]byte` type from the store. One idea for improvement would be to make this store generic and use jsons. Example:

```go
type RedisStore struct {
	// ...
}

func (r *RedisStore) Get[T any](key string) (T, error) {
	val, err := r.client.Get(r.ctx, key)
	if err != nil {
		return result, err
	}

	// Assume Redis values are stored as JSON strings
	var result T
	err = json.Unmarshal([]byte(val), &result)
	if err != nil {
		return result, ErrFailedToParseValue
	}

	return result, nil
}

func (r *RedisStore) Set[T any](key string, value T) error {
	// Marshal the value into JSON
	data, err := json.Marshal(value)
	if err != nil {
		return ErrFailedToMarshall
	}

	err = r.client.Set(r.ctx, key, string(data))
	if err != nil {
		return ErrCacheFailedToSet
	}

	return nil
}
```

kuba-4chain: Maybe I'm going too far and this is not required, or maybe it's for another task, but I'm throwing the idea here.

WDYT? @boecklim @arkadiuszos4chain @pawellewandowski98

internal/cache/in_memory.go
kuba-4chain: Example with enum:
```go
type HashSet string

const (
  CacheStatusUpdateHashSet HashSet = "status-update"
  OtherServiceHashSet      HashSet = "different-hash-set"
)

...

func (c *Cache) Get(hash HashSet, key string) (value []byte, err error) {
  ...
}
```

internal/cache/in_memory.go
```diff
@@ -0,0 +1,139 @@
...
+	if !ok {
+		return nil, ErrCacheFailedToGet
+	}
+
+	return hashMap, nil
+}
+
+// GetAllForHashAndDelete retrieves all key-value pairs for a specific hash and deletes the hash.
+func (s *MemoryStore) GetAllForHashAndDelete(hash string) (map[string][]byte, error) {
+	hashMap, err := s.GetAllForHash(hash)
```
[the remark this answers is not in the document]
pawellewandowski98: Done

internal/metamorph/processor_helpers.go
pawellewandowski98: The cache interface was changed, TTL is used only for methods without hash which are currently not used so I'm not using this default value

internal/cache/redis.go
pawellewandowski98: The cache interface was changed, TTL is used only for methods without hash

internal/cache/redis.go
boecklim: Hey @kuba-4chain 
Right here I can't see how the generic functions are of much use as there are no 2 functions which are almost the same and a generic function could reduce duplication.

Or if I just not see it, maybe it could be added in a follow-up PR

internal/cache/redis.go
boecklim, line 1: For the Redis store an integration test would be good to have

internal/metamorph/processor_helpers.go
```diff
@@ -6,12 +6,13 @@ import (
...
 
+	"github.com/bitcoin-sv/arc/internal/cache"
 	"github.com/bitcoin-sv/arc/internal/metamorph/store"
 )
 
-//lint:file-ignore U1000 Ignore all unused code, functions are temporarily not used
+type StatusUpdateMap map[chainhash.Hash]store.UpdateStatus
 
-const CacheStatusUpdateKey = "status-updates"
+var CacheStatusUpdateHash = "status-update"
```
arkadiuszos4chain, line 15:
```suggestion
const CacheStatusUpdateKey = "mtm-tx-status-update"
```

internal/cache/cache.go
```diff
@@ -6,14 +6,24 @@ import (
...
+	ErrCacheFailedToExecuteTx = errors.New("failed to execute transaction")
 )
 
 type Store interface {
 	Get(key string) ([]byte, error)
 	Set(key string, value []byte, ttl time.Duration) error
-	Del(key string) error
+	Del(keys ...string) error
+
+	MapGet(hash string, key string) ([]byte, error)
```
arkadiuszos4chain, line 23: I would change args names:
`hash -> key|mapKey|mapID|hashsetKey|hashsetName`, `key->field|item`

Just suggestion, not a blocker

internal/metamorph/processor_helpers.go
```diff
@@ -22,64 +23,86 @@ func (p *Processor) GetProcessorMapSize() int {
...
+		return nil, err
 	}
 
-	err = p.cacheStore.Set(CacheStatusUpdateKey, bytes, processStatusUpdatesIntervalDefault)
-	if err != nil {
-		return err
+	for key, value := range keys {
+		hash, err := chainhash.NewHashFromStr(key)
+		if err != nil {
+			return nil, err
```
arkadiuszos4chain, line 84: I think we should log the error here, including all the data needed for manual incident handling, and continue execution

internal/metamorph/processor_helpers.go
```diff
@@ -22,64 +23,86 @@ func (p *Processor) GetProcessorMapSize() int {
...
+		}
+
+		var status store.UpdateStatus
+		err = json.Unmarshal(value, &status)
```
arkadiuszos4chain, line 88: I think we should log the error here, including all the data needed for manual incident handling, and continue execution

arkadiuszos4chain (changes requested):

internal/cache/redis.go
```diff
@@ -25,31 +25,118 @@ func NewRedisStore(ctx context.Context, c redis.UniversalClient) *RedisStore {
...
+}
+
+// MapDel removes a value by key in specific hash.
+func (r *RedisStore) MapDel(hash string, keys ...string) error {
+	result, err := r.client.HDel(r.ctx, hash, keys...).Result()
+
 	if err != nil {
 		return errors.Join(ErrCacheFailedToDel, err)
 	}
 	if result == 0 {
```
arkadiuszos4chain, line 93: Do we need `ErrCacheNotFound` from `*Del()` functions?

internal/cache/cache.go
pawellewandowski98: Changed

internal/cache/redis.go
pawellewandowski98: Added

boecklim (approved):

arkadiuszos4chain (approved):

PR #648, opened 2024-11-14
fix: Get unmined ancestors

boecklim:
## Description of Changes

The recursive search for unmined ancestors lead to sometimes long response times when submitting requests with the cumulative fee validation. The bitcoin node function GetMempoolAncestors offers an easy to use and fast alternative to return all unmined ancestors to a transaction. 
- Instead of recursively searching for unmined ancestors, use the node function GetMempoolAncestors from bitcoin node
- Move finder out of internal package

## Testing Procedure

- [X] All tests pass locally

## Checklist:


internal/validator/default/helpers.go
```diff
@@ -56,110 +57,71 @@ func extendTx(ctx context.Context, f validator.TxFinderI, rawTx *sdkTx.Transacti
...
-		if !found {
-			return nil, ErrParentNotFound
-		}
-
-		// fulfill data about the parent for further validation
-		bTx, err := sdkTx.NewTransactionFromBytes(p.Bytes)
-		if err != nil {
-			return nil, fmt.Errorf("cannot parse parent tx: %w", err)
-		}
+	ancestorTxs, err := txFinder.GetRawTxs(ctx, validator.SourceNodes, allTxIDs)
```
arkadiuszos4chain, line 112: Shouldn't be `validator.SourceTransactionHandler | validator.SourceWoC` ?

kuba-4chain, line 112: As I understand, previously we were using flags to get those transactions from (in order):
1. Metamorph store
2. Nodes
3. WoC

Thanks to that we had a fallback, in case the txs were not found in any of these methods.

Right now, we only try to retrieve these txs from Nodes. Was that intentional?

If it was intentional, we can remove the other flags and the code for retrieving transactions from the store and WoC.

Otherwise, maybe we could use the flags as they were used previously. WDYT?

kuba-4chain: Update:

I see the flags are used in `extendTx`, forget my comment 😄

boecklim: Yes, but I think we can actually put here again all sources

```go
const finderSource = validator.SourceTransactionHandler | validator.SourceNodes | validator.SourceWoC
```

arkadiuszos4chain (approved):

kuba-4chain (approved):

PR #656, opened 2024-11-21
feat: Error traces

boecklim:
## Description of Changes

- Add logs to metamorph client
- Add traces to message queue client
- Additional traces in ProcessTransaction function
- Set span status to ERROR if an error occurs with error message

Screenshot 2024-11-21 at 09 37 35

## Testing Procedure

run `make run_e2e_tests_with_tracing`

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/internal/tracing/helper.go
+++ b/internal/tracing/helper.go
@@ -31,8 +31,8 @@ func StartTracing(ctx context.Context, spanName string, tracingEnabled bool, att
 
 func EndTracing(span trace.Span, err error) {
 	if span != nil {
-		span.RecordError(err)
 		if err != nil {
+			span.RecordError(err)
 			span.SetStatus(codes.Error, err.Error())
 		}
 		span.End()
--- a/internal/tracing/helper.go
+++ b/internal/tracing/helper.go
@@ -5,6 +5,7 @@ import (
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 )
 
@@ -28,8 +29,12 @@ func StartTracing(ctx context.Context, spanName string, tracingEnabled bool, att
 	return ctx, span
 }
 
-func EndTracing(span trace.Span) {
+func EndTracing(span trace.Span, err error) {
 	if span != nil {
+		span.RecordError(err)
+		if err != nil {
+			span.SetStatus(codes.Error, err.Error())
+		}
 		span.End()
 	}
 }
--- a/internal/tx_finder/tx_finder.go
+++ b/internal/tx_finder/tx_finder.go
@@ -76,7 +76,9 @@ func (f Finder) GetMempoolAncestors(ctx context.Context, ids []string) ([]string
 
 func (f Finder) GetRawTxs(ctx context.Context, source validator.FindSourceFlag, ids []string) (txs []*sdkTx.Transaction, err error) {
 	ctx, span := tracing.StartTracing(ctx, "Finder_GetRawTxs", f.tracingEnabled, f.tracingAttributes...)
-	defer tracing.EndTracing(span, err)
+	defer func() {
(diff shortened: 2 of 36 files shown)


pawellewandowski98 (approved):

internal/tx_finder/tx_finder.go
```diff
@@ -75,8 +75,9 @@ func (f Finder) GetMempoolAncestors(ctx context.Context, ids []string) ([]string
 }
 
 func (f Finder) GetRawTxs(ctx context.Context, source validator.FindSourceFlag, ids []string) ([]*sdkTx.Transaction, error) {
+	var err error
 	ctx, span := tracing.StartTracing(ctx, "Finder_GetRawTxs", f.tracingEnabled, f.tracingAttributes...)
-	defer tracing.EndTracing(span)
+	defer tracing.EndTracing(span, err)
```
arkadiuszos4chain, line 80: I think we should pass nil here.
1. we do not always stop execution on error
2. it will only log the last error

IMO it would be more valuable if you passed the error to the subtraces `thGetRawTxSpan` and `wocSpan`.

internal/tracing/helper.go
```diff
@@ -28,8 +29,12 @@ func StartTracing(ctx context.Context, spanName string, tracingEnabled bool, att
...
 }
 
-func EndTracing(span trace.Span) {
+func EndTracing(span trace.Span, err error) {
 	if span != nil {
+		span.RecordError(err)
+		if err != nil {
+			span.SetStatus(codes.Error, err.Error())
+		}
 		span.End()
```
kuba-4chain, line 38: Shouldn't `span.RecordError(err)` be inside `if err != nil`?

pkg/api/handler/default.go
```diff
@@ -586,8 +600,9 @@ func (m ArcDefaultHandler) submitTransactions(ctx context.Context, txs []*sdkTx.
 }
 
 func (m ArcDefaultHandler) getTransactionStatus(ctx context.Context, id string) (*metamorph.TransactionStatus, error) {
+	var err error
 	ctx, span := tracing.StartTracing(ctx, "getTransactionStatus", m.tracingEnabled, m.tracingAttributes...)
-	defer tracing.EndTracing(span)
+	defer tracing.EndTracing(span, nil)
```
kuba-4chain, line 606: Shouldn't be `err`? Why define `var err error` if we're passing `nil` anyways?

internal/tracing/helper.go
boecklim: Hello @kuba-4chain 

Yes, it could be but it will not be a problem because: 

> ... If this span is not being recorded or err is nil then this method does nothing.

https://pkg.go.dev/go.opentelemetry.io/otel/trace@v1.31.0#Span.RecordError

arkadiuszos4chain (approved):

internal/tracing/helper.go
boecklim: Moved it inside

kuba-4chain (approved):

internal/tx_finder/tx_finder.go
```diff
@@ -131,8 +134,9 @@ func (f Finder) GetRawTxs(ctx context.Context, source validator.FindSourceFlag,
 	if source.Has(validator.SourceWoC) && len(ids) > 0 {
 		var wocSpan trace.Span
 		ctx, wocSpan = tracing.StartTracing(ctx, "WocClient_GetRawTxs", f.tracingEnabled, f.tracingAttributes...)
-		wocTxs, err := f.wocClient.GetRawTxs(ctx, ids)
-		defer tracing.EndTracing(wocSpan)
+		var wocTxs []*woc_client.WocRawTx
+		wocTxs, err = f.wocClient.GetRawTxs(ctx, ids)
+		defer tracing.EndTracing(wocSpan, err)
```
kuba-4chain, line 139: Should this `EndTracing` be without `defer` here, as this is just the span for `wocClient.GetRawTxs`, similar to TransactionHandler above?

PR #669, opened 2024-11-26
feat(ARCO-296): add blocktx stats collector

shotasilagadzetaal:
## Description of Changes

We are adding a new Prometheus custom metric e.g. arc_block_gaps_count

The value of this metric should be the number of block gaps which can be found

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/internal/blocktx/store/postgresql/stats.go
+++ b/internal/blocktx/store/postgresql/stats.go
@@ -1,0 +1,32 @@
+package postgresql
+
+import (
+	"context"
+
+	"github.com/bitcoin-sv/arc/internal/blocktx/store"
+)
+
+const BlockDistance = 2016
+
+func (p *PostgreSQL) GetStats(ctx context.Context) (*store.Stats, error) {
+	q := `
+	SELECT count(*) FROM (
+	SELECT unnest(ARRAY(
+			SELECT a.n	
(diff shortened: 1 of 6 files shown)


internal/blocktx/store/postgresql/stats.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/bitcoin-sv/arc/internal/blocktx/store"
+)
+
+func (p *PostgreSQL) GetStats(ctx context.Context) (*store.Stats, error) {
+	q := `
+	SELECT count(*) FROM (
+	SELECT unnest(ARRAY(
+			SELECT a.n	
+			FROM generate_series((SELECT max(height) - 2016 AS block_height FROM blocktx.blocks b), (SELECT max(height) AS block_height FROM blocktx.blocks b)) AS a(n)
```
kuba-4chain, line 14: Why 2016? 

It would be better to have this as a const in the top of the file rather than directly in the query

internal/blocktx/store/postgresql/stats.go
```diff
@@ -0,0 +1,30 @@
+package postgresql
+
+import (
+	"context"
+
+	"github.com/bitcoin-sv/arc/internal/blocktx/store"
+)
+
+func (p *PostgreSQL) GetStats(ctx context.Context) (*store.Stats, error) {
```
boecklim, line 9: Can you please create an integration test for this function?

boecklim (approved):

PR #679, opened 2024-11-28
feat: replace go-p2p with internal implementation in blocktx

arkadiuszos4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [x] All tests pass locally

## Checklist:


internal/blocktx/blockchain_communication/block_message.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"io"
+
+	"github.com/libsv/go-p2p/chaincfg/chainhash"
+	"github.com/libsv/go-p2p/wire"
+)
+
+// BlockMessage only stores the transaction IDs of the block, not the full transactions
+type BlockMessage struct {
```
boecklim, line 11: This was previously part of `go-p2p`
I think we should move this to `internal/p2p`

internal/blocktx/blockchain_communication/p2p/init.go
```diff
@@ -1,25 +1,22 @@
-package blocktx
+package blocktx_p2p
```
boecklim, line 1: I think this package could be called just `p2p`, not?
From the path it is clear that it is for `blocktx`

internal/blocktx/blockchain_communication/p2p/init.go
```diff
@@ -1,25 +1,22 @@
...
 
 import (
 	"encoding/binary"
-	"errors"
 	"io"
-	"log/slog"
+
+	blockchain "github.com/bitcoin-sv/arc/internal/blocktx/blockchain_communication"
```
boecklim, line 7: `blockchain_communication` is a very long package name
How about `peer_communication`?

kuba-4chain (approved):

internal/blocktx/blockchain_communication/block_message.go
kuba-4chain: In favour ✋

arkadiuszos4chain: Yes it is in `go-p2p` but it is structure created specifically for ARC BlockTX (to use it you need to override BLOCK msg reading). That's why I created this structure here and I insist to left it here

internal/blocktx/blockchain_communication/p2p/init.go
arkadiuszos4chain: I use the `p2p` package here, so I thought it would be a good idea to distinguish `blocktx_p2p`, `metamorph_p2p`, or `whatever_p2p` from the general `p2p` package. However, I can also use alias for the general package as well

internal/blocktx/blockchain_communication/p2p/init.go
arkadiuszos4chain: Yeah, I know... I don't like it either. But I have in mind that we'll add multicast soon, and I think the code will eventually be arranged something like this:
```
blocktx
   |
   |- blockchain_comunication
           |
           |- p2p/
           |- multicast/
           | some files to contoll how to use p2p and multicast
           
So, `blockchain_communication` isn't ideal (maybe it's bad), but `peer_communication` doesn't accurately describe what code will ultimately include.

kuba-4chain: Couple of ideas for better name for the package:
- `blockchain`
- `network`
- `blocknet`
- `blockchain_network`
- `chain_communication`
- `netlayer`
- `bcnet`
- `nodes_communication`

internal/blocktx/blockchain_communication/p2p/init.go
kuba-4chain: In favour of alias

internal/blocktx/blockchain_communication/p2p/init.go
arkadiuszos4chain: I'll change this name in follow up PR

internal/blocktx/blockchain_communication/p2p/init.go
arkadiuszos4chain: I will experiment with this in follow up PR

PR #684, opened 2024-12-02
fix(ARCO-298): buffered callbacker channels

boecklim:
## Description of Changes

- Make entries channels buffered. 
- Store entries on DB in case of full channels
- Rename single letter variables
- Pass send configuration to send manager

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/callbacker/send_manager.go
+++ b/internal/callbacker/send_manager.go
@@ -67,6 +67,8 @@ const (
 	ActiveMode
 	QuarantineMode
 	StoppingMode
+
+	entriesBufferSize = 10000
 )
 
 func runNewSendManager(url string, sender SenderI, store store.CallbackerStore, logger *slog.Logger, quarantinePolicy *quarantinePolicy,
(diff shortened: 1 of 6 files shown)


internal/callbacker/send_manager.go
```diff
@@ -67,46 +67,85 @@ const (
...
+		opt(m)
+	}
+
 	m.run()
 	return m
 }
 
 func (m *sendManager) Add(entry *CallbackEntry, batch bool) {
-	m.entriesWg.Add(1) // count the callbacks accepted for processing
-	go func() {
```
arkadiuszos4chain, line 101: Why would this lead to abandoning callbacks?

boecklim: Maybe the consumption of single callbacks here is not fast enough
https://github.com/bitcoin-sv/arc/blob/0629dab3a7fb6305a8ecfd8d9ff7129e091e6ad8/internal/callbacker/send_manager.go#L155

and when callbacker restarts all open go routines could be lost. Not 100% sure it is the problem. But the unbuffered channel is the only thing which looks like it could be the issue.

internal/callbacker/send_manager.go
```diff
@@ -40,10 +40,10 @@ type sendManager struct {
...
-	s          store.CallbackerStore
-	l          *slog.Logger
-	quarantine *quarantinePolicy
+	sender           SenderI
+	store            store.CallbackerStore
+	logger           *slog.Logger
+	quarantinePolicy *quarantinePolicy
 
 	// internal state
 	entriesWg    sync.WaitGroup
```
arkadiuszos4chain, line 49: Can be removed

arkadiuszos4chain (approved):

kuba-4chain (approved):

PR #687, opened 2024-12-02
feat(ARCO-299): merge register and request transactions in blocktx

kuba-4chain:
## Description of Changes

Previously, there were separate goroutines for registering and requesting transactions in blocktx.

This caused an issue - when a transaction was not registered for some reason (e.g. nats queue temporarily down), we wouldn't get it back when requesting from metamorph, even though the transaction was mined and existing in blocktx database.

Now, with this PR, they are merged together, so that even if transaction would not be registered in the first call, it would be registered in any of the subsequent calls to `request-tx` topic.

## Linked Issues / Tickets

## Testing Procedure

- [x] All tests pass locally

## Checklist:


config/example_config.yaml
```diff
@@ -102,7 +102,7 @@ blocktx:
       maxOpenConns: 80 # maximum open connections
       sslMode: disable
   recordRetentionDays: 28 # number of days for which data integrity is ensured
-  registerTxsInterval: 10s # time interval to read from the channel registered transactions
+  requestTxsInterval: 10s # time interval to register and request mined transactions
```
boecklim, line 105: I think we should keep here the name `registerTxsInterval` and also topic `RegisterTxTopic = "register-tx"` and drop `requestTxsInterval` instead because a new tx is always registered. So that is kind a core functionality.

The functionality for requesting a tx because it wasn't mined for too long is rather an exception in case that e.g. `blocktx` was failing for some reason. So that is like an auxiliary functionality.

internal/blocktx/store/postgresql/postgres_test.go
```diff
@@ -681,92 +679,8 @@ func TestPostgresStore_UpsertBlockTransactions_CompetingBlocks(t *testing.T) {
 	require.NoError(t, err)
 
 	// then
-	actual, err := sut.GetMinedTransactions(ctx, []*chainhash.Hash{txHash})
+	actual, err := sut.UpsertAndGetMinedTransactions(ctx, [][]byte{txHash[:]})
```
boecklim, line 682: Does this really cover all of the test cases from `TestPostgresStore_RegisterTransactions` ?

Maybe we should keep and adapt `TestPostgresStore_RegisterTransactions` to `TestPostgresStore_UpsertAndGetMinedTransactions`

boecklim:
Closing as not a high priority for now

PR #701, opened 2024-12-09
feat: checking shared cache for registered txs

kuba-4chain:
## Description of Changes

- Checking transactions in shared cache before saving their statuses into the store
- Integration test

## Linked Issues / Tickets

## Testing Procedure

- [x] All tests pass locally

## Checklist:


internal/metamorph/processor.go
```diff
@@ -316,6 +318,9 @@ func (p *Processor) updateMined(ctx context.Context, txsBlocks []*blocktx_api.Tr
 		if len(data.Callbacks) > 0 {
 			p.callbackSender.SendCallback(ctx, data)
 		}
+
+		// remove tx from cache - we don't expect any more status updates
+		p.delTxFromCache(data.Hash)
```
boecklim, line 323: I think we can already delete the tx from cache when it was seen on network, don't you think?

internal/metamorph/processor.go
```diff
@@ -390,13 +411,10 @@ func (p *Processor) StartSendStatusUpdate() {
...
 
-				// if we receive new update check if we have client connection waiting for status and send it
-				p.responseProcessor.UpdateStatus(msg.Hash, StatusAndError{
-					Hash:         msg.Hash,
-					Status:       msg.Status,
-					Err:          msg.Err,
-					CompetingTxs: msg.CompetingTxs,
-				})
+				// if tx is rejected, we don't expect any more status updates - remove from cache
+				if msg.Status == metamorph_api.Status_REJECTED {
```
boecklim, line 415: As stated above I think this can be 

```suggestion
if msg.Status == metamorph_api.Status_REJECTED || msg.Status == metamorph_api.Status_SEEN_ON_NETWORK {
```

internal/metamorph/processor.go
```diff
@@ -382,6 +387,22 @@ func (p *Processor) StartSendStatusUpdate() {
...
+				})
+
+				if !found {
+					found = p.txFoundInCache(msg.Hash)
+				}
+
+				if !found {
+					continue
+				}
+
```
boecklim, line 405: Please add here a debug log for the status message e.g.

```go
p.logger.Debug("Status update received", slog.String("hash", msg.Hash), slog.String("status", msg.Status))
```

internal/metamorph/processor.go
```diff
@@ -857,3 +878,25 @@ func callbackExists(callback store.Callback, data *store.Data) bool {
...
+	}
+}
+
+func (p *Processor) txFoundInCache(hash *chainhash.Hash) (found bool) {
+	if p.cacheStore.IsShared() {
+		value, _ := p.cacheStore.MapGet(CacheRegisteredTxsHash, hash.String())
+		return value != nil
+	}
+
+	// if we don't have a shared cache running, treat all transactions as found
```
boecklim, line 894: I think if the cache is not shared (in-memory) then we could still use this optimisation.
The only difference will be that if another metamorph sees a transaction it will not be updated to seen on network.

internal/metamorph/processor.go
```diff
@@ -857,3 +878,25 @@ func callbackExists(callback store.Callback, data *store.Data) bool {
 	}
 	return false
 }
+
+func (p *Processor) saveTxToCache(hash *chainhash.Hash) {
+	if p.cacheStore.IsShared() {
+		_ = p.cacheStore.MapSet(CacheRegisteredTxsHash, hash.String(), []byte("1"))
```
boecklim, line 884: I think here we can set the hash directly and not as a map with a certain ttl e.g. 5min

```suggestion
p.cacheStore.Set(hash.String(), []byte("1"), 5 * time.Minute)
```

Also we should handle the error

internal/metamorph/processor.go
kuba-4chain: Done

internal/metamorph/processor.go
kuba-4chain: You're right, removed here.

internal/metamorph/processor.go
kuba-4chain: You're right, added that

internal/metamorph/processor.go
kuba-4chain: You're right. In case the metamorph instance goes down/unresponsive/overloaded, we will lose statuses up until `SEEN_ON_NETWORK`, but then it should still get updated to `MINED`. Worst case scenario, the broken metamorph instance will have its transactions unlocked by k8s-watcher and they can be picked up by a different instance.

Okay, removed the condition.

internal/metamorph/processor.go
kuba-4chain: > When we re-broadcast the transaction here, we need to make sure that the hash is still in the cache

You're right, I added that.

---

> Also we should handle the error

You're right, I added logging the error, but I don't know whether to return an error in `ProcessTransaction` or just proceed? If the error is just some temporary hiccup, we could still proceed to announce the transaction and it will try to be added to the cache again when re-broadcasting. WDYT?

---

> I think here we can set the hash directly and not as a map with a certain ttl e.g. 5min

I didn't want to set the hash directly, in case we might need to use something like `GetAllFromMap()` from cache. Then, we wouldn't want to get everything that's in the general cache, but only things from the set. But for now, I guess we don't need it in this case.

As for TTL, I think you're right that it would be useful to clear the cache in case the transactions are not removed in the "normal usecases" for some reason.

If we want to keep it in the Map, there's a way of setting TTL with the newer version of `go-redis` library, with method `cache.HExpire()`. I can update it and use it.

Otherwise, if we don't care about map, I will just use the general cache and TTL that's already implemented. Let me know @boecklim

internal/metamorph/processor.go
```diff
@@ -857,3 +888,25 @@ func callbackExists(callback store.Callback, data *store.Data) bool {
...
+		return true
+	}
+
+	return value != nil
+}
+
+func (p *Processor) delTxFromCache(hash *chainhash.Hash) {
+	err := p.cacheStore.MapDel(CacheRegisteredTxsHash, hash.String())
+	if err != nil {
+		p.logger.Error("unable to delete transaction from cache", slog.String("hash", hash.String()), slog.String("err", err.Error()))
```
boecklim, line 910: IMO this should return the error, not log it as in `saveTxToCache`

internal/metamorph/processor.go
```diff
@@ -390,13 +410,10 @@ func (p *Processor) StartSendStatusUpdate() {
...
-				// if we receive new update check if we have client connection waiting for status and send it
-				p.responseProcessor.UpdateStatus(msg.Hash, StatusAndError{
-					Hash:         msg.Hash,
-					Status:       msg.Status,
-					Err:          msg.Err,
-					CompetingTxs: msg.CompetingTxs,
-				})
+				// if tx is seen on network or rejected, we don't expect any more status updates on this channel - remove from cache
+				if msg.Status == metamorph_api.Status_SEEN_ON_NETWORK || msg.Status == metamorph_api.Status_REJECTED {
+					p.delTxFromCache(msg.Hash)
```
boecklim, line 415: When the error is returned here, it should be logged but only if it is not of type `key not found in cache`

internal/metamorph/processor.go
kuba-4chain: Done

internal/metamorph/processor.go
```diff
@@ -857,3 +888,25 @@ func callbackExists(callback store.Callback, data *store.Data) bool {
...
 }
+
+func (p *Processor) saveTxToCache(hash *chainhash.Hash) error {
+	return p.cacheStore.MapSet(CacheRegisteredTxsHash, hash.String(), []byte("1"))
+}
+
+func (p *Processor) txFoundInCache(hash *chainhash.Hash) (found bool) {
+	value, err := p.cacheStore.MapGet(CacheRegisteredTxsHash, hash.String())
+	if err != nil {
+		p.logger.Error("count not get the transaction from hash", slog.String("hash", hash.String()), slog.String("err", err.Error()))
```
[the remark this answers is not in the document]
kuba-4chain: Done

internal/cache/in_memory.go
```diff
@@ -43,6 +63,23 @@ func (s *MemoryStore) Del(keys ...string) error {
...
+		for now := range time.Tick(interval) {
+			s.data.Range(func(key, value any) bool {
+				cacheItem, ok := value.(cacheItem)
+				if !ok {
+					return true // continue iteration
+				}
+				if now.After(cacheItem.expiration) {
+					s.data.Delete(key.(string))
+				}
+				return true // continue iteration
```
boecklim, line 77: Very nice!

boecklim (approved):

PR #725, opened 2024-12-19
feat: add timeout

shotasilagadzetaal:
## Description of Changes

Adding timeout for the whole processing of the transaction(s)

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/internal/api/handler/default.go
+++ b/internal/api/handler/default.go
@@ -58,6 +58,11 @@ type ArcDefaultHandler struct {
 	tracingAttributes             []attribute.KeyValue
 }
 
+type PostResponse struct {
+	StatusCode int
+	response   interface{}
+}
+
 func WithNow(nowFunc func() time.Time) func(*ArcDefaultHandler) {
 	return func(p *ArcDefaultHandler) {
 		p.now = nowFunc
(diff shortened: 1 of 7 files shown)


internal/api/handler/default.go
```diff
@@ -272,19 +282,45 @@ func (m ArcDefaultHandler) POSTTransaction(ctx echo.Context, params api.POSTTran
...
+	defer timer.Stop()
+
+	// Use a select statement to wait for either the function to complete or the timeout
+	select {
+	case resp := <-response:
+		return ctx.JSON(resp.StatusCode, resp.response)
+	case <-timer.C:
+		e := api.NewErrorFields(api.ErrStatusTimeout, errors.New("transaction processing timeout").Error())
+		return ctx.JSON(e.Status, e)
+	}
```
boecklim, line 323: I see a problem here. The way this is implemented it seems to me that false negative could likely happen i.e. the transaction could be being processed in metamorph already but the timeout duration has passed and this new error is returned to the user. This needs to be avoided.

How about enforcing the timeout in `processTransactions()`?

```go
func (m ArcDefaultHandler) processTransactions(ctx context.Context, txsHex []byte, options *metamorph.TransactionOptions) (
	submittedTxs []*sdkTx.Transaction, successes []*api.TransactionResponse, fails []*api.ErrorFields, processingErr *api.ErrorFields,
) {
	var err error
	ctx, span := tracing.StartTracing(ctx, "processTransactions", m.tracingEnabled, m.tracingAttributes...)
	defer func() {
		tracing.EndTracing(span, err)
	}()

	// Todo: enforce max timeout on validation step. Return error if max timeout exceeded

	// decode and validate txs
	...

	// Todo: Going forward from here subtract time spent on validation from options.MaxTimeout so that overall maximum time spent is equal to max timeout

	// submit valid transactions to metamorph
	txStatuses, e := m.submitTransactions(ctx, submittedTxs, options)
	if e != nil {
		return nil, nil, nil, e
	}
        ...
```

PR #741, opened 2025-01-10
feat: add timeout

shotasilagadzetaal:
## Description of Changes

This approach basically follows the standard timeout implementation.  We set a specific timeout in the beginning of the request and on the way always check if it is expired. If that expires before we have any transaction status then we return expiration error. 

If we have started processing the transaction (thus have some status like received) we will be returning that status upon reaching ctx expiration. 

The only issue was the context deadline also cancels grpc call which I'm trying to prevent by adding extra 2 seconds for that specific call and subtracting those 2 seconds from the beginning of the call. This ensures that the expiration happens in our controlled environment - in our code where we check `ctx.Done` so that we properly return any transaction status available at that moment. 

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


internal/metamorph/server.go
```diff
@@ -261,20 +280,14 @@ func (s *Server) processTransaction(ctx context.Context, waitForStatus metamorph
...
+		return nil
+	default:
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, timeDuration)
-	defer func() {
-		cancel()
-		close(responseChannel)
-	}()
-
```
boecklim, line 277: Yes, I think not removing this may have been the reason for false negatives previously

internal/metamorph/client.go
```diff
@@ -109,10 +104,9 @@ func WithClientTracer(attr ...attribute.KeyValue) func(s *Metamorph) {
 // NewClient creates a connection to a list of metamorph servers via gRPC.
 func NewClient(client metamorph_api.MetaMorphAPIClient, opts ...func(client *Metamorph)) *Metamorph {
 	m := &Metamorph{
-		client:     client,
-		logger:     slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo})),
-		now:        time.Now,
-		maxTimeout: maxTimeoutDefault,
```
boecklim, line 115: You can also remove `maxTimeout` from `Metamorph` struct and 
```go
func WithClientMaxTimeoutDefault(d time.Duration) func(*Metamorph) {
	return func(m *Metamorph) {
		m.maxTimeout = d
	}
}
```

internal/api/handler/default.go
```diff
@@ -262,7 +273,8 @@ func (m ArcDefaultHandler) POSTTransaction(ctx echo.Context, params api.POSTTran
 			span.SetAttributes(attr...)
 		}
 		// if an error is returned, the processing failed
-		return ctx.JSON(e.Status, e)
+		fmt.Println("aqedan brundeba")
```
boecklim, line 276: Remove this line

internal/api/handler/default.go
```diff
@@ -70,6 +76,12 @@ func WithCallbackURLRestrictions(rejectedCallbackURLSubstrings []string) func(*A
 	}
 }
 
+func WithServerMaxTimeoutDefault(timeout time.Duration) func(*ArcDefaultHandler) {
+	return func(s *ArcDefaultHandler) {
+		s.defaultTimeout = timeout
+	}
+}
```
boecklim, line 83: This function is not used anywhere

shotasilagadzetaal: yeah I know but if we needed it on metamorph's side let's have it in case we need to change it from conf 🤔

boecklim (approved):

PR #745, opened 2025-01-14
feat: add mined update

shotasilagadzetaal:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


internal/metamorph/server.go
```diff
@@ -144,8 +144,8 @@ func (s *Server) PutTransaction(ctx context.Context, req *metamorph_api.Transact
 
 	// decrease time to get initial deadline
 	newDeadline := deadline
-	if time.Now().Add(2 * time.Second).Before(deadline) {
-		newDeadline = deadline.Add(-(time.Second * 2))
+	if time.Now().Add(30 * time.Second).Before(deadline) {
+		newDeadline = deadline.Add(-(time.Second * 30))
```
boecklim, line 148: Why is it needed to change this time duration?

shotasilagadzetaal: So, with that I don't want the rpc call to expire, I want the internal functions to expire and return pretty response. I changed it to 30 seconds because that's the max request timeout that could be set.

boecklim: Could we put this time delta as a constant with a name which explains the purpose?

I guess it's not an issue, but I think this time delta which allows for the internal function to respond when the context deadline exceeds is not dependent on the max timeout, is it?

shotasilagadzetaal: No, it's just giving rpc call itself enough time NOT to expire. That's all. That const is in api side but I need that value on both api and metamorph side. Should I redefine it on metamorph side as well ? 🤔

shotasilagadzetaal: It's circular dependence I cannot import api from metamorph. I guess I will redefine that value there as well.

boecklim: It's fine that time delta is defined in both services with same name. IMO it is better to have a const there instead of just a number

boecklim (approved):

PR #755, opened 2025-01-23
feat(ARCO-291): Ordered callbacks

boecklim:
## Description of Changes

- Use ordered send manager to guarantee callbacks are sent in order
- Refactor callbacker

## Testing Procedure

- [X] All tests pass locally

## Checklist:


internal/callbacker/send_manager/send_manager.go
```diff
@@ -0,0 +1,442 @@
...
+func (m *SendManager) Enqueue(entry callbacker.CallbackEntry) {
+	if m.callbackQueue.Len() >= m.bufferSize {
+		m.storeToDB(entry)
+		return
+	}
+
+	m.callbackQueue.PushBack(entry)
+}
+
+// sortByTimestampAsc sorts the callback queue by timestamp in ascending order
```
shotasilagadzetaal, line 156: This sorting is working in N^2 time I think. Above 10,000 entries this will be a dead code.

Isn't `https://github.com/google/btree` an option here? we would simply add items there and not worry about sorting at all. Front would always return smallest (by timestamp).

boecklim: Probably this btree package would be faster to sort. But then I need to use another data structure than that linked list which is used here.

There is a maximum of items for the queue so that it cannot have more than 10'000 items: https://github.com/bitcoin-sv/arc/blob/38b431a0b181fa8c60fb9f3f536b0aa1e107356f/internal/callbacker/send_manager/send_manager.go#L54

I could make sure that this value is not exceeded like this for example

```go
func WithBufferSize(size int) func(*SendManager) {
	return func(m *SendManager) {
		if size >= entriesBufferSize {
			m.bufferSize = entriesBufferSize
			return
		}
		m.bufferSize = size
	}
}
```

internal/callbacker/processor.go
```diff
@@ -144,6 +152,101 @@ func (p *Processor) Start() error {
...
+			case <-p.ctx.Done():
+				return
+			case <-ticker.C:
+				url, err := p.store.GetUnmappedURL(ctx)
+				if err != nil {
+					p.logger.Error("Failed to fetch unmapped url", slog.String("err", err.Error()))
+					continue
+				}
+
+				if url == "" {
```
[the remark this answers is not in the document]
boecklim: Yes, it should definitely never be `url == ""`
Removed

internal/callbacker/send_manager/send_manager.go
shotasilagadzetaal: Why linked list? if you had simply slice and just called sort function every time it would be even faster I think. Anyway, you decide.

boecklim: I added another commit implementing the send manager not using the linked list but a slice

shotasilagadzetaal (approved):

PR #756, opened 2025-01-23
feat(ARCO-320): force validation

shotasilagadzetaal:
## Description of Changes

Now if the header is set we don't check if tx is present in the db or not to skip its validation. We just go ahead and validate it regardless. 

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/internal/api/handler/default.go
+++ b/internal/api/handler/default.go
@@ -228,40 +228,42 @@ func (m ArcDefaultHandler) postTransaction(ctx echo.Context, params api.POSTTran
 		return PostResponse{e.Status, e}
 	}
 
-	// check if we already have the transaction in db (so no need to validate)
-	tx, err := m.getTransactionStatus(reqCtx, txIDs[0])
-	if err != nil {
-		// if we have error which is NOT ErrTransactionNotFound, return err
-		if !errors.Is(err, metamorph.ErrTransactionNotFound) {
-			e := api.NewErrorFields(api.ErrStatusGeneric, err.Error())
-			return PostResponse{e.Status, e}
-		}
-	} else {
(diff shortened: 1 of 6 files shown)


internal/api/handler/default.go
```diff
@@ -228,40 +228,44 @@ func (m ArcDefaultHandler) postTransaction(ctx echo.Context, params api.POSTTran
...
-	// check if we already have the transaction in db (so no need to validate)
-	tx, err := m.getTransactionStatus(reqCtx, txIDs[0])
-	if err != nil {
-		// if we have error which is NOT ErrTransactionNotFound, return err
-		if !errors.Is(err, metamorph.ErrTransactionNotFound) {
-			e := api.NewErrorFields(api.ErrStatusGeneric, err.Error())
-			return PostResponse{e.Status, e}
-		}
+	if params.XForceValidation != nil && *params.XForceValidation {
+		transactionOptions.SkipTxValidation = false
```
boecklim, line 232: I think this logic should be at the end of `getTransactionsOptions` https://github.com/bitcoin-sv/arc/blob/c1921162ec9d709bba6a3505f11492abb461c978/internal/api/handler/default.go#L527

internal/api/handler/default.go
```diff
@@ -228,40 +228,44 @@ func (m ArcDefaultHandler) postTransaction(ctx echo.Context, params api.POSTTran
...
 	} else {
```
boecklim, line 233: After change below here

```go
	if !transactionOptions.ForceValidation { 
       ...
       }
```

boecklim (approved):

PR #759, opened 2025-01-27
fix: Tx incorrect status

boecklim:
## Description of Changes

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/internal/metamorph/processor.go
+++ b/internal/metamorph/processor.go
@@ -2,15 +2,18 @@ package metamorph
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"log/slog"
+	"net/url"
 	"os"
 	"sync"
 	"time"
 
 	"github.com/libsv/go-p2p"
 	"github.com/libsv/go-p2p/chaincfg/chainhash"
+	"github.com/ordishs/go-bitcoin"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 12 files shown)


internal/metamorph/processor.go
```diff
@@ -732,6 +744,38 @@ func (p *Processor) StartProcessExpiredTransactions() {
...
+		return err
+	}
+
+	// If all inputs already exist, then the status is rejected as this transaction can not possibly be mined
+	for _, input := range tx.Inputs {
+		txInfo, err := p.nodeClient.GetTransactionInfo(ctx, hex.EncodeToString(input.SourceTXID))
+		if err != nil {
+			return err
+		}
+		if txInfo.BlockHash != "" || txInfo.BlockHeight > 0 {
```
shotasilagadzetaal, line 771: So, as I was saying there still is case where we assign rejected status but transaction gets mined eventually.

Imagine we submit chained transaction and for some transactions they don't see parent yet in mempool so we receive ORPHANED status. 

Now we do this logic ⬆️ and before we started this check those parent transactions got mined in a new block. So those input transactions now have `txInfo.BlockHash != "" || txInfo.BlockHeight > 0` so we assign REJECT status to this one but in fact it will be mined (probably in the next block).

I admit this case is pretty rare but that's why I didn't like the approach because there is the case when we tell user that transaction is rejected but it gets mined eventually which is very bad.

PR #776, opened 2025-02-12
test(ARCO-332): Coverage review

domingotaal:
## Description of Changes

Added tests for the files with lowest coverage so far

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/logger/logger_test.go
+++ b/internal/logger/logger_test.go
@@ -16,15 +16,39 @@ func Test_NewLogger(t *testing.T) {
 		expectedError error
 	}{
 		{
-			name:          "valid logger",
+			name:          "valid logger - INFO",
 			loglevel:      "INFO",
 			logformat:     "text",
 			expectedError: nil,
 		},
 		{
-			name:          "valid logger",
+			name:          "valid logger - INFO",
 			loglevel:      "INFO",
 			logformat:     "json",
+			expectedError: nil,
+		},
+		{
+			name:          "valid logger - WARN",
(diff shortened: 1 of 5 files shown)


internal/logger/logger_test.go
```diff
@@ -47,13 +71,27 @@ func Test_NewLogger(t *testing.T) {
...
+				}
+				if tc.loglevel == "WARN" {
+					assert.Equal(t, sut.Enabled(context.Background(), slog.LevelWarn), true)
+				}
+				if tc.loglevel == "DEBUG" {
+					assert.Equal(t, sut.Enabled(context.Background(), slog.LevelDebug), true)
+				}
+				if tc.loglevel == "TRACE" {
+					assert.Equal(t, sut.Enabled(context.Background(), slog.LevelInfo), true)
+				}
```
boecklim, line 94: Here you could use a switch statement instead which is in general a bit easier to read


```go
				switch tc.loglevel {
				case "ERROR":
					assert.Equal(t, sut.Enabled(context.Background(), slog.LevelError), true)
				case "INFO":
					assert.Equal(t, sut.Enabled(context.Background(), slog.LevelInfo), true)
				case "WARN":
					...
				}
```

boecklim (approved):

internal/logger/logger_test.go
domingotaal: Yes I added the ifs temporary but I forgot later when I started having issues with lint. I will fix those in the next changes.

PR #782, opened 2025-02-18
test(ARCO-332): ZMQ coverage review

domingotaal:
## Increase coverage of ZMQ . 
- Added tests for the ZMQ Handler
- Added tests for ZMQ socket
- Lint cleanup

## Testing Procedure

- [x] All tests pass locally

## Checklist:


internal/cache/in_memory.go
```diff
@@ -72,7 +72,9 @@ func (s *MemoryStore) startClearCache(interval time.Duration) {
 					return true // continue iteration
 				}
 				if now.After(cacheItem.expiration) {
-					s.data.Delete(key.(string))
+					item, err := key.(string)
+					_ = err
+					s.data.Delete(item)
```
boecklim, line 77: IMO the correct way to handle the error is like it is done on lines 70-73
`err` is not an error, but a boolean. I'd suggest this
```go
item, ok := key.(string)
if !ok {
  return true
}
s.data.Delete(item)


```

internal/metamorph/zmq_test.go
```diff
@@ -1,25 +1,39 @@
...
+	"time"
 )
 
+var endpoint = "tcp://127.0.0.1:5555"
+var zmqEndpointURL, _ = url.Parse(endpoint)
+var zmqURLDOWN, _ = url.Parse("tcp://URLDOWN:5556")
+var statusMessageCh chan *metamorph_p2p.TxStatusMessage
+var zmqMessages = make(chan []string, 10)
+
+// var validTopics = []string{"hashblock", "hashblock2", "hashtx", "discardedfrommempool", "hashtx2", "invalidtx", "rawblock", "rawblock2", "rawtx", "rawtx2", "removedfrommempoolblock"}
```
boecklim, line 27: remove line

internal/metamorph/zmq_test.go
```diff
@@ -148,3 +162,139 @@ func TestZMQDoubleSpend(t *testing.T) {
 
 	assert.Equal(t, numberOfMsgs, sCounter)
 }
+
+func TestNewZMQHandler(t *testing.T) {
```
boecklim, line 166: I think this test can be made a bit simpler and table based.
Using a go routine does not give an added value.

```go

func TestNewZMQHandler(t *testing.T) {
	// Test Case
	// Given I want to test metamorph handler
	// When I have a ZMQ publisher up
	// Then I want to make sure the handler can
	// subscribe and unsubscribe to all topics
	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{}))
	var handler *metamorph.ZMQHandler
	var zmq *metamorph.ZMQ
	handler = metamorph.NewZMQHandler(context.Background(), zmqEndpointURL, logger)
	require.NotNil(t, handler)
	zmq, err := metamorph.NewZMQ(zmqEndpointURL, statusMessageCh, handler, logger)
	require.NoError(t, err)
	closeZMQ, err := zmq.Start()
	require.NoError(t, err)
	defer closeZMQ()
	tt := []struct {
		name        string
		expectedErr bool
	}{
		{name: "notvalid", expectedErr: true},
		{name: "hashblock"},
		{name: "discardedfrommempool"},
		{name: "hashtx2"},
		{name: "invalidtx"},
	}

	for _, tc := range tt {
			logger.Info("Listening to ZMQ", slog.String("host", zmqEndpointURL.Hostname()), slog.String("port", zmqEndpointURL.Port()))

			err = handler.Subscribe(tc.name, zmqMessages)
			if tc.expectedErr {
				require.Error(t, err)
			} else {
				require.NoError(t, err)
			}

			err = handler.Unsubscribe(tc.name, zmqMessages)
			if tc.expectedErr {
				require.Error(t, err)
			} else {
				require.NoError(t, err)
			}
		})
	}
}

```

internal/metamorph/zmq_test.go
```diff
@@ -148,3 +162,139 @@ func TestZMQDoubleSpend(t *testing.T) {
...
+	// Then I want to make sure that the appropriate error arises
+
+	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{}))
+	handler := metamorph.NewZMQHandler(ctx, zmqURLDOWN, logger)
+	zmqObj, err := metamorph.NewZMQ(zmqURLDOWN, statusMessageCh, handler, logger)
+	if err != nil {
+		logger.Error("failed to create ZMQ: %v")
+	} else {
+		logger.Info("No error was raised using a made up URL")
+	}
```
boecklim, line 238: Do you expect here an error?
If no
```
require.NoError(t, err)
```

internal/metamorph/zmq_test.go
```diff
@@ -148,3 +162,139 @@ func TestZMQDoubleSpend(t *testing.T) {
...
+
+		err = handler.Unsubscribe("invalidtx", zmqMessages)
+		require.NoError(t, err)
+		wg.Done()
+	}()
+	wg.Wait()
+	defer closezmq()
+}
+
+func TestZMQHandlerNoErrorServiceDown(t *testing.T) {
```
boecklim, line 223: What is here tested, that wasn't already tested int `TestNewZMQHandler`?

internal/metamorph/zmq_test.go
```diff
@@ -148,3 +162,139 @@ func TestZMQDoubleSpend(t *testing.T) {
...
+	defer cleanup()
+
+	err = handler.Subscribe(ZmqFirstTopic, zmqMessages)
+	require.NoError(t, err)
+
+	err = handler.Unsubscribe(ZmqFirstTopic, nil)
+	require.NoError(t, err)
+}
+
+func TestZMQ4SendReceiveMessage(t *testing.T) {
```
boecklim, line 250: What functionality of ARC is tested here?
It looks like only the `github.com/go-zeromq/zmq4` package is tested here which is an external dependency. That does not need to be tested here

internal/cache/in_memory.go
domingotaal: Thanks for explaining. I agree. Changed.

internal/metamorph/zmq_test.go
domingotaal: Ok. Done.

internal/metamorph/zmq_test.go
domingotaal: Same comment as before. Just for you guys to check the use case. If the handler can connect to any url. If yes, we remove the test entirely.

internal/metamorph/zmq_test.go
domingotaal: Yes, I will remove that. I was using it during my tests but that's not relevant anymore.

internal/metamorph/zmq_test.go
domingotaal: I tried that approach but the handler does not connect so it did not cover almost any code. I will try a different approach based on this. Thanks.

internal/cache/in_memory.go
```diff
@@ -72,7 +72,12 @@ func (s *MemoryStore) startClearCache(interval time.Duration) {
 					return true // continue iteration
 				}
 				if now.After(cacheItem.expiration) {
-					s.data.Delete(key.(string))
+					item, ok := key.(string)
+					if !ok {
+						return true
+					}
+					s.data.Delete(item)
+					s.data.Delete(item)
```
boecklim, line 80: This line is a duplicate and can be removed

internal/metamorph/zmq_test.go
```diff
@@ -148,3 +193,87 @@ func TestZMQDoubleSpend(t *testing.T) {
 
 	assert.Equal(t, numberOfMsgs, sCounter)
 }
+
+func TestNewZMQHandler(t *testing.T) {
```
boecklim, line 197: IMO this test should be moved to `zmq_handler_test.go`

internal/cache/in_memory.go
```diff
@@ -72,7 +72,12 @@ func (s *MemoryStore) startClearCache(interval time.Duration) {
 					return true // continue iteration
 				}
 				if now.After(cacheItem.expiration) {
-					s.data.Delete(key.(string))
+					item, ok := key.(string)
+					if !ok {
+						return true
```
shotasilagadzetaal, line 77: Isn't that unacceptable? shouldn't we, at least, log some warning? 🤔

shotasilagadzetaal (approved):

internal/cache/in_memory.go
boecklim: Yes, a warning log would be good maybe 👍

domingotaal: Is it possible that we may be flooded with warning logs when a big cache expire?
Maybe log one warning outside of the loop?

boecklim: I don't expect it to ever happen because the key is always a string: https://github.com/bitcoin-sv/arc/blob/3d152111b3c6910a63ef0457e44c5d380ad1cd7a/internal/cache/in_memory.go#L51

But if it still happens then, we'll see it in the logs giving us the possibility to see the issue.

internal/metamorph/zmq_handler_test.go
```diff
@@ -0,0 +1,96 @@
...
+				} else {
+					require.NoError(t, err)
+				}
+			}()
+			time.Sleep(1 * time.Second)
+		})
+	}
+}
+
+func ZMQ4StartServer(t *testing.T) (zmq4.Socket, zmq4.Socket) {
```
boecklim, line 76: I don't think this is starting a server

domingotaal: Yes, that's the issue. With pebbe library it kind of did and coverage went up to 60% or so. We can leave the ToDo as at least this tests some of the code.

PR #801, opened 2025-03-03
feat(BPAAS-1990): add api health check

shotasilagadzetaal:
## Description of Changes

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


cmd/arc/services/api.go
```diff
@@ -121,6 +122,36 @@ func StartAPIServer(logger *slog.Logger, arcConfig *config.ArcConfig, shutdownCh
...
+
+		resp, err = btxClient.Check(ctx, &grpc_health_v1.HealthCheckRequest{Service: ""})
+		if err != nil {
+			return fmt.Errorf("health check failed for blocktx grpc: %v", err)
+		}
+
+		if resp.Status != grpc_health_v1.HealthCheckResponse_SERVING {
+			return fmt.Errorf("blocktx grpc server not serving: %v", resp.Status)
+		}
+		return nil
```
boecklim, line 152: I think this logic should be in `GETHealth(ctx echo.Context) (err error)` directly so that a test can be written which covers the logic

internal/grpc_utils/server.go
```diff
@@ -35,6 +36,10 @@ func NewGrpcServer(logger *slog.Logger, cfg ServerConfig) (GrpcServer, error) {
 	}
 
 	grpcSrv := grpc.NewServer(grpcOpts...)
+
+	// register health server endpoint
+	grpc_health_v1.RegisterHealthServer(grpcSrv, health.NewServer())
```
boecklim, line 41: Is this line needed?
We have it here as well: https://github.com/bitcoin-sv/arc/blob/d457c82c460a3def9375a963339b32e75c13172f/internal/grpc_utils/server.go#L50

shotasilagadzetaal: As I check that I  think I'm doing something different. There is no need to `srv.ListenAndServe` for health server I think. It just needs to be part of the existing server. 

So the line `grpc_health_v1.RegisterHealthServer(grpcSrv, health.NewServer())` is enough. I removed other functions.

cmd/arc/services/api.go
shotasilagadzetaal: yes I though so but the thing is that I don't have `conn` and `btcConn` connection variables there. Then I need to pass them as well in constructor and have this function in api that I thought would be more complicated.

internal/grpc_utils/server.go
boecklim: The health server removed here is necessary: https://github.com/bitcoin-sv/arc/pull/801/commits/d9fdacc4fef2f832c115f54d0fa57f8ab116a12e

You need the specific health server because it implements the grpc health checking protocol: https://github.com/grpc/grpc/blob/master/doc/health-checking.md

This server is responding to the grpc health probe: https://github.com/bitcoin-sv/arc/blob/f248943c4c84d00143eaa1f577ff2be237f125c1/docker-compose.yaml#L184

cmd/arc/services/api.go
boecklim: It may be more complicated

internal/grpc_utils/server.go
shotasilagadzetaal: Yes but without starting that server it still responds to `Check` call 🤔  I tested it locally and it succeeded.

What am I missing here

cmd/arc/services/api.go
boecklim: I think this could be done like this

```go
func (m ArcDefaultHandler) GETHealth(ctx echo.Context) (err error) {
	err = m.TransactionHandler.Health(ctx.Request().Context())
        ...
```

Maybe for now we can leave the ping to `blocktx` and add this later as most important here is the ping to `metamorph`.

boecklim: If you use the already available function `err = m.TransactionHandler.Health(ctx.Request().Context())`, then you don't need the `conn` variable.

docker-compose.yaml
```diff
@@ -181,7 +181,7 @@ services:
       migrate-blocktx:
         condition: service_completed_successfully
     healthcheck:
-      test: ["CMD", "/bin/grpc_health_probe", "-addr=:8006", "-service=liveness", "-rpc-timeout=5s"]
+      test: ["CMD", "/bin/grpc_health_probe", "-addr=:8011", "-service=", "-rpc-timeout=5s"]
```
boecklim, line 184: If the health server keeps working as it did before, then e2e tests should pass while having here 
`"-service=liveness"`

docker-compose.yaml
```diff
@@ -203,7 +203,7 @@ services:
       migrate-callbacker:
         condition: service_completed_successfully
     healthcheck:
-      test: ["CMD", "/bin/grpc_health_probe", "-addr=:8025", "-service=liveness", "-rpc-timeout=5s"]
+      test: ["CMD", "/bin/grpc_health_probe", "-addr=:8021", "-service=", "-rpc-timeout=5s"]
```
boecklim, line 206: If the health server keeps working as it did before, then e2e tests should pass while having here 
`"-service=liveness"`

config/example_config.yaml
```diff
@@ -56,9 +56,6 @@ metamorph:
     fromAgo: 24h # start of moving time window
     untilAgo: 1h # end of moving time window
   monitorPeers: false # if enabled, peers which do not receive alive signal from nodes will be restarted
-  health:
-    serverDialAddr: localhost:8005 # address at which the grpc health server is exposed
-    minimumHealthyConnections: 2 # minimum number of healthy peer connections for the processor to be considered healthy
```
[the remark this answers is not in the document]
shotasilagadzetaal: As it's the only parameter there I thought having it as a constant in code would be ok

cmd/arc/services/api.go
```diff
@@ -131,6 +131,34 @@ func StartAPIServer(logger *slog.Logger, arcConfig *config.ArcConfig) (func(), e
 		policy = arcConfig.API.DefaultPolicy
 	}
 
+	// apiOpts = append(apiOpts, apiHandler.WithReadinessCheck(func() error {
```
shotasilagadzetaal, line 134: I'll leave it commented here.

internal/api/handler/default.go
```diff
@@ -154,26 +162,16 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) (err error) {
...
+		errMsg := err.Error()
+		reason = &errMsg
 	}
 
 	return ctx.JSON(http.StatusOK, api.Health{
-		Healthy: PtrTo(true),
+		Healthy: PtrTo(m.TransactionHandler.Health(ctx.Request().Context()) == nil),
 		Version: &version.Version,
-		Reason:  nil,
+		Reason:  reason,
```
boecklim, line 174: In order to call the health endpoint only once I'd suggest this

```suggestion
	var reason *string
	err := m.TransactionHandler.Health(ctx.Request().Context())
	if err != nil {
		errMsg := err.Error()
		reason = &errMsg
	}

	return ctx.JSON(http.StatusOK, api.Health{
		Healthy: PtrTo(err == nil),
		Version: &version.Version,
		Reason:  reason,
}

```

internal/callbacker/callbacker_api/callbacker_api.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.5
+// 	protoc-gen-go v1.34.2
```
boecklim, line 3: This has been generated with a lower version of `protoc-gen-go`
IMO we should if the version changes it should only be a later version, not a previous version

internal/metamorph/metamorph_api/metamorph_api.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.5
+// 	protoc-gen-go v1.34.2
```
boecklim, line 3: This has been generated with a lower version of `protoc-gen-go`
IMO we should if the version changes it should only be a later version, not a previous version

internal/blocktx/blocktx_api/blocktx_api.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.5
+// 	protoc-gen-go v1.34.2
```
boecklim, line 3: This has been generated with a lower version of `protoc-gen-go`
IMO we should if the version changes it should only be a later version, not a previous version

cmd/arc/services/api.go
```diff
@@ -131,6 +131,34 @@ func StartAPIServer(logger *slog.Logger, arcConfig *config.ArcConfig) (func(), e
...
+	// 	resp, err = btxClient.Check(ctx, &grpc_health_v1.HealthCheckRequest{Service: ""})
+	// 	if err != nil {
+	// 		return fmt.Errorf("health check failed for blocktx grpc: %v", err)
+	// 	}
+
+	// 	if resp.Status != grpc_health_v1.HealthCheckResponse_SERVING {
+	// 		return fmt.Errorf("blocktx grpc server not serving: %v", resp.Status)
+	// 	}
+	// 	return nil
+	// }))
```
boecklim, line 160: remove

internal/api/handler/default.go
```diff
@@ -41,6 +41,7 @@ var (
 	ErrStatusNotSupported       = errors.New("status not supported")
 	ErrDecodingBeef             = errors.New("error while decoding BEEF")
 	ErrMaxTimeoutExceeded       = fmt.Errorf("max timeout can not be higher than %d", metamorph.MaxTimeout)
+	ReadinessNotDefined         = string("readiness check not defined")
```
boecklim, line 44: Not needed anymore

internal/api/handler/default.go
```diff
@@ -87,6 +89,12 @@ func WithCacheExpiryTime(d time.Duration) func(*ArcDefaultHandler) {
 	}
 }
 
+func WithReadinessCheck(f func() error) func(*ArcDefaultHandler) {
+	return func(p *ArcDefaultHandler) {
+		p.ready = f
+	}
+}
```
boecklim, line 96: Not needed anymore

internal/api/handler/default.go
```diff
@@ -49,6 +50,7 @@ type ArcDefaultHandler struct {
 
 	logger                        *slog.Logger
 	now                           func() time.Time
+	ready                         func() error
```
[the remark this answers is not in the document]
shotasilagadzetaal: Won't we need it in the future? that's why I commented that ready function implementation 🤔

boecklim (approved):

PR #802, opened 2025-03-03
feat(BPAAS-1990): add nats status check

shotasilagadzetaal:
## Description of Changes

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


internal/blocktx/server.go
```diff
@@ -52,8 +55,15 @@ func NewServer(logger *slog.Logger, store store.BlocktxStore, pm *p2p.PeerManage
 }
 
 func (s *Server) Health(_ context.Context, _ *emptypb.Empty) (*blocktx_api.HealthResponse, error) {
+	ok := false
+	status := ""
+	if s.mq != nil && s.mq.Status() == nats.CONNECTED {
+		ok = true
+		status = s.mq.Status().String()
+	}
```
boecklim, line 63: This check has to be added also to the health server in order to have any effect to the K8s probes: https://github.com/bitcoin-sv/arc/blob/59689ba9983f087546c90d0185b063200c554faa/internal/blocktx/health_check.go#L20

internal/metamorph/server.go
```diff
@@ -130,7 +132,13 @@ func (s *Server) Health(ctx context.Context, _ *emptypb.Empty) (healthResp *meta
 		}
 	}
 
+	status := ""
+	if s.mq != nil {
+		status = s.mq.Status().String()
+	}
+
```
boecklim, line 139: This check has to be added also to the health server in order to have any effect to the K8s probes:
https://github.com/bitcoin-sv/arc/blob/dd704e1237a457e497fc4d04ec3459ac287380cf/internal/metamorph/health_check.go#L21

internal/callbacker/server.go
```diff
@@ -43,7 +46,12 @@ func NewServer(logger *slog.Logger, dispatcher Dispatcher, callbackerStore store
 }
 
 func (s *Server) Health(_ context.Context, _ *emptypb.Empty) (*callbacker_api.HealthResponse, error) {
+	status := nats.DISCONNECTED.String()
+	if s.mqClient != nil {
+		status = s.mqClient.Status().String()
+	}
```
boecklim, line 52: This check has to be added also to the health server in order to have any effect to the K8s probes:

https://github.com/bitcoin-sv/arc/blob/ef984dc0ed9befd820eb4ba6b9c8c5af3761d832/internal/callbacker/health_check.go#L9

internal/blocktx/blocktx_api/blocktx_api.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.5
+// 	protoc-gen-go v1.34.2
```
boecklim, line 3: When updating generated protobuf files, then we should use at least the same version of protoc-gen-go or heigher

internal/blocktx/server.go
```diff
@@ -24,11 +26,12 @@ type Server struct {
 	pm                            *p2p.PeerManager
 	store                         store.BlocktxStore
 	maxAllowedBlockHeightMismatch int
+	mq                            mq.MessageQueueClient
```
boecklim, line 29: I think there is no need to add a separate message queue object.
You could access it like this
```go
s.processor.mqClient
```

internal/callbacker/server.go
```diff
@@ -43,7 +47,12 @@ func NewServer(logger *slog.Logger, dispatcher Dispatcher, callbackerStore store
...
```
boecklim, line 53: This check is done 3 times in a different way. I think it should be done the same way in all places
`metamorph/server.go`
```go
	status := ""
	if s.mq != nil {
		status = s.mq.Status().String()
	}
```

`callbacker/server.go`
```go
	status := nats.DISCONNECTED.String()
	if s.mqClient != nil {
		status = s.mqClient.Status().String()
	}
```

`blocktx/server.go`
```go
	ok := false
	status := ""
	if s.mq != nil && s.mq.Status() == nats.CONNECTED {
		ok = true
		status = s.mq.Status().String()
	}
```

boecklim (approved):

PR #810, opened 2025-03-10
feat: tx submission counter

boecklim:
## Description of Changes

New prometheus metric counting submissions in `api` after successful validation

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/api/handler/stats.go
+++ b/internal/api/handler/stats.go
@@ -1,0 +1,45 @@
+package handler
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var ErrFailedToRegisterStats = errors.New("failed to register stats collector")
+
+type Stats struct {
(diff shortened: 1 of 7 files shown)


internal/api/handler/stats.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return nil, errors.Join(ErrFailedToRegisterStats, err)
+	}
+
+	return p, nil
+}
+
+func (s *Stats) Add(inc int) {
+	s.mu.Lock()
+	s.apiTxSubmissions.Add(float64(inc))
```
shotasilagadzetaal, line 35: I think prometheus.Counter is already thread safe, no need for locking

boecklim: Ok, that would be nice 👍 
Will check...

shotasilagadzetaal (approved):

PR #820, opened 2025-03-24
feat(BPAAS-2022): update sdk repository

shotasilagadzetaal:
## Description of Changes

As a note: `github.com/bsv-blockchain/go-sdk@v1.1.22 requires go >= 1.24.0; switching to go1.24.1`

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


Makefile
```diff
@@ -138,7 +138,7 @@ install_gen:
 	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.5
 	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.5.1
 	go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@v2.4.1
-	go install github.com/matryer/moq@v0.4.0
+	go install github.com/matryer/moq@latest
```
boecklim, line 141: IMO it is better to have a fixed version here and upgrade it once in a while, because that defines which version to use for all maintainers. Different versions of `moq` can lead to differences in generated mock files.

shotasilagadzetaal: That's the part I was curious about. Please also note I do `version: latest` in pipeline. Is that Ok? 

How do I specify the version here if not? So which version then?

boecklim: You find the latest version here: https://github.com/matryer/moq/releases

That would be 0.5.3: https://github.com/matryer/moq/releases/tag/v0.5.3

In the pipeline the same version of `moq` is used as defined in Makefile here

```
      - name: Install code generation tools
        run: make install_gen

      - name: Check go format
        run: gofmt -s -w . && git diff --exit-code

      - name: Check go generate
        run: make gen_go && git diff --exit-code
```

boecklim (approved):

Makefile
shotasilagadzetaal: Yea I found that, what about `staticcheck` and `lint` I have `latest` there as well or is it ok for them? 🤔

PR #822, opened 2025-03-28
Metamorph Server Coverage

domingotaal:
## Description of Changes

## Testing Procedure

- [X] All tests pass locally

## Checklist:


internal/metamorph/server_test.go
```diff
@@ -347,6 +347,273 @@ func TestServer_GetTransactionStatus(t *testing.T) {
...
+			require.NoError(t, err)
+			defer sut.GracefulStop()
+
+			// when
+			got, err := sut.GetTransactionStatuses(context.Background(), tt.req)
+
+			// then
+			if !tt.wantErr(t, err, fmt.Sprintf("GetTransactionStatus(%v)", tt.req)) {
+				return
+			}
```
boecklim, line 611: In almost every other test we use this pattern here

```go
			if tc.expectedError != nil {
				require.ErrorIs(t, actualErr, tc.expectedError)
				return
			}
			require.NoError(t, actualErr)
```

where `tc.expectedError` is of type `error`

internal/metamorph/server_test.go
```diff
@@ -794,3 +1061,102 @@ func TestGetTransactions(t *testing.T) {
...
+			// then
+			if tc.getFromStoreErr != nil {
+				require.Equal(t, tc.getFromStoreErr, err)
+			} else {
+				require.NoError(t, err)
+				require.NotNil(t, res)
+				if tc.getFromStoreResponseCount > 0 {
+					require.NotNil(t, res.Txid)
+				}
+			}
```
boecklim, line 1136: That is similar to most other tests, but with different naming

```go
			if tc.expectedError != nil {
				require.ErrorIs(t, actualErr, tc.expectedError)
				return
			}
			require.NoError(t, actualErr)
```

where `tc.expectedError` is of type `error`

internal/metamorph/server_test.go
domingotaal: Done.

internal/metamorph/server_test.go
domingotaal: Done

internal/metamorph/server_test.go
boecklim: Hey @domingotaal 
It seems not to have changed ;)

internal/metamorph/server_test.go
boecklim: Hey @domingotaal
It seems not to have changed ;)

domingotaal: :-(, somehow when I reverted the changes from golangci-lint this got reverted too. Now should be fine 🤞

boecklim (approved):

PR #827, opened 2025-04-01
refactor: dockerfile

boecklim:
## Description of Changes

- Reuse Dockerfile for multicast case
- Update multicast e2e test config & docker compose file
- Resolve a couple of issues indicated by SonarQube
- Updgrade golang.org/x/net to fix https://github.com/bitcoin-sv/arc/security/dependabot/29

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/internal/blocktx/processor.go
+++ b/internal/blocktx/processor.go
@@ -839,7 +839,7 @@ func (p *Processor) acceptIntoChain(ctx context.Context, blocks []*blocktx_api.B
 
 func (p *Processor) publishMinedTxs(ctx context.Context, txs []store.BlockTransactionWithMerklePath) error {
 	var publishErr error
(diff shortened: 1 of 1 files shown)


test/config/config_mcast.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 logFormat: tint
 logLevel: INFO
-profilerAddr: localhost:9999
+profilerAddr: "[::]:9999"
```
shotasilagadzetaal, line 4: I'm not sure if we need ipv6 for profiler or health check endpoints 🤔

shotasilagadzetaal (approved):

test/config/config_mcast.yaml
boecklim: Yes, that may be true. But I actually just moved it from line 96 and 52

PR #835, opened 2025-04-08
refactor: rate broadcaster

boecklim:
## Description of Changes

- Use Ticker interface in rate broadcaster
- Replace math/rand with crypto/rand

## Testing Procedure

- [X] All tests pass locally

## Checklist:


internal/broadcaster/rate_broadcaster_test.go
```diff
@@ -170,28 +168,37 @@ func TestRateBroadcaster(t *testing.T) {
...
 			err = sut.Start()
-			defer sut.Shutdown()
+
 			if tc.expectedError != nil {
 				require.ErrorIs(t, err, tc.expectedError)
 				return
 			}
 			require.NoError(t, err)
 			time.Sleep(tc.waitingTime)
-			go sut.Wait()
```
domingotaal, line 185: go sut.Wait() was added to cover that specific function that otherwise is left uncovered

domingotaal (approved):

internal/broadcaster/rate_broadcaster_test.go
```diff
@@ -170,28 +168,37 @@ func TestRateBroadcaster(t *testing.T) {
...
 
 			// then
 			require.Equal(t, tc.expectedBroadcastTransactionsCalls, len(client.BroadcastTransactionsCalls()))
 			require.Equal(t, tc.expectedLimit, sut.GetLimit())
 			require.Equal(t, int64(0), sut.GetConnectionCount())
-			if !tc.asyncTest {
```
domingotaal, line 191: In the case of the asynchronous test, this was not being tested because the response was not consistent. It seems that after the refactor it is the same and this might be needed again. Then maybe we can investigate whether it is ok that the response is not consistent.

internal/broadcaster/rate_broadcaster_test.go
boecklim: Ok. I added it again at the end of the test after `Shutdown()` where it will not be blocking

PR #841, opened 2025-04-14
feat(BPAAS-2064): add gosec linter

shotasilagadzetaal:
## Description of Changes

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/internal/api/handler/default.go
+++ b/internal/api/handler/default.go
@@ -146,7 +146,17 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) (err error) {
 	}()
 
 	satoshis, bytes := calcFeesFromBSVPerKB(m.NodePolicy.MinMiningTxFee)
-	maxscriptsizepolicy, err := iapi.SafeUint64(m.NodePolicy.MaxScriptSizePolicy)
+	maxscriptsizepolicy, err := iapi.SafeIntToUint64(m.NodePolicy.MaxScriptSizePolicy)
+	if err != nil {
+		return err
+	}
+
+	maxTxSigopsCountsPolicy, err := iapi.SafeInt64ToUint64(m.NodePolicy.MaxTxSigopsCountsPolicy)
+	if err != nil {
+		return err
+	}
+
+	maxTxSizePolicy, err := iapi.SafeIntToUint64(m.NodePolicy.MaxTxSizePolicy)
 	if err != nil {
 		return err
 	}
@@ -154,8 +164,8 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) (err error) {
 
 		Policy: api.Policy{
 			Maxscriptsizepolicy:     maxscriptsizepolicy,
-			Maxtxsigopscountspolicy: uint64(m.NodePolicy.MaxTxSigopsCountsPolicy),
-			Maxtxsizepolicy:         uint64(m.NodePolicy.MaxTxSizePolicy),
+			Maxtxsigopscountspolicy: maxTxSigopsCountsPolicy,
+			Maxtxsizepolicy:         maxTxSizePolicy,
 			MiningFee: api.FeeAmount{
 				Bytes:    bytes,
 				Satoshis: satoshis,
--- a/internal/api/handler/default.go
+++ b/internal/api/handler/default.go
@@ -146,6 +146,9 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) (err error) {
 
 	satoshis, bytes := calcFeesFromBSVPerKB(m.NodePolicy.MinMiningTxFee)
 
+	if m.NodePolicy.MaxScriptSizePolicy < 0 {
+		return fmt.Errorf("invalid MaxScriptSizePolicy: must be non-negative")
+	}
 	return ctx.JSON(http.StatusOK, api.PolicyResponse{
 		Policy: api.Policy{
 			Maxscriptsizepolicy:     uint64(m.NodePolicy.MaxScriptSizePolicy),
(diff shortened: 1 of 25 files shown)


internal/api/helpers.go
```diff
@@ -1,10 +1,62 @@
...
+	}
+	return int64(u), nil
+}
+
+func SafeIntToInt32(i int) (int32, error) {
+	if i < math.MinInt32 || i > math.MaxInt32 {
+		return 0, fmt.Errorf("value out of range for int32: %d", i)
+	}
+	return int32(i), nil
 }
```
boecklim, line 62: Not sure if you've seen it in the ticket, but there is a package which provides safe casting: "github.com/ccoveille/go-safecast"

shotasilagadzetaal: I have seen those but do we need to use that library just for those functions? 🤔 

Even now I can replace them with `go-safecast` easily

boecklim: That library is specialized in doing exactly this. I think it is OK to import it just for that purpose

shotasilagadzetaal: Do you want me to use library for these?

internal/api/helpers.go
```diff
@@ -2,9 +2,14 @@ package api
...
 	feemodel "github.com/bsv-blockchain/go-sdk/transaction/fee_model"
+	"github.com/ccoveille/go-safecast"
 )
 
 func FeesToFeeModel(minMiningFee float64) *feemodel.SatoshisPerKilobyte {
 	satoshisPerKB := int(minMiningFee * 1e8)
-	return &feemodel.SatoshisPerKilobyte{Satoshis: uint64(satoshisPerKB)}
+	satoshisPerKBuint64, err := safecast.ToUint64(satoshisPerKB)
+	if err != nil {
+		return nil
```
boecklim, line 12: In this case we should return an error

internal/broadcaster/utxo_creator.go
```diff
@@ -246,14 +263,25 @@ func (b *UTXOCreator) splitToFundingKeyset(tx *sdkTx.Transaction, splitSatoshis,
...
 		fee, err = ComputeFee(tx, b.feeModel)
 		if err != nil {
 			return 0, err
 		}
-		if uint64(remaining)-requestedSatoshis < fee {
+		remainingUint64, err := safecast.ToUint64(remaining)
+		if err != nil {
+			return 0, err
+		}
+		if remainingUint64-requestedSatoshis < fee {
```
boecklim, line 284: I think the easier fix here would be to keep all numbers as uint64

```go
	remaining := splitSatoshis
	
	for remaining > requestedSatoshis && counter < requestedOutputs {
		fee, err = ComputeFee(tx, b.feeModel)
		if err != nil {
			return 0, err
		}
		if remaining-requestedSatoshis < fee {
			break
		}

		err = PayTo(tx, fundingKeySet.Script, requestedSatoshis)
		if err != nil {
			return 0, errors.Join(ErrFailedToAddOutput, err)
		}

		remaining -= requestedSatoshis
		counter++
	}

	fee, err = ComputeFee(tx, b.feeModel)
	if err != nil {
		return 0, err
	}

	err = PayTo(tx, fundingKeySet.Script, remaining-fee)
	if err != nil {
		return 0, errors.Join(ErrFailedToAddOutput, err)
	}
```

internal/callbacker/store/postgresql/postgres_test.go
```diff
@@ -390,8 +391,12 @@ func ReadAllCallbacks(t *testing.T, db *sql.DB) []*store.CallbackData {
 		if bh.Valid {
 			c.BlockHash = &bh.String
 		}
+		bheightUint64, err := safecast.ToUint64(bheight.Int64)
+		if err != nil {
+			t.Fatal(err)
+		}
```
boecklim, line 397: require.NoError(t, err)

internal/api/handler/default.go
```diff
@@ -145,12 +146,26 @@ func (m ArcDefaultHandler) GETPolicy(ctx echo.Context) (err error) {
...
+
+	maxTxSigopsCountsPolicy, err := safecast.ToUint64(m.NodePolicy.MaxTxSigopsCountsPolicy)
+	if err != nil {
+		return err
+	}
 
+	maxTxSizePolicy, err := safecast.ToUint64(m.NodePolicy.MaxTxSizePolicy)
+	if err != nil {
+		return err
+	}
```
boecklim, line 162: If these conversions were done in the [constructor](https://github.com/bitcoin-sv/arc/blob/d54984d8425cc6283e85038aed28dba1ae3ec6e8/internal/api/handler/default.go#L113) then the they have to be done just once in the beginning. The vars can be stored in `ArcDefaultHandler`

internal/beef/beef.go
```diff
@@ -147,9 +148,13 @@ func decodeTransactionsWithPathIndexes(beefBytes []byte) ([]*TxData, []byte, err
...
 
-	transactions := make([]*TxData, 0, int(nTransactions))
+	ntxs, err := safecast.ToInt(nTransactions)
+	if err != nil {
+		return nil, nil, err
+	}
+	transactions := make([]*TxData, 0, ntxs)
 
-	for i := 0; i < int(nTransactions); i++ {
+	for i := 0; i < ntxs; i++ {
```
boecklim, line 157: This cast can be avoided like this

```go
	beefBytes = beefBytes[bytesUsed:]

	transactions := make([]*TxData, 0, nTransactions)

	for i := uint64(0); i < uint64(nTransactions); i++ {
		tx, bytesUsed, err := sdkTx.NewTransactionFromStream(beefBytes)
		if err != nil {
			return nil, nil, err
		}
```

boecklim (approved):

PR #844, opened 2025-04-15
test(BPAAS-2070): additional tests for rate_broadcaster

tngdomingo:
new opts added for rate_broadcaster.go tests

## Description of Changes

## Testing Procedure

- [X] All tests pass locally

## Checklist:


pkg/message_queue/nats/client/nats_jetstream/nats_jetstream_client_test.go
boecklim, line 1: It looks like this file is just copied from here `pkg/message_queue/nats/client/nats_jetstream/integration_test/client_integration_test.go` and therefore doesn't add any value

The reason that the integration test is not counted in the test coverage may be that it is not in the same package.

This PR should solve that issue and this additional test file should not be necessary: https://github.com/bitcoin-sv/arc/pull/846

tngdomingo: That makes sense. I will remove that and leave only the new tests for rate_broadcaster

boecklim (approved):

PR #852, opened 2025-04-22
feat(BPAAS-2078): RPC bitcoin client with context

boecklim:
## Description of Changes

- RPC bitcoin client with context

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/node_client/rpc_client_test.go
+++ b/internal/node_client/rpc_client_test.go
@@ -1,0 +1,37 @@
+package node_client_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/bitcoin-sv/arc/internal/node_client"
+)
+
+func TestRPCClient(t *testing.T) {
+	t.Helper()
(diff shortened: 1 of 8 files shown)


internal/node_client/rpc_client_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/bitcoin-sv/arc/internal/node_client"
+)
+
+func TestRPCClient(t *testing.T) {
+	t.Helper()
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	ctx := context.Background()
```
shotasilagadzetaal, line 18: Want's the main idea to check if ctx worked? so cancelling it should return the corresponding error?

shotasilagadzetaal (approved):

internal/node_client/rpc_client_test.go
boecklim: Sorry, don't get the question.
Are you suggesting to test cancelling the context? That would make sense

PR #856, opened 2025-04-22
feat: Rebroadcast seen txs

boecklim:
## Description of Changes

- Periodically re-announce transactions which were seen on network, but haven't been mined for a specified time duration since the last transaction was marked as mined
- Refactor and rename functions and variables for improved understandability 

## Testing Procedure

- [X] All tests pass locally

## Checklist:


internal/metamorph/processor.go
```diff
@@ -666,7 +670,68 @@ func (p *Processor) StartRequestingSeenOnNetworkTxs() {
...
+			case <-ticker.C:
+				ctx, span := tracing.StartTracing(p.ctx, "StartReAnnounceSeenTxs", p.tracingEnabled, p.tracingAttributes...)
+
+				var offset int64
+				var totalSeenOnNetworkTxs int
+				var seenOnNetworkTxs []*store.Data
+				var err error
+
+				for {
+					seenOnNetworkTxs, err = p.store.GetSeenSinceLastMined(ctx, p.rebroadcastExpiration, p.reAnnounceSeen, loadSeenOnNetworkLimit, offset)
```
shotasilagadzetaal, line 707: I don't understand this part, why are we considering other mined transaction here. Maybe for some reason problematic transaction was rejected from the mempool but other transactions get mined without problem.

In that case will we be re-broadcasting `SeenOnNetwork` txs?

shotasilagadzetaal (approved):

internal/metamorph/processor.go
boecklim: The idea was that we should only re announce the already seen on network transaction if there had been a transaction in the meantime which was mined, which indicates that there was now enough time for those still seen on network to be mined.

PR #863, opened 2025-05-02
fix(BPAAS-2094): Fix flaky unit test

boecklim:
## Description of Changes

- Fix flaky unit test `TestRateBroadcasterStartAsync`
- Log each updated orphaned block

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/broadcaster/rate_broadcaster_test.go
+++ b/internal/broadcaster/rate_broadcaster_test.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"log/slog"
 	"os"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/broadcaster/rate_broadcaster_test.go
```diff
@@ -196,39 +185,34 @@ func TestRateBroadcasterStart(t *testing.T) {
...
-		jitterSize               int64
+		name             string
+		limit            int64
+		rateTxsPerSecond int
+		batchSize        int
+		expectedTxCount  int64
+		jitterSize       int64
 
 		expectedBroadcastTransactionsCalls int
 		expectedError                      error
```
tngdomingo, line 201: I think now we can get rid of expectedError, right?

tngdomingo (approved):

internal/broadcaster/rate_broadcaster_test.go
boecklim: Yes, you're right. Will remove later.

PR #868, opened 2025-05-07
feat(BPAAS-2079): Unorphan recent wrongly orphaned blocks

tngdomingo:
## Description of Changes

Added a retry for the handleSkippedBlock and a new AutoHeal job for the background_workers.go
This should ensure that orphan blocks that are not marked as LONGEST during heavy DB usage periods are retried or at least auto-healed a few minutes later

## Testing Procedure

- [X] All tests pass locally

## Checklist:


internal/blocktx/store/postgresql/auto_heal_orphans.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/bitcoin-sv/arc/internal/blocktx/blocktx_api"
+)
+
+// AutoHealOrphans searches for the LONGEST blocks in the last 1000
+// then recursively searches for blocks marked as ORPHANED
+// whose hash matches the prevhash of the given block,
+// and then repeats that recursively for each newly found orphaned block until
+// it has the entire orphaned chain.
+func (p *PostgreSQL) AutoHealOrphans(ctx context.Context) (healedOrphans []*blocktx_api.Block, err error) {
```
boecklim, line 14: The name of this function is misleading as it suggests that this function does "heal" something, but it actually just "searches for blocks marked as ORPHANED whose hash matches the prevhash of the given block".

There is no update to any row, so maybe the title could be called "FindUnorphanedOrphans" or something like that.

internal/blocktx/background_workers.go
```diff
@@ -100,3 +100,17 @@ func (w *BackgroundWorkers) fillGaps(peer p2p.PeerI, retentionDays int, blockReq
...
+
+func (w *BackgroundWorkers) StartAutoHeal() {
+	w.workersWg.Add(1)
+	go func() {
+		_ = func() error {
+			defer w.workersWg.Done()
+			_, err := w.store.AutoHealOrphans(w.ctx)
+			if err != nil {
+				w.logger.Error("failed to auto heal orphans", slog.String("err", err.Error()))
+			}
```
boecklim, line 112: It seems this functionality just finds the orphans but doesn't update it as the result here is ignored

internal/blocktx/store/postgresql/auto_heal_orphans.go
```diff
@@ -0,0 +1,74 @@
...
+		SELECT 	hash
+				,prevhash
+				,merkleroot
+				,height
+				,processed_at
+				,status
+				,chainwork
+		FROM recent_orphans
+		WHERE status = $2
+		ORDER BY height
```
boecklim, line 59: This function could actually be updating the block status if this were an `UPDATE ...` statement

Event then though, I wouldn't call the function `AutoHealing...` but rather `UpdateUnorphanedOrphans` or something like that.

internal/blocktx/store/postgresql/auto_heal_orphans.go
tngdomingo: Done

internal/blocktx/store/postgresql/auto_heal_orphans.go
tngdomingo: Done

internal/blocktx/background_workers.go
tngdomingo: Done

internal/blocktx/background_workers.go
```diff
@@ -100,3 +100,17 @@ func (w *BackgroundWorkers) fillGaps(peer p2p.PeerI, retentionDays int, blockReq
...
+	w.workersWg.Add(1)
+	go func() {
+		_ = func() error {
+			defer w.workersWg.Done()
+			_, err := w.store.UnorphanRecentWrongOrphans(w.ctx)
+			if err != nil {
+				w.logger.Error("failed to auto heal orphans", slog.String("err", err.Error()))
+			}
+			return err
+		}()
```
boecklim, line 114: This will only run once. The interval defined in the config is not used here.

This function should be setup similarly to the fill gaps function: https://github.com/bitcoin-sv/arc/blob/bdcd464b01c4b91cd4c35246eae0bc60adc38d02/internal/blocktx/background_workers.go#L44-L70

config/config.go
```diff
@@ -179,6 +180,10 @@ type FillGapsConfig struct {
 	Interval time.Duration `mapstructure:"interval"`
 }
 
+type AutoHealConfig struct {
+	Enabled  bool          `mapstructure:"enabled"`
+	Interval time.Duration `mapstructure:"interval"`
+}
```
boecklim, line 186: This would need to be added to https://github.com/bitcoin-sv/arc/blob/c14ece2ee7f8817b6d1f886d24af60605cad5d03/config/example_config.yaml as well.

And also the test configs for consistency

https://github.com/bitcoin-sv/arc/blob/08ccfe0e7972c0d0cfa998455de5c38404a8e348/test/config/config.yaml
https://github.com/bitcoin-sv/arc/blob/08ccfe0e7972c0d0cfa998455de5c38404a8e348/test/config/config_mcast.yaml

config/defaults.go
```diff
@@ -196,6 +197,13 @@ func getFillGapsConfig() *FillGapsConfig {
 	}
 }
 
+func getAutoHealConfig() *AutoHealConfig {
```
boecklim, line 200: I think `FixOrphanBlocksConfig` would be better here as a name

internal/blocktx/store/postgresql/postgres_test.go
```diff
@@ -366,6 +366,33 @@ func TestPostgresDB(t *testing.T) {
...
+		}
+
+		// when
+		actualHealedOrphans, err := postgresDB.UnorphanRecentWrongOrphans(ctx)
+		require.NoError(t, err)
+
+		// then
+		require.Equal(t, len(expectedHealedOrphans), len(actualHealedOrphans))
+		for i, b := range actualHealedOrphans {
+			require.Equal(t, expectedHealedOrphans[i][:], b.Hash)
```
boecklim, line 392: In order to be sure the function actually updated the rows, just counting the rows might not be enough.

It would be safer to check if the correct changes have been done. Here is an example of how that can be achieved from another test

https://github.com/bitcoin-sv/arc/blob/574a6a69beb279d1af38d467158e4069627c9335/internal/blocktx/store/postgresql/postgres_test.go#L802-L812

internal/blocktx/store/postgresql/unorphan_recent_wrong_orphans.go
```diff
@@ -0,0 +1,91 @@
...
+           , status
+      FROM recent_orphans
+      WHERE status = $2
+) as btu;
+`
+	qUpdateUnorphanBlocks := `
+	UPDATE blocktx.blocks b
+	SET status = $1
+	FROM blocks_to_unorphan
+	WHERE blocks_to_unorphan.hash = b.hash;
```
boecklim, line 53: I think instead of adding an additional query, you could add here a `RETURNING b.hash` which will return all block hashes that have been effectively updated. Then you may not need the additional `qSelectUpdatedBlocks` query

internal/blocktx/background_workers.go
```diff
@@ -100,3 +100,17 @@ func (w *BackgroundWorkers) fillGaps(peer p2p.PeerI, retentionDays int, blockReq
 
 	return nil
 }
+
+func (w *BackgroundWorkers) StartAutoHeal() {
```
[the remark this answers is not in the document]
tngdomingo: I kept this because fixing the orphan blocks is just the first action but there could be more in the future and StartAutoHeal could trigger
- StartFixOrphanBlocks
- StartFixXYZ, anything that is manually done now but could be automated in the future

config/defaults.go
tngdomingo: Same as above. If the AutoHeal feature is to be enhanced in the future with further functionalities, this might be a good name. On the other hand, as now there is only one, yes, we can rename it to better describe this particular one.

internal/blocktx/store/postgresql/postgres_test.go
tngdomingo: It is actually checked, the function returns the rows for:
rows, err := p.db.QueryContext(ctx, qSelectUpdatedBlocks, blocktx_api.Status_LONGEST)
which retrieves the blocks that were initially orphaned and now are LONGEST. 
Then the hashes of the returned (therefore unorphaned) are checked to verify that those 4 are actually the 4 we are expecting and no others.

internal/blocktx/store/postgresql/unorphan_recent_wrong_orphans.go
tngdomingo: Nice! I didn't know about that one.

internal/blocktx/store/postgresql/unorphan_recent_wrong_orphans.go
```diff
@@ -0,0 +1,91 @@
...
+	//
+	// Then entire recursion happens in the second SELECT, after UNION ALL,
+	// and the first SELECT is just to set up the `recent_orphans` variable with
+	// the first, initial value. Then, the `recent_orphans` variable is recursively
+	// updated with values returned from the second SELECT.
+	qDropTempTable := `drop table if exists tmp_blocks_to_unorphan;`
+	qCreateTempTable := `
+create table blocks_to_unorphan AS
+SELECT btu.hash, btu.prevhash, btu.status FROM (
+    WITH RECURSIVE recent_orphans AS (SELECT *
```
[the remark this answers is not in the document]
tngdomingo: Thanks, will try that.

config/config.go
tngdomingo: Done

config/example_config.yaml
```diff
@@ -111,6 +111,9 @@ blocktx:
   fillGaps:
     enabled: true
     interval: 15m # time interval to check and fill gaps in processed blocks
+  unorphanRecentWrongOrphans:
```
boecklim, line 114: I think the "recent" could be omitted as the functionality checks all blocks that are in the table.

I'd rename it to `unorphanWrongOrphans` and accordingly all other functions

internal/blocktx/background_workers.go
```diff
@@ -100,3 +100,33 @@ func (w *BackgroundWorkers) fillGaps(peer p2p.PeerI, retentionDays int, blockReq
...
+				rows, err := w.store.UnorphanRecentWrongOrphans(w.ctx)
+				if err != nil {
+					w.logger.Error("failed to unorphan recent wrong orphans", slog.String("err", err.Error()))
+				}
+				for _, b := range rows {
+					w.logger.Info("Successfully unorphaned block", slog.Uint64("height", b.Height))
+				}
+
+				i++
+				ticker.Reset(interval)
```
boecklim, line 125: I think that is not needed here

config/example_config.yaml
tngdomingo: It only checks for the last 1000 blocks. That's hardcoded in the query. We can do it for all of them but just to speed the query up.

internal/blocktx/background_workers_test.go
```diff
@@ -108,3 +109,46 @@ func TestStartFillGaps(t *testing.T) {
...
+					Height: 822015,
+				},
+			},
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			// given
+			const fillUnorphanRecentWrongOrphansInterval = 50 * time.Millisecond
```
boecklim, line 134: The "fill" here can be removed

internal/blocktx/background_workers.go
```diff
@@ -100,3 +100,33 @@ func (w *BackgroundWorkers) fillGaps(peer p2p.PeerI, retentionDays int, blockReq
...
 }
+
+func (w *BackgroundWorkers) StartUnorphanRecentWrongOrphans(interval time.Duration) {
+	w.workersWg.Add(1)
+
+	go func() {
+		defer w.workersWg.Done()
+
+		ticker := time.NewTicker(interval)
+		i := 0
```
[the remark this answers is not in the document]
tngdomingo: That's a leftover. Removed.

internal/blocktx/background_workers.go
tngdomingo: Removed.

internal/blocktx/background_workers_test.go
tngdomingo: Done.

internal/blocktx/background_workers.go
```diff
@@ -100,3 +100,33 @@ func (w *BackgroundWorkers) fillGaps(peer p2p.PeerI, retentionDays int, blockReq
...
+
+		for {
+			select {
+			case <-ticker.C:
+				rows, err := w.store.UnorphanRecentWrongOrphans(w.ctx)
+				if err != nil {
+					w.logger.Error("failed to unorphan recent wrong orphans", slog.String("err", err.Error()))
+				}
+				for _, b := range rows {
+					w.logger.Info("Successfully unorphaned block", slog.Uint64("height", b.Height))
```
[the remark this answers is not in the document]
tngdomingo: Done.

boecklim (approved):

PR #870, opened 2025-05-09
fix(BPAAS-2073): Request pending seen transactions

boecklim:
## Description of Changes

Transactions can get stuck in SEEN_ON_NETWORK status. Re-announcing does not always solve the issue. With this change transactions which are in SEEN_ON_NETWORK status for a certain time (pending) additionally will get re-requested. 

A separate table `requested_transactions` tracks whether the re-requested transactions get confirmed by any connected node and thereby proving that there is at least one connected node which has this transaction in the mempool.

All other transactions which are not getting confirmed by sending the transaction, can thus not be found in any mempool and will transition to REJECTED status after a specified amount of time. 

:warning: The status update to REJECTED is temporarily deactivated until the functionality is tested by way of analysing the logs

## Testing Procedure

- [X] All tests pass locally

## Checklist:


internal/metamorph/store/postgresql/migrations/000020_requested_transactions.up.sql
```diff
@@ -0,0 +1,7 @@
+
+
+CREATE TABLE metamorph.requested_transactions (
```
shotasilagadzetaal, line 3: Wouldn't it better to simply have two new fields in existing `transactions` table? 🤔

internal/metamorph/processor.go
```diff
@@ -429,6 +430,13 @@ func (p *Processor) StartSendStatusUpdate() {
 				return
 
 			case msg := <-p.statusMessageCh:
+				if msg.ReceivedRawTx {
+					err := p.store.MarkConfirmedRequested(p.ctx, msg.Hash)
```
shotasilagadzetaal, line 434: shouldn't we simply remove those transactions from db instead of "confirming" them? 🤔 so we will not need to clean that up later.

internal/metamorph/processor_routines.go
```diff
@@ -70,36 +71,102 @@ func ReAnnounceUnseen(ctx context.Context, p *Processor) []attribute.KeyValue {
...
-// ReAnnounceSeen re-broadcasts SEEN_ON_NETWORK transactions
+// RejectUnconfirmedRequested re-broadcasts SEEN_ON_NETWORK transactions
+func RejectUnconfirmedRequested(ctx context.Context, p *Processor) []attribute.KeyValue {
+	var offset int64
+	var totalRejected int
+	var txHashes []*chainhash.Hash
+	var err error
+
+	// first delete all requested transactions which have transitioned to another status
+	rows, err := p.store.DeleteConfirmedRequested(ctx)
```
shotasilagadzetaal, line 82: This line will run once 🤔 shouldn't we have this in for loop?

boecklim: This can run just once because it removes all those entries in the `requested_transactions` where the tx status has changed to another one thatn SEEN_ON_NETWORK

shotasilagadzetaal (approved):

PR #875, opened 2025-05-13
chore(BPAAS-2118): Reducing complexity based on Sonarcloud feedback

tngdomingo:
## Description of Changes

Reducing complexity of:
- parsers.go 
- blocktx/processor.go
- broadcaster/api.go 
- blocktx/.../verify_merkle_roots.go

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/api/handler/parsers_test.go
+++ b/internal/api/handler/parsers_test.go
@@ -17,7 +17,8 @@ func TestParseTransactionFromRequest(t *testing.T) {
 		tx      string
 		txBytes []byte
 
-		expectedHex []byte
+		expectedHex        []byte
+		invalidContentType bool
 	}{
 		{
 			name:        "valid Raw Transaction",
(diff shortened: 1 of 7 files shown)


internal/api/handler/parsers_test.go
```diff
@@ -33,31 +34,43 @@ func TestParseTransactionFromRequest(t *testing.T) {
...
 	for _, tc := range testCases {
 		var requests []*http.Request
 
-		for _, ct := range contentTypes {
+		if tc.invalidContentType {
 			var r *http.Request
+			body := fmt.Sprintf("{\"rawTx\": \"%s\"}", tc.tx)
+			r, _ = http.NewRequest("GET", "", bytes.NewBuffer([]byte(body)))
+			r.Header.Set(echo.HeaderContentType, echo.MIMEMultipartForm)
+			requests = append(requests, r)
```
boecklim, line 52: by putting here a `return` you could remove the `else` clause which I usually find easier to read

internal/broadcaster/rate_broadcaster.go
```diff
@@ -118,33 +118,36 @@ func (b *UTXORateBroadcaster) Start() error {
...
-
-				if b.limit > 0 && atomic.LoadInt64(&b.totalTxs) >= b.limit {
-					b.logger.Info("limit reached", slog.Int64("total", atomic.LoadInt64(&b.totalTxs)), slog.Int64("limit", b.limit))
-					b.shutdown <- struct{}{}
-				}
+	return nil
+}
 
-				b.broadcastBatchAsync(txs, errCh, b.waitForStatus)
+func checkTicker(b *UTXORateBroadcaster, tickerCh <-chan time.Time, errCh chan error) {
```
boecklim, line 127: nit: Maybe a more telling name for this function would be 

```go
(b *UTXORateBroadcaster)startBroadcastSelfPayingTxs(tickerCh <-chan time.Time, errCh chan error)
```
or something like that

Also I'd add the creation of the go routine in the function as well

```go
        b.wg.Add(1)
	go func() {
		defer func() {
			b.logger.Info("shutting down broadcaster")
			b.wg.Done()
		}()
...
```

internal/broadcaster/rate_broadcaster.go
```diff
@@ -263,6 +222,58 @@ utxoLoop:
 	return txs, nil
 }
 
+func addAdditionalInputs(b *UTXORateBroadcaster, tx **sdkTx.Transaction, amount *uint64) error {
```
boecklim, line 225: nit: Alternative name suggestion: `addRandomDataInputs(...)`

tngdomingo: Done

internal/broadcaster/rate_broadcaster.go
tngdomingo: Done

internal/api/handler/parsers_test.go
tngdomingo: But in this case there is some logic afterwards so the return would skip that. But good suggestion, thanks!. I will keep it in mind for future ifs

PR #883, opened 2025-05-27
chore(BPAAS-2123): Review SonarCloud complexities

tngdomingo:
## Description of Changes

Small refactors on the functions with the highest punctuations according to SonarCloud metrics

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/p2p/peer.go
+++ b/internal/p2p/peer.go
@@ -255,8 +255,14 @@ func (p *Peer) handshake(c net.Conn) (ok bool) {
 		}
 	}(handshakeReadCtx)
 
+	return p.performHandShake(c, read, readController, handshakeFailed)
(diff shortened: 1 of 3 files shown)


internal/p2p/peer.go
```diff
@@ -310,17 +310,14 @@ handshakeLoop:
...
 
-				if receivedVerAck {
-					break handshakeLoop
-				}
-
 			default:
 				p.logger.Warn("Handshake: received unexpected message. Message was ignored", slogUpperString(commandKey, nmsg.Command()))
 			}
 		}
+		if receivedVerAck && sentVerAck {
```
boecklim, line 317: That should be rather an or `||` instead of an and `&&` I think

boecklim: Yes, you may be right

PR #884, opened 2025-05-28
chore(BPAAS-2123): Review SonarCloud complexities 

tngdomingo:
## Description of Changes

Small refactors on the functions with the highest punctuations according to SonarCloud metrics

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/internal/callbacker/send_manager/send_manager.go
+++ b/internal/callbacker/send_manager/send_manager.go
@@ -174,7 +174,101 @@ func (m *SendManager) Start() {
 	var callbackBatch []*callbacker.CallbackEntry
 
 	go func() {
-		m.sendCallbacks(callbackBatch, queueTicker, sortQueueTicker, backfillQueueTicker, batchSendTicker)
+		var err error
+		defer func() {
+			m.storeRemainingCallbacks(callbackBatch)
+		}()
+		lastIterationWasBatch := false
+
+		for {
+			const queueLength = "queue length"
+			const failedToSendCallbacks = "Failed to send batch of callbacks"
+			const callbackElements = "callback elements"
+			select {
+			case <-m.ctx.Done():
+				return
+			case <-sortQueueTicker.C:
+				m.mu.Lock()
+				sort.Slice(m.callbackQueue, func(i, j int) bool {
+					return m.callbackQueue[j].Data.Timestamp.After(m.callbackQueue[i].Data.Timestamp)
+				})
+				m.mu.Unlock()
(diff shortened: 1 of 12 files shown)


pkg/message_queue/nats/nats_connection/nats_connection.go
```diff
@@ -145,3 +135,16 @@ func New(natsURL string, logger *slog.Logger, opts ...Option) (*nats.Conn, error
...
+	if len(cfg.clientClosedCh) > 0 {
+		for _, ch := range cfg.clientClosedCh {
+			select {
+			case ch <- struct{}{}:
+			default:
+			}
+		}
+	}
+	return nil
+}
```
boecklim, line 150: I think this is not correct. Here a function needs to be returned and the logic should happen in that returned function. That is the logic to be executed when the closed handler event happens, not when the the options are set

```go
func closeHandler(logger *slog.Logger, cfg *NatsConfig) nats.ConnHandler {
	return func(_ *nats.Conn) {
		logger.Warn("client closed")
		if len(cfg.clientClosedCh) > 0 {
			for _, ch := range cfg.clientClosedCh {
				select {
				case ch <- struct{}{}:
				default:
				}
			}
		}
	}
}

```

tngdomingo: Thanks I get it now. In these cases the anonymous functions cannot be externalized to lower the complexity 👍

internal/callbacker/send_manager/send_manager.go
```diff
@@ -175,101 +174,7 @@ func (m *SendManager) Start() {
...
-				success, retry := m.sender.Send(m.url, callbackEntry.Token, callbackEntry.Data)
-				if !retry || success {
-					m.callbackQueue = m.callbackQueue[1:]
-					m.logger.Debug("Single callback sent", slog.Int(callbackElements, len(callbackBatch)), slog.Int(queueLength, m.CallbacksQueued()), slog.String("url", m.url))
-					continue
-				}
-				m.logger.Warn("Failed to send single callback", slog.String("url", m.url))
-			}
-		}
+		m.sendCallbacks(callbackBatch, queueTicker, sortQueueTicker, backfillQueueTicker, batchSendTicker)
```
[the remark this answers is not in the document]
tngdomingo: Yeah, sorry for that. Now should be ok.

boecklim (approved):

PR #893, opened 2025-06-04
feat(BPAAS-2127): fix stale tx status

shotasilagadzetaal:
## Description of Changes

The problem in the test was the following: we expected that the transaction belonging to the stale chain would remain there. The thing is that a transaction can be part of the both chains - canonical and stale. If it still exists in the pool or is re-broadcasted it can also be mined in a forked chain. 

That's what was happening sometimes - tx from stale block was also included in the canonical chain.

In the end we somehow expected that it would remain in stale chain which is incorrect. Removing that assumption from the test.

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

### DIFF
--- a/test/submit_05_reorg_test.go
+++ b/test/submit_05_reorg_test.go
@@ -189,6 +189,12 @@ func TestReorg(t *testing.T) {
 			switch status.Txid {
 			// verify that callback for tx2 was received
 			case tx2.TxID().String():
+				if status.TxStatus == StatusMined {
+					require.Equal(t, staleHash, *status.BlockHash)
+				}
+				if status.TxStatus == StatusMinedInStaleBlock {
+					require.Equal(t, tx2BlockHash, *status.BlockHash)
+				}
 			// verify that callback for tx1 was received with status MINED and updated merkle path
 			case tx1.TxID().String():
 				require.Equal(t, StatusMined, status.TxStatus)


test/submit_05_reorg_test.go
```diff
@@ -180,20 +180,15 @@ func TestReorg(t *testing.T) {
...
-
 	merklePathTx1 = getMerklePath(t, tx1.TxID().String())
 
 	// expect 2 callbacks
 	for range 2 {
 		select {
 		case status := <-callbackReceivedChan:
 			switch status.Txid {
-			// verify that callback for tx2 was received with status MINED_IN_STALE_BLOCK
+			// verify that callback for tx2 was received
```
boecklim, line 190: Actually nothing is now verified in case of `tx2.TxID().String()`

test/submit_05_reorg_test.go
```diff
@@ -180,20 +180,15 @@ func TestReorg(t *testing.T) {
 	// verify that previous mined tx1 have updated block info
 	checkStatusBlockHash(t, tx1.TxID().String(), StatusMined, staleHash)
 
-	// verify that tx2 is now MINED_IN_STALE_BLOCK
-	checkStatusBlockHash(t, tx2.TxID().String(), StatusMinedInStaleBlock, tx2BlockHash)
```
boecklim, line 184: In this case that you wrote here
> That's what was happening sometimes - tx from stale block was also included in the canonical chain.
then we should check here that the status of `tx2` is either `StatusMinedInStaleBlock` or `StatusMined`. 

If it was status `StatusMined`, then the block hash would need to be not equal to `tx2BlockHash` but the new block hash.

If it was status `StatusMinedInStaleBlock`, then we should also expect and verify that the corresponding callback was received.
Wdyt?

boecklim: We shouldn't remove the check completely here but check that status of tx2 is either `StatusMinedInStaleBlock` or `StatusMined` with corresponding block hash

Also 
1. if status of tx2 was `StatusMinedInStaleBlock` => check that there is 1 callback with status `StatusMinedInStaleBlock`
2. if status of tx2 was `StatusMined` => check that there is 1 callback with status `StatusMined`

Not?

PR #897, opened 2025-06-12
chore(BPAAS-2141): Change deprecated woc api

tngdomingo:
## Description of Changes

Migrate woc_client endpoints to use the latest (not deprecated) versions of UTXOs and Balance API calls.

## Testing Procedure

- [X] All tests pass locally

## Checklist:

### DIFF
--- a/pkg/woc_client/woc_client_integration_test.go
+++ b/pkg/woc_client/woc_client_integration_test.go
@@ -1,0 +1,85 @@
+//go:build woc
+
+package woc_client
+
+import (
+	"context"
+	sdkTx "github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func Test_GetBalanceFromWoC(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	tt := []struct {
+		name          string
+		address       string
+		responseOk    bool
+		responseBody  any
+		expectedError error
+		expected      sdkTx.UTXOs
+	}{
+		{
+			name:       "response OK",
+			address:    testnetAddr,
+			responseOk: true,
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			// given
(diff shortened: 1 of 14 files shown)


pkg/woc_client/woc_client_test.go
```diff
@@ -172,12 +201,69 @@ func Test_GetUTXOs(t *testing.T) {
...
+
+				jsonResp, err := json.Marshal(tc.responseBody)
+				require.NoError(t, err)
+				_, err = w.Write(jsonResp)
+				require.NoError(t, err)
+			}))
+			defer svr.Close()
+
+			// given
+			sut := New(false, WithURL("https://api.whatsonchain.com/v1/bsv/"))
```
boecklim, line 251: I think here the URL has to be from the httptest server

```go
sut := New(false, WithURL(svr.URL))
```

Other wise the request is sent to the actual WoC API

test/submit_06_woc_connection_test.go
boecklim, line 1: IMO the e2e tests are not the right place for this test, because this is meant to test the whole system end to end.

Also this would run each time in the pipeline and be dependent on the external API of WoC.

Because of the dependency of an external API running this test should be opt-in.

The test could be in `pkg/woc_client/woc_client_integration_test.go` for instance with a build tag `//go:build woc`

Then this test can be run optionally with like this `go test -v --tags=woc ./pkg/woc_client/...`

tngdomingo: Done

pkg/woc_client/woc_client_integration_test.go
```diff
@@ -0,0 +1,92 @@
...
+			name:       "response OK",
+			address:    testnetAddr,
+			responseOk: true,
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			svr := httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, _ *http.Request) {}))
+			defer svr.Close()
```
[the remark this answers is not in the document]
tngdomingo: 👍 Done.

Issue #900, opened 2025-06-16
[BUG] Transaction stuck in status DOUBLE_SPEND_ATTEMPTED


boecklim:
## Bug Description

Transaction `c12d6cc2c4a5386ab4597d621ad552d877c677d9ced758a076dd174571a4eed9` was initially broadcasted to `https://arc.taal.com` and reached SEEN_ON_NETWORK, it went a couple of second later into double spend status. But never got any resolution later. It was successfully double spent though

This transaction has now status 

```
{
    "blockHash": "",
    "blockHeight": 0,
    "competingTxs": [
        "f8abc5b86d918bedbd43ce28ba7181018b80e42c559ed81e36841f7f2caddd9f"
    ],
    "extraInfo": "",
    "merklePath": "",
    "timestamp": "2025-06-16T14:04:42.501005446Z",
    "txStatus": "DOUBLE_SPEND_ATTEMPTED",
    "txid": "c12d6cc2c4a5386ab4597d621ad552d877c677d9ced758a076dd174571a4eed9"
}
```

## Steps to Reproduce

Not easily possible to reproduce, but the issue may be possible to be found in the logs

## Expected Behavior

The transaction should be rejected and transition to `REJECTED` status

## Actual Behavior

The transaction remained in status `DOUBLE_SPEND_ATTEMPTED`

boecklim:
ARC doesn’t know the other transaction. Therefore the tx just remains in DOUBLE_SPEND_ATTEMPTED status
```
curl --location 'https://arc.taal.com/v1/tx/f8abc5b86d918bedbd43ce28ba7181018b80e42c559ed81e36841f7f2caddd9f'
```
```json
{
    "detail": "The requested resource could not be found",
    "extraInfo": "transaction not found",
    "instance": null,
    "status": 404,
    "title": "Not found",
    "txid": null,
    "type": "https://bitcoin-sv.github.io/arc/#/errors?id=_404"
}
```
Currently this works only if ARC knows also all the competing tx.
One easy solution here may be:
1. Check in intervals if there are txs with DOUBLE_SPEND_ATTEMPTED status which have been stored longer than e.g. 10min ago
2. For each tx of those txs, take the competingTxs and check in a grpc request to blocktx if any of them have been mined in a longest block.
    1. If yes, we can mark the tx as REJECTED

PR #906, opened 2025-06-20
feat(BPAAS-2172): reject tx if competing tx is mined

shotasilagadzetaal:
## Description of Changes

we now reject transactions if their competing txs get mined. 

#900 

## Testing Procedure

- [x] All tests pass locally

## Checklist:

### DIFF
--- a/internal/blocktx/server_test.go
+++ b/internal/blocktx/server_test.go
@@ -93,7 +93,7 @@ func TestAnyTransactionsMined(t *testing.T) {
 				{Hash: []byte("tx1")},
 			}})
 			require.NoError(t, err)
-			require.Equal(t, res.Transactions[0].Mined, tc.isMined)
+			require.Equal(t, tc.isMined, res.Transactions[0].Mined)
 		})
 	}
 }


internal/blocktx/blocktx_api/blocktx_api.proto
```diff
@@ -29,6 +29,12 @@ service BlockTxAPI {
   // CurrentBlockHeight returns current block height
   rpc CurrentBlockHeight (google.protobuf.Empty) returns (CurrentBlockHeightResponse) {}
 
+  // IsCompetingTransactionMined returns true if any of the transactions is mined
+  rpc IsCompetingTransactionMined(CompetingTxs) returns (CompetingTxMined) {}
+}
+
+message CompetingTxMined {
```
boecklim, line 36: Accordingly to the other comment I'd name this `AnyTransactionsMinedResponse` as it is not blocktx's concern whether these transactions are competing.

internal/metamorph/store/postgresql/postgres.go
```diff
@@ -312,6 +312,40 @@ func (p *PostgreSQL) GetMany(ctx context.Context, keys [][]byte) (data []*store.
 	return getStoreDataFromRows(rows)
 }
 
+func (p *PostgreSQL) GetDoubleSpendTxs(ctx context.Context, older time.Duration) (data []*store.Data, err error) {
```
boecklim, line 315: The integration test for this function is missing

internal/metamorph/processor_helpers.go
```diff
@@ -274,3 +275,20 @@ func (p *Processor) StartRoutine(tickerInterval time.Duration, routine func(cont
...
+	if len(txs) == 0 {
+		return nil, nil
+	}
+
+	hashes := make([][]byte, 0, len(txs))
+	for _, tx := range txs {
+		hash, err := hex.DecodeString(tx)
+		if err != nil {
+			return nil, err
+		}
```
boecklim, line 289: I think if you would use here `chainhash.NewHashFromStr(...)` like this

```go
func txBytesFromHex(txs []string) ([][]byte, error) {
	if len(txs) == 0 {
		return nil, nil
	}

	hashes := make([][]byte, 0, len(txs))
	for _, tx := range txs {
		hash, err := chainhash.NewHashFromStr(tx)
		if err != nil {
			return nil, err
		}
		hashes = append(hashes, hash[:])
	}

	return hashes, nil
}

```

then you wouldn't have to reverse the byte array here: https://github.com/bitcoin-sv/arc/blob/50c9c8bc56b98b5a8ff189825babd45dcc03cff1/internal/blocktx/server.go#L120

shotasilagadzetaal: somehow didn't work like this

test/submit_06_double_spend_rejected_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+		// make sure tx1 is mined in the block
+		minedTx, err := client.GetRawTransactionVerbose(context.Background(), tx1ID)
+		require.NoError(t, err)
+		require.NotEqual(t, "", minedTx.BlockHash)
+
+		// check that we have double spend status
+		tx2StatusURL := fmt.Sprintf("%s/%s", arcEndpointV1Tx, tx2.TxID())
+		tx2StatusResp := getRequest[TransactionResponse](t, tx2StatusURL)
+		require.Equal(t, StatusDoubleSpendAttempted, tx2StatusResp.TxStatus)
```
boecklim, line 67: I think this is what we expect before generating a new block, not?

```
// mine the first tx
		node_client.Generate(t, bitcoind, 3)
```

If ARC was fast to update the status then this would fail, as it already has status REJECTED, I think.

shotasilagadzetaal: No this is what we expect before updating status to rejected, we control it by correct timing. at that moment the `last_modified_at` is not old enough to change status to rejected

boecklim (approved):

Issue #907, opened 2025-06-20
[FEATURE] Mark SEEN_ON_NETWORK transactions as REJECTED if not mined in previous blocks


boecklim:
Currently transactions submitted to ARC can be stuck in SEEN_ON_NETWORK status. The feature which ensures such transactions which are not in any mempool of a connected peer has already been addressed: https://github.com/bitcoin-sv/arc/pull/886

In addition it has to be ensured that transactions don't get rejected just because there was a high block delta time.

A transaction which is in SEEN_ON_NETWORK status for a defined amount of time, should only be rejected if it hasn't been in this status while at least e.g. 3 blocks were mined without it. The number of blocks can be configured in the settings.

That will prevent a lot of incorrect rejected states that would otherwise hit during the rare times with large block time deltas

PR #924, opened 2025-07-06
feat(BPAAS-2199): consider number of blocks when rejecting pending txs

shotasilagadzetaal:
## Description of Changes

Before rejecting SEEN_ON_NETWORK transactions now we consider how many blocks have been mined since tx submission. #907

## Testing Procedure

- [x] All tests pass locally

## Checklist:


test/config/config.yaml
```diff
@@ -55,9 +55,11 @@ metamorph:
   reAnnounceSeen:
     pendingSince: 10m
     lastConfirmedAgo: 10m
+    blocksSince: 3
```
boecklim, line 58: It seems this setting should not be here

internal/metamorph/processor_routines.go
```diff
@@ -95,9 +96,22 @@ func RejectUnconfirmedRequested(ctx context.Context, p *Processor) []attribute.K
 	var offset int64
 	var totalRejected int
 	var txHashes []*chainhash.Hash
+	var blocksSinceLastRequestedAlgo *blocktx_api.NumOfBlocksSinceResponse
 	var err error
 
 	for {
+		blocksSinceLastRequestedAlgo, err = p.blocktxClient.NumOfBlocksSince(ctx, p.now().Add(-p.rejectPendingSeenLastRequestedAgo))
```
boecklim, line 103: Typo in ...Algo => Ago

internal/metamorph/processor_routines.go
```diff
@@ -95,9 +96,22 @@ func RejectUnconfirmedRequested(ctx context.Context, p *Processor) []attribute.K
...
+		if err != nil {
+			p.logger.Error("Failed to get blocks since last requested", slog.String("err", err.Error()))
+			break
+		}
+
+		// if enough number of blocks are not mined since the time then skip rejecting
+		if blocksSinceLastRequestedAlgo.GetNumOfBlocks() < p.rejectPendingBlocksSince {
+			p.logger.Info("Skipping rejecting unconfirmed txs", slog.Uint64("numOfBlocks", blocksSinceLastRequestedAlgo.GetNumOfBlocks()), slog.Uint64("required", p.rejectPendingBlocksSince))
+			break
```
boecklim: I think this logic above does not implement the requirement that is specified in https://github.com/bitcoin-sv/arc/issues/907

What happens here is that if during a constant time of `p.rejectPendingSeenLastRequestedAgo` there were less than a constant `p.rejectPendingBlocksSince` blocks mined, then break here.

For example if in the last 10min less than 3 blocks were mined, then break and don't reject. This will almost always be false as there is on average 1 block per 10min

I think the logic should be as described here: https://github.com/bitcoin-sv/arc/pull/924/files#r2189999564

boecklim (approved):

PR #953, opened 2025-09-08
feat(BPAAS-2310): callback queue per tx

boecklim:
## Description of Changes

- Send callbacks in chronological order by transaction ID instead of URL

## Testing Procedure

- [X] All tests pass locally

## Checklist:

- [X] I have updated `CHANGELOG.md` with my changes

### DIFF
--- a/internal/callbacker/store/postgresql/postgres_test.go
+++ b/internal/callbacker/store/postgresql/postgres_test.go
@@ -237,7 +237,6 @@ func TestPostgresDBt(t *testing.T) {
 		require.True(t, isSorted, "records are not sorted by timestamp ascending")
 
 		expected := store.CallbackData{
-			ID:         4,
 			URL:        "https://arc-callback-1/callback",
 			Token:      "token",
 			Timestamp:  time.Date(2024, 9, 1, 12, 0, 1, 0, time.UTC),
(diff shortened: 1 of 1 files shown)


internal/callbacker/processor_routines.go
```diff
@@ -40,19 +45,62 @@ func LoadAndSendCallbacks(p *Processor, isBatch bool, sendFunc func(url string,
...
 
-	urlCallbacksMap := map[string][]*store.CallbackData{}
+	hashCallbacksMap := map[callbackKey][]*store.CallbackData{}
 	for _, callbackRecord := range callbackRecords {
-		urlCallbacksMap[callbackRecord.URL] = append(urlCallbacksMap[callbackRecord.URL], callbackRecord)
+		key := callbackKey{
+			txID: callbackRecord.TxID,
+			url:  callbackRecord.URL,
+		}
+		hashCallbacksMap[key] = append(hashCallbacksMap[key], callbackRecord)
```
shotasilagadzetaal, line 54: With specific tx id and URL shouldn't we just have one callback to send?

Like if there are two, can we send the latest status? why do we have slice here

shotasilagadzetaal (approved):

Issue #959, opened 2025-09-15
Maximum retries for callbacks


boecklim:
Implement a configurable nr for maximum retries per callback.

Once a callback was retried unsuccessfully that number of times, the callback is not retried anymore.

PR #960, opened 2025-09-15
feat(BPAAS-2341): callbacker max retries

boecklim:
## Description of Changes

- Store retries in callbacker and update
- Only attempt each callback a maximum nr of times

Closes https://github.com/bitcoin-sv/arc/issues/959

## Testing Procedure

- [x] All tests pass locally

## Checklist:

### DIFF
--- a/internal/callbacker/processor.go
+++ b/internal/callbacker/processor.go
@@ -27,6 +27,7 @@ const (
 	storeCallbacksIntervalDefault = 2 * time.Second
 	storeCallbackBatchSizeDefault = 500
 	sendCallbacksInterval         = 2 * time.Second
+	maxRetriesDefault             = 10
 )
 
 type Processor struct {
@@ -44,6 +45,7 @@ type Processor struct {
 	batchSendInterval      time.Duration
 	clearInterval          time.Duration
 	clearRetentionPeriod   time.Duration
+	maxRetries             int
 
 	wg        *sync.WaitGroup
 	cancelAll context.CancelFunc
@@ -98,6 +100,12 @@ func WithStoreCallbacksInterval(d time.Duration) func(*Processor) {
 	}
 }
 
+func WithMaxRetries(retries int) func(*Processor) {
+	return func(m *Processor) {
+		m.maxRetries = retries
+	}
+}
+
 func toEntry(callbackData *store.CallbackData) CallbackEntry {
 	return CallbackEntry{
 		Token:      callbackData.Token,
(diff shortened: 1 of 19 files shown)


internal/callbacker/store/postgresql/postgres.go
```diff
@@ -159,13 +159,13 @@ func (p *PostgreSQL) Insert(ctx context.Context, data []*store.CallbackData) (in
...
 
-func (p *PostgreSQL) GetUnsent(ctx context.Context, limit int, expiration time.Duration, batch bool) ([]*store.CallbackData, error) {
+func (p *PostgreSQL) GetUnsent(ctx context.Context, limit int, expiration time.Duration, batch bool, maxRetries int) ([]*store.CallbackData, error) {
 	const q = `
 				WITH updated AS (
 				UPDATE callbacker.transaction_callbacks c SET pending = $1
 				WHERE c.id IN (
 				    SELECT id FROM callbacker.transaction_callbacks c
-					WHERE timestamp > $2 AND allow_batch = $3 AND sent_at IS NULL AND (c.pending IS NULL OR c.pending < $5)
+					WHERE timestamp > $2 AND allow_batch = $3 AND sent_at IS NULL AND (c.pending IS NULL OR c.pending < $5) AND retries <= $6 AND disable IS NOT TRUE
```
tngdomingo, line 168: I might be missing something . If I have a callback c1 with maxRetries=1 , the first time unsetPending increases from 0 to 1 and then when you call getUnset, retries <= $6 (1) , will also include c1. Shouldn't that be retries<$6?

internal/callbacker/processor.go
```diff
@@ -27,6 +27,7 @@ const (
 	storeCallbacksIntervalDefault = 2 * time.Second
 	storeCallbackBatchSizeDefault = 500
 	sendCallbacksInterval         = 2 * time.Second
+	maxRetriesDefault             = 10
```
tngdomingo, line 30: Why is this value not retrieved from getCallbackerConfig() ? Are there different retries?

internal/callbacker/store/postgresql/postgres.go
boecklim: Yes, you're right 👍 

Will change that

internal/callbacker/processor.go
tngdomingo: Ok. Yes I saw that but I was confused because we were overwriting with the same value. 
My point probably is ... cannot we make
`maxRetriesDefault  = arcConfig.Callbacker.MaxRetries`
and remove 
`callbacker.WithMaxRetries(arcConfig.Callbacker.MaxRetries), `
and only add WithMaxRetries for when the value is not the default one. Not sure if that makes sense.

domingotaal (approved):

PR #969, opened 2025-09-29
fix(BPAAS-2151): broadcaster cli

boecklim:
## Description of Changes

- Extend time duration to get utxos from WoC

## Testing Procedure

- [ ] All tests pass locally

## Checklist:


internal/broadcaster/rate_broadcaster.go
```diff
@@ -269,7 +272,7 @@ func (b *UTXORateBroadcaster) broadcastBatchAsync(txs sdkTx.Transactions, errCh
 	go func() {
 		defer b.wg.Done()
```
tngdomingo, line 273: We could use wg.go here as well. Should we create a PR just to refactor that  or take advantage of other PRs like this to do it little by little?

